PR #7, opened 2025-12-29
Add headless auth API for native mobile apps

rabble:
## Summary

Adds pure JSON authentication endpoints for native mobile apps (Flutter, React Native, etc.) that want to handle their own UI instead of using web views.

## New Endpoints

- `POST /api/headless/register` - Register new user, returns `device_code` for email verification polling
- `POST /api/headless/login` - Login existing user, returns authorization code directly  
- `POST /api/headless/authorize` - Create new authorization for already-authenticated user

## Flow

**Registration:**
1. App calls `/api/headless/register` with email/password
2. Gets back `device_code` and shows "check your email" 
3. Polls `/api/oauth/poll?device_code=xxx` until email verified
4. Exchanges code for `bunker_url` via `/api/oauth/token`

**Login:**
1. App calls `/api/headless/login` with email/password
2. Gets back authorization `code` directly
3. Exchanges code for `bunker_url` via `/api/oauth/token`

## Files Changed

- `api/src/api/http/headless.rs` - New module with handlers
- `api/src/api/http/mod.rs` - Module declaration
- `api/src/api/http/routes.rs` - Route registration
- `api/tests/headless_auth_test.rs` - Integration tests (6 tests, all pass)
- `docs/HEADLESS_AUTH.md` - Documentation with Flutter examples

## Testing

All 6 new tests pass. No cookies, no web views, any origin allowed.

### DIFF
--- a/api/src/api/http/headless.rs
+++ b/api/src/api/http/headless.rs
@@ -1,0 +1,577 @@
+// ABOUTME: Headless authentication handlers for native mobile apps (Flutter, etc.)
+// ABOUTME: Pure JSON API - no cookies, no HTML, returns access_token directly
+
+use axum::{
+    extract::State,
+    http::StatusCode,
+    response::{IntoResponse, Response},
+    Json,
+};
+use bcrypt::verify;
+use chrono::{Duration, Utc};
+use keycast_core::metrics::METRICS;
+use keycast_core::repositories::{
+    OAuthCodeRepository, PolicyRepository, StoreOAuthCodeWithRegistrationParams,
+    UserRepository,
+};
+use nostr_sdk::Keys;
+use rand::Rng;
+use serde::{Deserialize, Serialize};
+
+use super::auth::{generate_secure_token, EMAIL_VERIFICATION_EXPIRY_HOURS};
+use super::oauth::{extract_origin, parse_policy_scope};
+
+// ============================================================================
+// Headless Registration
+// ============================================================================
+
+#[derive(Debug, Deserialize)]
+pub struct HeadlessRegisterRequest {
+    pub email: String,
+    pub password: String,
+    /// Client app identifier (e.g., "Divine Video", "My Nostr App")
+    pub client_id: String,
+    /// OAuth redirect URI (used to derive origin for bunker)
+    pub redirect_uri: String,
+    /// Optional: import existing Nostr key (nsec1... or hex)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub nsec: Option<String>,
+    /// OAuth scope (e.g., "policy:social")
+    pub scope: Option<String>,
+    /// PKCE code challenge (S256)
+    pub code_challenge: Option<String>,
+    /// PKCE challenge method (should be "S256")
+    pub code_challenge_method: Option<String>,
+    /// OAuth state parameter for CSRF protection
+    pub state: Option<String>,
+}
+
+#[derive(Debug, Serialize)]
+pub struct HeadlessRegisterResponse {
+    pub success: bool,
+    /// Nostr public key (hex)
+    pub pubkey: String,
+    /// Email verification is required before login
+    pub verification_required: bool,
+    /// Device code for polling (RFC 8628 pattern)
+    pub device_code: String,
+    /// Email address for display
+    pub email: String,
+}
+
+/// POST /api/headless/register
+/// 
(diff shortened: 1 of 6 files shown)


dcadenas:
I have significant concerns about this approach, some strategic, some technical.

  **Security concerns**

  The immediate issue is permissive CORS on these endpoints - any website can call `/api/headless/login` and harvest credentials directly.

  This also implements OAuth 2.0 ROPC (Resource Owner Password Credentials), which I intentionally removed. [RFC 9700](https://datatracker.ietf.org/doc/rfc9700/) (January 2025) states: **"The resource owner password credentials grant MUST NOT be used."** [OAuth 2.1](https://oauth.net/2.1/) removes it entirely. While there's [debate about first-party apps](https://blog.logto.io/deprecated-ropc-grant-type), the standards don't carve out an exception. ROPC is a single request/response, so challenge-based authentication is impossible - [it's incompatible with MFA](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth-ropc) and passkeys out of the box. You could build those on top, but then you're implementing authentication features in the app instead of centralizing them at login.divine.video - duplicating work we'll need to do (or already have) in the provider anyway.

  **What the industry does**

  Google, GitHub, and ChatGPT all use system browser auth for their own first-party apps:

  Google:
  

  ChatGPT:
  

  Github:
  

  Looking at federated social networks specifically: [Mastodon removed password grant entirely in version 4.4](https://docs.joinmastodon.org/spec/oauth/). Bluesky's app still uses ROPC, but their official OAuth announcement states the app "has not **yet** been updated to use OAuth" - that "yet" signals intent. [Blacksky is actively implementing OAuth](https://github.com/blacksky-algorithms/rsky/issues/60), and the [recommended OAuth implementation for AT Protocol](https://github.com/graze-social/aip) doesn't include password grant at all.

  **Why this matters for us**

  We're building an identity provider for the Nostr ecosystem. Third-party apps shouldn't collect passwords directly - that's the whole point of [OAuth's delegated authorization model](https://developer.okta.com/docs/concepts/oauth-openid/). If Divine uses a different flow than what we'd recommend for third-parties, we're not training users on the pattern we want them to expect: trust login.divine.video, just like users trust accounts.google.com or github.com.

  With headless auth, users never see the domain. The password field in Divine looks identical to a password field in a phishing app that anyone can implement. If we train users to trust login.divine.video from the start, we're in a much better position to help Nostr adoption - new non-technical users get a familiar, secure login experience across the ecosystem. We have the opportunity to start right rather than migrate later - and we already have a server implementation doing exactly this.

  **On "web views"**

  The PR describes this as avoiding "web views," but if that means avoiding web pages entirely, that misses the core security model of OAuth. The redirect to a centralized login page IS the point - users provide their password to the identity provider, not to the app. The current flow uses ASWebAuthenticationSession (iOS) and Chrome Custom Tabs (Android) - system browser sheets, not embedded web views. This can be fully Divine-branded, just like Google/GitHub/ChatGPT brand their OAuth pages.

  We could discuss what specific UX friction we're trying to solve here.

rabble:
Yah na... @dcadenas you don't understand the primary goal of keycast for me and divine. The primary goal is a FIRST PARTY auth system and that it's seemlessly integrated in to divine. The goal 10000% is to never have to train users on anything. First and foremost what we're doing is providing a login system to divine, for divine users, from divine. Anything else like oauth to other nostr services or ability to get an nsec bunker is cool, supporting an open ecosystem, definitely something we want, but not the goal or purpose of keycast for divine. 

I have zero desire to support the corruption and broken newer versions of OAUTH. It was a good standards at 1.0, broken but workable at 2.0... beyond that it became corporate slop that was designed for big companies like OKTA and google and not workable for an open ecosystem. 

So i'm fine with adding CORS protection back, doesn't matter from the mobile app. We are not making users from divine, login with divine... doesn't make sense. It's our app, our auth server, we control everything, we ask suers for username and password directly in the app and do not throw up a web UI.. this is not negotiable.

dcadenas:
Sounds good. I'll fix the CORS and merge, and draft follow up issues for attestation (App Attest / Play Integrity) and native passkeys for when you're ready to prioritize them.

There's an [IETF draft](https://datatracker.ietf.org/doc/draft-ietf-oauth-first-party-apps/) for exactly this use case, first-party native auth without browser redirects. It discusses passkeys and attestation alongside it. Both are independent and can be added incrementally.

Context: divinevideo/divine-context#65

PR #24, opened 2026-02-07
fix: update cors configuration to allow login from divine.video

hm21:
Users on Discord are reporting that they can't log in to the web app. I tried it myself and got a "cors" error. 

I took a quick look and i guess we just need to update the cors settings. However it used `CORS_ALLOWED_ORIGIN` but in the code I see it actually reads from `ALLOWED_ORIGINS` so changed that part too:[https://github.com/divinevideo/keycast/blob/32acaad6191064c15d39f1f51085e5310e0faf9c/keycast/src/main.rs#L522-L523](https://github.com/divinevideo/keycast/blob/32acaad6191064c15d39f1f51085e5310e0faf9c/keycast/src/main.rs#L522-L523) Feel free to close it directly if I'm completely wrong here ;)

### DIFF
--- a/keycast-oauth-unified.yaml
+++ b/keycast-oauth-unified.yaml
@@ -47,8 +47,8 @@ spec:
               value: "true"
             - name: GCP_PROJECT_ID
               value: "openvine-co"
-            - name: CORS_ALLOWED_ORIGIN
-              value: "https://login.divine.video"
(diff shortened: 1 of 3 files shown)


kirstenswasey:
awesome have a good day thankyou for opening and replying

PR #70, opened 2026-04-08
Adding AWS KMS encryption provider

sergey3bv:
Should solve #61

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,14 +19,24 @@ ALLOWED_PUBKEYS=
 # Master encryption key path (generated via bun run key:generate)
 MASTER_KEY_PATH=./master.key
 
-# Use Google Cloud KMS instead of local file-based encryption
+# KMS provider selection: file, gcp, aws
+# Default: file
+KMS_PROVIDER=file
+
+# Legacy fallback (backward compatibility):
+# If KMS_PROVIDER is not set, USE_GCP_KMS=true selects gcp, otherwise file.
 USE_GCP_KMS=false
 
-# Google Cloud KMS configuration (only if USE_GCP_KMS=true)
+# Google Cloud KMS configuration (when KMS_PROVIDER=gcp)
 GCP_PROJECT_ID=your-project-id
 GCP_KMS_LOCATION=global
 GCP_KMS_KEY_RING=keycast-keys
 GCP_KMS_KEY_NAME=master-key
+
+# AWS KMS configuration (when KMS_PROVIDER=aws)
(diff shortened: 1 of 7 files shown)


sergey3bv:
Hey, @NotThatKindOfDrLiz, could you please review?

NotThatKindOfDrLiz (changes requested):
Thanks for putting this together. The overall shape of the change looks good: the AWS implementation follows the existing GCP key manager closely, the dependency is feature-gated, and the legacy `USE_GCP_KMS` fallback is preserved so existing deployments should not be disrupted.

I do want one change before merge:

`core/src/encryption/aws_key_manager.rs`
- `test_new_requires_key_id` mutates process env but does not have the `#[serial]` guard that the sibling env-dependent tests use. Because it calls `remove_var("AWS_KMS_KEY_ID")`, it can race with other tests under parallel execution and introduce flakiness. Please add `#[serial]` there before this lands.

Non-blocking follow-ups I’d still like tracked after this:
- Hoist the AWS retry constants to module-level `const`s for consistency with `GcpKeyManager`.
- Consider extracting the duplicated key-manager construction in `keycast/src/main.rs` into a small helper so the provider selection logic only lives in one place.
- When both `KMS_PROVIDER` and `USE_GCP_KMS` are set and disagree, log a warning so operators have a clearer migration path.
- Add docs for `KMS_PROVIDER`, `AWS_KMS_KEY_ID`, and `AWS_REGION` anywhere we currently describe KMS configuration.

Once the test isolation issue is fixed, this looks close.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I think I fixed issues from your comment, could you please take a look?

NotThatKindOfDrLiz:
Thanks, this addresses the requested fix and the follow-ups I called out.

One tiny test-isolation cleanup before I approve: please also add `#[serial]` to `test_encrypt_decrypt_roundtrip`. It reads `AWS_KMS_KEY_ID`, while `test_new_requires_key_id` removes that same process env var. Since unmarked tests can still run alongside `#[serial]` tests, this can still race on machines with AWS KMS env configured.

I ran the updated branch locally and both passed:
- `cargo test -p keycast_core --features aws`
- `cargo test -p keycast --features aws`

Once that annotation is added, I am good to approve.

sergey3bv:
> Once that annotation is added, I am good to approve.

Done

NotThatKindOfDrLiz:
Closing as superseded by #100. The AWS KMS implementation and related wiring landed through #100, which is the branch that completed review and was merged.

PR #71, opened 2026-04-08
Adding AWS SES email provider

sergey3bv:
Should close #62

### DIFF
--- a/api/src/email_service.rs
+++ b/api/src/email_service.rs
@@ -2,9 +2,18 @@
 // ABOUTME: Supports SendGrid for production and DevEmailSender for local development/testing
 
 use async_trait::async_trait;
+#[cfg(feature = "aws")]
+use aws_sdk_sesv2::types::{
+    Body as SesBody, Content as SesContent, Destination as SesDestination,
+    EmailContent as SesEmailContent, Message as SesMessage,
+};
+#[cfg(feature = "aws")]
+use aws_sdk_sesv2::Client as SesClient;
 use serde::Serialize;
 use std::env;
 use std::sync::{Arc, Mutex};
+#[cfg(feature = "aws")]
+use tokio::sync::OnceCell;
 
 /// Captured email for testing/inspection
 #[derive(Debug, Clone)]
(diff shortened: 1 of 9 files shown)


sergey3bv:
Hey, @NotThatKindOfDrLiz, could you please review?

NotThatKindOfDrLiz (changes requested):
I pulled the PR patch locally and checked the branch head before writing this. There are a couple of real issues here, but some of the earlier review feedback is overstated or incorrect, so I want to separate the blocking concerns from the noise.

The main blocker is feature wiring. `api/Cargo.toml` adds an `aws` feature and gates the SES implementation behind `#[cfg(feature = "aws")]`, but the shipped binary still depends on `keycast_api = { path = "../api" }` without enabling that feature. The Docker image also builds the unified `keycast` binary. In the current packaging, SES is not compiled into the artifact we actually ship, so `EMAIL_PROVIDER=ses` will not produce a real SES sender in deployment. It does log and falls back to dev rather than failing silently, but functionally this still means the new provider is not wired into production. That needs to be fixed before merge.

The second issue is the SES client lifecycle. `SesEmailSender::send_email()` loads AWS config and constructs a new SES client on every send. That means credential/provider resolution and client setup are repeated for every verification, reset, or claim email. Even if the SDK caches some internals, this is still the wrong shape for a long-lived service. The sender should initialize the client once and reuse it, either by making construction async, using a lazy/once cell, or storing an already-built client in the sender.

I also think the provider-selection code should be simplified before this lands. `decide_provider()` looks like a runtime decision function, but the `ses` path is really compile-time-dependent via `cfg!(feature = "aws")`, and `create_email_sender()` then applies additional `#[cfg]` guards on top. That does work, but it makes the control flow and the tests more misleading than they need to be. The current SES decision test is effectively asserting whatever the current compile target baked in. I would prefer either: 1) keep compile-time gating out of `decide_provider()` and make that function strictly about env intent, or 2) fold the decision and construction together so there is one obvious place where unsupported providers are handled.

A few notes on things I do *not* think should block this PR:
- The `DISABLE_EMAILS` inconsistency called out elsewhere is not present here. SendGrid already checks `DISABLE_EMAILS` in this branch.
- The template extraction itself looks fine and is a net cleanup.
- The larger `Cargo.lock` is real but expected when pulling in the AWS SDK; that is something to be conscious of, not by itself a reason to block.
- The `.env.example` `AWS_REGION` comment is acceptable as documentation of the SDK default chain, even though the code does not read it directly.

Please address the following before merge:
1. Wire the `aws` feature into the binary or deployment path that actually ships, so SES can be compiled into the production artifact.
2. Rework `SesEmailSender` so the AWS config/client is created once and reused.
3. Simplify the provider-selection flow so the compile-time feature gating is not presented as a runtime choice.
4. Add at least one build/test path that exercises the AWS-gated code (`--features aws`) so this cannot regress unnoticed.

Once those are fixed, I’m comfortable with the overall direction of the change.

sergey3bv:
Hey, @dcadenas and @rabble, I added fixes which @NotThatKindOfDrLiz requested. Could you please review or point out further directions?

NotThatKindOfDrLiz (changes requested):
Thanks for the updates here. I checked the current branch head again, and the earlier blockers around wiring the `aws` feature into the shipped binary and reusing the SES client look addressed: `keycast` now enables `keycast_api`'s `aws` feature, and `SesEmailSender` caches the client via `OnceCell`. The template extraction is also a good cleanup.

I still want one change before merge:

`api/src/email_service.rs`
- Explicit `EMAIL_PROVIDER` misconfiguration still silently degrades to `DevEmailSender`. Right now `create_email_sender()` falls back to dev for `EMAIL_PROVIDER=sendgrid` without `SENDGRID_API_KEY`, for unknown `EMAIL_PROVIDER` values, and for unsupported explicit-provider cases that resolve to `ProviderDecision::Dev`.
- In production, dev sender is not a harmless fallback: it logs live verification/reset links to stdout/stderr instead of sending them. That means a bad config can both break account verification/password reset and leak active auth tokens into centralized logs.
- I think the right behavior is: if `EMAIL_PROVIDER` is explicitly set, and the requested provider cannot actually be satisfied, startup should fail hard with a clear error. Auto-detect fallback to dev is fine only when `EMAIL_PROVIDER` is unset.

Non-blocking follow-ups I’d still like tracked before merge:
- Refactor provider selection so there is one source of truth instead of `decide_provider()` plus warning-specific re-derivation in `create_email_sender()`.
- Collapse the duplicated auto-detect branches in `decide_provider()`.
- Rename or move the SES “initialized” log so it does not claim success before the lazy client has actually been created.
- Add a unit test covering `EMAIL_PROVIDER=""` behaving like unset.
- Decide whether AWS support should be opt-in or baked in consistently across SES and KMS, since this PR bakes SES into `keycast` while the KMS work used opt-in features.
- Review whether `FROM_NAME` should be validated/escaped for RFC 5322 display-name safety before interpolating it into the SES `from` string.

Once explicit-provider misconfiguration fails closed instead of falling through to dev, I’m comfortable with the direction of this change.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I think I eliminated blockers, could you please review

NotThatKindOfDrLiz (changes requested):
Thanks for the updates. The provider-selection logic now looks correct, but this still is not ready to merge.

Blocking changes needed:

1. Fix the compile error in `keycast/src/main.rs`.

`errors` is a `Vec<String>`, but this line pushes a leaked `str`:

```rust
errors.push(Box::leak(e.into_boxed_str()));
```

Please change it to:

```rust
errors.push(e);
```

`e` is already a `String`.

2. Decide and fix the Docker production build behavior.

Right now `Dockerfile` still builds the production binary without the `aws` feature:

```dockerfile
RUN cargo build --release --bin keycast
```

Since this PR adds AWS SES support and `EMAIL_PROVIDER=ses` now correctly fails closed when the binary lacks AWS support, the shipped Docker image still cannot actually use SES.

My preference is to build the production binary with AWS support:

```dockerfile
RUN cargo build --release --features aws --bin keycast
```

If you intentionally do not want SES in the Docker image, please document that explicitly in the PR description and `.env.example`, but I think enabling the feature in Docker is the better outcome for this PR.

3. After pushing those fixes, I will approve the blocked fork workflow so CI can actually run.

Please verify locally before handing it back:

```bash
cargo check -p keycast
cargo check -p keycast --features aws
cargo fmt --check
```

CI is currently blocked in `action_required`, and only `license/cla` has run. Once the branch compiles and the Docker build behavior is fixed, I will unblock CI and re-review.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I updated the PR according to your last comment. Everything seems ready for a review

api/src/email_service.rs
```diff
@@ -43,6 +52,115 @@ pub trait EmailSender: Send + Sync {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct EmailTemplate {
+    subject: String,
+    html_content: String,
+    text_content: String,
+}
+
+fn build_verification_email_template(verification_url: &str) -> EmailTemplate {
+    EmailTemplate {
+        subject: "Verify your diVine email address".to_string(),
```
NotThatKindOfDrLiz: These new extracted templates should preserve the current product branding from `main`. Please change the helper copy back to `Divine` here and in the related HTML/text strings in these template builders; otherwise this introduces a user-visible regression in outbound emails.

api/src/email_service.rs
```diff
@@ -358,149 +476,336 @@ impl EmailSender for SendGridEmailSender {
...
+    from_name: String,
+    base_url: String,
+}
+
+#[cfg(feature = "aws")]
+impl SesEmailSender {
+    pub fn new() -> Self {
+        let from_email =
+            env::var("FROM_EMAIL").unwrap_or_else(|_| "noreply@divine.video".to_string());
+        let from_name = env::var("FROM_NAME").unwrap_or_else(|_| "diVine".to_string());
```
NotThatKindOfDrLiz: `main` already defaults `FROM_NAME` to `Divine`. Please keep that value here as well so the SES sender path matches the existing SendGrid path and does not regress branding.

api/src/email_service.rs
```diff
@@ -574,10 +879,104 @@ mod tests {
     }
 
     #[test]
-    fn test_production_without_sendgrid_key_fails() {
+    fn verification_template_includes_url_and_subject() {
+        let url = "https://example.com/verify-email?token=abc";
+        let template = build_verification_email_template(url);
+
+        assert_eq!(template.subject, "Verify your diVine email address");
```
NotThatKindOfDrLiz: These assertions should be updated to `Divine` to match current behavior on `main`. Please also review the related new template tests in this file so the new helper coverage aligns with the corrected copy.

NotThatKindOfDrLiz (changes requested):
Thanks for addressing the earlier blockers. The compile/build/feature-gating issues I called out before are fixed, and the SES/template refactor is directionally good.

I still need changes before this can be approved. This head reintroduces the old `diVine` branding in the extracted email templates, the SES default `FROM_NAME`, and the corresponding tests, while `main` already uses `Divine`. That is a user-visible regression and needs to be corrected before merge.

Please update the new helper-based email copy, the SES default sender name, and the test assertions so they match current `main`.

Separately, I need to be explicit about process: this PR has required multiple review rounds to catch issues that should have been found by the branch author through a thorough self-review, rebasing against current `main`, and running the relevant tests before requesting review again. That amount of back-and-forth is resource-draining on our side. Before re-requesting review, please make sure the branch is fully rebased, the diff has been reviewed end-to-end against current `main`, and the affected paths have been exercised so we are not spending additional cycles catching avoidable regressions.

sergey3bv:
> Separately, I need to be explicit about process: this PR has required multiple review rounds to catch issues that should have been found by the branch author through a thorough self-review, rebasing against current `main`, and running the relevant tests before requesting review again.

It is a totally valid point. [GitHub provides a setting](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-suggestions-to-update-pull-request-branches) to prevent merging a stale branch, requiring a rebase. Could you please enable it to enforce behaviour you described.

NotThatKindOfDrLiz:
Superseded by #111, which contains the SES provider work plus the RFC-safe display-name fix on top. Closing this stale duplicate branch to keep the queue to one active SES line.

Issue #43, opened 2026-02-23
feat(nip05): assign npub to the name server for verification


rabble-claw:
## Summary
Allow users to assign their npub to a diVine-operated name server for NIP-05 identity verification (P1 launch feature).

## Requirements
- User can claim a username@divine.video (or similar domain) identifier
- Map username to npub on the name server
- Serve `.well-known/nostr.json` responses for verification
- Handle conflicts (username already taken)
- Integration with the mobile app profile settings
- Display NIP-05 verified badge on profiles

## Context
P1 launch feature per Liz's planning. NIP-05 verification gives users a human-readable identity and builds trust.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I am aiming to solve the issue with my PR. Currently I am [working on solving blockers](https://github.com/divinevideo/keycast/pull/72#issuecomment-4267316331).

PR #72, opened 2026-04-08
Adding NIP-05 verification (Assign npub to name server)

sergey3bv:
Should solve #43

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -197,6 +197,7 @@ Optional:
 - `USE_GCP_KMS`: Use Google Cloud KMS instead of file-based encryption (default: `false`)
 - `USE_VALKEY_IAM`: Enable GCP IAM authentication for Memorystore Valkey (default: `false`). When enabled, uses Workload Identity to obtain access tokens for Redis authentication. Tokens are cached and refreshed automatically before expiry.
 - `BUNKER_RELAYS`: Comma-separated relay URLs for NIP-46 communication (required, no default)
+- `NIP05_DOMAIN`: When the tenant host is `localhost` or `127.0.0.1` only, overrides the domain used in profile `nip05` (with `DOMAIN` as second choice); does **not** override the tenant domain for real hosts—profile and `/.well-known/nostr.json` must stay on the same host
 - `RUST_LOG`: Log level configuration (default: `info`)
 - `SQLX_POOL_SIZE`: Database connection pool size (should match Cloud Run concurrency, default: `50`)
 - `VITE_ALLOWED_PUBKEYS`: Comma-separated pubkeys for whitelist access (web frontend)
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -3843,7 +3843,10 @@ mod tests {
     #[test]
     fn test_normalize_nip05_username_rejects_invalid_chars() {
         let result = super::normalize_nip05_username("alice+name");
-        assert!(result.is_err(), "plus sign is not allowed in NIP-05 local-part");
+        assert!(
+            result.is_err(),
+            "plus sign is not allowed in NIP-05 local-part"
+        );
     }
 
     #[test]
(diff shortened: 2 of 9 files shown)


sergey3bv:
Hey, @rabble, could you please review?

NotThatKindOfDrLiz (changes requested):
Thanks for the NIP-05 work. The direction is good: returning `nip05` from `/api/user/profile`, serving `/.well-known/nostr.json?name=...`, and fixing the unknown-name response to return an empty `names` map are all useful changes. I also verified the schema and repository methods locally before writing this review.

I need to request changes before this can merge.

## Merge blockers

1. Case-insensitive username uniqueness is not enforced.

`normalize_nip05_username()` now lowercases new writes, but the current database uniqueness is still case-sensitive. `database/migrations/0001_initial_schema.sql` has `idx_users_username_tenant` on `(tenant_id, username)`, while `database/migrations/0007_admin_search_index.sql` adds only a non-unique `LOWER(username)` search index. `UserRepository::check_username_available()` and `find_pubkey_by_username()` also compare with `username = $1`.

That means an existing mixed-case username such as `Alice` could coexist with a newly normalized `alice`. Both collapse to the same NIP-05 local-part, which can produce two users for one `alice@domain` identity. That is identity-corrupting and should block merge.

Please add a migration that either backfills and enforces lowercase usernames or adds a unique index like `(tenant_id, LOWER(username)) WHERE username IS NOT NULL`, and update the availability/lookup paths to be case-insensitive or guaranteed-normalized. Please also add a regression test that specifically covers an existing mixed-case row conflicting with a new lowercase claim.

2. NIP-05 domain precedence needs to be fixed or explicitly justified.

`resolve_nip05_domain()` currently prefers `NIP05_DOMAIN`, then `DOMAIN`, then the tenant domain. In a multi-tenant deployment, that can show every tenant user as `name@process-domain` even when the tenant is served from a different host. Since `/.well-known/nostr.json` is resolved against the tenant domain, this can cause the profile response and the actual verification domain to diverge.

If Keycast is intentionally single-tenant for this deployment, please add a comment and documentation explaining why env vars intentionally override tenant domains. Otherwise, tenant domain should win and env vars should be fallback/local-dev overrides.

3. The dashboard badge should not overstate verification when domain resolution can be wrong.

The UI labels the local profile value as `NIP-05 Verified`. Once the domain precedence and uniqueness issues are fixed, that is probably fine because this service controls the corresponding `nostr.json` response. Until then, the badge can present a misleading verified identity if the profile returns a domain that does not match where discovery is actually served. Please resolve this together with blocker #2.

## Follow-up issues created

Per our review workflow, I created the non-blocking follow-ups and assigned them to @sergey3bv:

- #74: Follow-up: return 4xx validation errors from profile username updates
- #75: Follow-up: add tests for NIP-05 domain resolution
- #76: Follow-up: remove hardcoded default NIP-05 domain
- #77: Follow-up: make NIP-05 badge e2e test domain-independent
- #78: Follow-up: consider stricter dot rules for NIP-05 usernames
- #79: Follow-up: document nostr.json no-name response shape change
- #80: Follow-up: add NIP-05 username length limit

Once the case-insensitive uniqueness/data-safety issue and the domain precedence issue are addressed, this should be in much better shape.

sergey3bv:
@NotThatKindOfDrLiz Thank you for the review. I will immediately fix the two merge blockers (case‑insensitive uniqueness and domain precedence) as you requested.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I think I fixed blockers, could you please take a look?

NotThatKindOfDrLiz (approved):
Re-reviewed the current diff and I’m approving.

The prior merge blockers are resolved: case-insensitive username uniqueness is now enforced with the new lower(username) unique index and case-insensitive lookup/availability paths, and NIP-05 domain resolution now keeps public tenant domains authoritative while limiting env overrides to local development. The requested mixed-case regression coverage is present.

Non-blocking deploy note: please dry-run the username migration against staging/real data before production rollout, since the rare conflict path intentionally nulls later duplicate usernames.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I opened following PRs to solve follow-ups you created:
- #105: Adding validation error, solving #74;
- #109: Adding tests for NIP-05 domain resolution, solving #75;
- #103: Fixing default domain for NIP-05, solving #76;
- #102: Making e2e test domain independent, solving #77;
- #104: Making stricter rules for NIP-05, solving #78;
- #101: Documenting no-name response shape change, solving #79;
- #99: Adding username boundaries check, solving #80;

I would be happy if you could review them, so I can assist with other issues.

cc @rabble

PR #93, opened 2026-04-18
Improve keycast health probes

rabble:
## Summary
- return a small human-readable body from `/health`
- split liveness, startup, and readiness handling into dedicated endpoints
- make readiness fail during shutdown or when the database is unavailable, and update QA/cloudbuild checks to use the new endpoints

## Test Plan
- [x] `cargo test`

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -93,8 +93,8 @@ steps:
       - '--cpu-boost'
       - '--session-affinity'
       - '--use-http2'
-      - '--startup-probe=httpGet.path=/health,httpGet.port=3000,initialDelaySeconds=0,periodSeconds=3,failureThreshold=20,timeoutSeconds=2'
-      - '--liveness-probe=httpGet.path=/health,httpGet.port=3000,periodSeconds=30,timeoutSeconds=5,failureThreshold=3'
+      - '--startup-probe=httpGet.path=/healthz/startup,httpGet.port=3000,initialDelaySeconds=0,periodSeconds=3,failureThreshold=20,timeoutSeconds=2'
+      - '--liveness-probe=httpGet.path=/livez,httpGet.port=3000,periodSeconds=30,timeoutSeconds=5,failureThreshold=3'
       - '--add-cloudsql-instances=openvine-co:us-central1:keycast-db-plus'
       - '--network=default'
       - '--subnet=default'
@@ -114,12 +114,12 @@ steps:
         SERVICE_URL=$$(gcloud run services describe keycast --region=us-central1 --format='value(status.url)')
         echo "Service URL: $$SERVICE_URL"
 
-        echo "🔍 Testing API health endpoint..."
-        if ! curl -f -s "$$SERVICE_URL/health"; then
-          echo "❌ Health check failed"
+        echo "🔍 Testing API readiness endpoint..."
+        if ! curl -f -s "$$SERVICE_URL/healthz/ready"; then
+          echo "❌ Readiness check failed"
           exit 1
         fi
-        echo "✅ Health check passed"
+        echo "✅ Readiness check passed"
 
         PREVIEW_ORIGIN="https://preview.openvine-app.pages.dev"
 
--- a/keycast/src/main.rs
+++ b/keycast/src/main.rs
@@ -40,8 +40,18 @@ use tracing::Level;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
 use zeroize::Zeroizing;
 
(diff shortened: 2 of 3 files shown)


cloudbuild.yaml
```diff
@@ -114,12 +114,12 @@ steps:
         SERVICE_URL=$$(gcloud run services describe keycast --region=us-central1 --format='value(status.url)')
         echo "Service URL: $$SERVICE_URL"
 
-        echo "🔍 Testing API health endpoint..."
-        if ! curl -f -s "$$SERVICE_URL/health"; then
-          echo "❌ Health check failed"
+        echo "🔍 Testing API readiness endpoint..."
+        if ! curl -f -s "$$SERVICE_URL/healthz/ready"; then
```
dcadenas, line 118: The smoke test now requires the database to be reachable.
It used to hit `/health`, which passed even if the database was down.
It now hits `/healthz/ready`, which fails when the database does not respond within 800 ms.
A brief database hiccup during a deploy will now fail the smoke step and stop the pipeline.
Flagging so the first such failure is not a surprise.

keycast/src/main.rs
```diff
@@ -37,8 +40,70 @@ use tracing::Level;
...
+
+async fn livez(state: Arc<LivenessState>) -> impl IntoResponse {
+    if state.shutting_down.load(Ordering::Relaxed) {
+        (StatusCode::OK, "OK (shutting down)")
+    } else {
+        (StatusCode::OK, "OK")
+    }
+}
+
+async fn startupz() -> impl IntoResponse {
```
dcadenas, line 87: The startup probe returns OK without checking the database.
On Cloud Run this means the service is marked started as soon as the HTTP server answers.
That works today because the TCP listener only binds after `Database::new().await` and migrations have run inside `async_main`.
If someone later moves the listener bind earlier, `/healthz/startup` will start returning OK before the database is reachable.
Worth a one-line note in the handler so the invariant is not silent.

keycast/src/main.rs
```diff
@@ -37,8 +40,70 @@ use tracing::Level;
...
+    (StatusCode::OK, "OK")
+}
+
+async fn readyz(state: Arc<ReadinessState>) -> impl IntoResponse {
```
dcadenas, line 91: When shutting down, the code skips the database check and passes `database_ready=false` into `readiness_response`.
The outer shutting-down branch then wins, so the response is 503 `Shutting down` rather than 503 `Database unavailable`.
A one-line comment on the `if is_shutting_down` arm (for example `skip DB check during shutdown; outer branch wins`) would make this ordering intent obvious.

keycast/src/main.rs
```diff
@@ -905,6 +982,7 @@ async fn async_main(worker_threads: usize) -> Result<(), Box<dyn std::error::Err
 
     // Wait for shutdown signal
     wait_for_shutdown_signal().await;
+    shutting_down.store(true, Ordering::Relaxed);
```
dcadenas, line 985: Cloud Run does not use readiness probes.
It routes traffic based on startup and liveness only.
Making `/healthz/ready` return 503 during shutdown therefore does not speed up traffic drain on Cloud Run.
Drain there still relies on SIGTERM and the platform's 30-second grace window.
The signal is useful for the Cloud Build smoke test and for any Kubernetes consumer, which is fine; worth being explicit about in a comment here.

keycast/src/main.rs
```diff
@@ -1126,4 +1209,64 @@ mod tests {
...
+    #[test]
+    fn test_readiness_response_reports_database_failures_as_unavailable() {
+        let (status, body) = readiness_response(false, false);
+
+        assert_eq!(status, StatusCode::SERVICE_UNAVAILABLE);
+        assert_eq!(body, "Database unavailable");
+    }
+
+    /// Router-level smoke test: verifies the probe routes are mounted at the
+    /// exact paths the kubelet hits. Catches typos like `/healthz/startup`
```
dcadenas, line 1250: Consider extending this router smoke test to `/healthz/ready` in a follow-up.
Using `sqlx::PgPool::connect_lazy` against an unreachable URL would let the 503 `Database unavailable` path run end-to-end through the router, not just through `readiness_response`.
The current pure-function tests already catch most regressions, so this is a suggestion, not a requirement.

dcadenas (approved):
Approve. The endpoint split and the shutdown and database behaviors on `/healthz/ready` match the issue, and four small tests cover the response logic and the route mounts.

Docs referencing `/health` as the liveness or startup endpoint (`docs/DEPLOYMENT.md`, `docs/MONITORING.md`, `docs/SIGNER_RELIABILITY_GUIDE.md`) were not updated in this PR. Worth a short follow-up.

PR #95, opened 2026-04-20
feat(nip65): publish kind 10002 relay lists on account provisioning

### LINKED ISSUE #94: feat: publish kind:10002 NIP-65 relay list on account provisioning
## Problem
Divine/Keycast-provisioned accounts have no kind:10002 (NIP-65 relay list) on any public indexer. When the mobile app signs a user in, `RelayDiscoveryService` queries `purplepag.es`, `user.kindpag.es`, `relay.nos.social` and gets nothing back, so the client falls back to a generic relay set (or, on pre-#2955 builds, to Divine relay only). This degrades reachability for every Keycast user and contributes to the failures in divinevideo/divine-mobile#3152, divinevideo/divine-mobile#3161, divinevideo/divine-mobile#3162.

## Proposed behavior
At new account creation, Keycast publishes a kind:10002 on behalf of the user with at minimum:

```
["r", "wss://relay.divine.video"]
```

and broadcasts it to the three indexers the mobile client queries:

- `wss://purplepag.es`
- `wss://user.kindpag.es`
- `wss://relay.nos.social`

## Why here
Keycast owns the signing keys for these accounts; the mobile client uses a remote-signer relationship and never holds the nsec, so it cannot reliably own this producer responsibility.

## Acceptance criteria
- [ ] New Keycast accounts have a kind:10002 discoverable on all three indexers within N seconds of provisioning
- [ ] Broadcast is idempotent/safe to retry (publish once, overwritable by the user from another client via standard replaceable-event semantics)
- [ ] Backfill path for existing Keycast-provisioned accounts that don't have a kind:10002 today
- [ ] Telemetry: log publish success/failure per indexer

## Evidence
- divinevideo/divine-mobile#3152, divinevideo/divine-mobile#3161, divinevideo/divine-mobile#3162 — three independent Zendesk reports within 48h, all showing `⚠️ No relay list found for <npub> on any indexer`
- Mobile consumer: `mobile/lib/services/relay_discovery_service.dart:17-24` (default indexers) in `divinevideo/divine-mobile`
- Mobile fallback (partial mitigation only): `divinevideo/divine-mobile` PR #2955

## Risk
Low. kind:10002 is NIP-16 replaceable; users who later edit via another client overwrite cleanly.

sergey3bv:
Should close #94


NotThatKindOfDrLiz:
Tagging @dcadenas for design/ownership review before approval.

I do not see this as blocked on implementation craftsmanship so much as on three product/security decisions that need an explicit owner call:

1. Policy bypass
This PR signs and publishes kind `10002` directly instead of going through the normal signer policy path. That may be the right choice for a system-issued protocol event, but it needs to be explicit: should these events bypass user signing policy, or should they be enforced the same way as user-initiated signing?

2. Privacy behavior change
This auto-publishes every new user's relay list to public indexers. That is a real privacy/product change, not just protocol plumbing. Should this be unilateral, opt-out, opt-in, or otherwise surfaced to users?

3. Backfill safety
The backfill example decrypts user keys and publishes for the full user set. Even with `DRY_RUN=true` by default, this feels like an ops-sensitive action rather than a casual example. Should this live as a shipped example at all, or should backfill be handled as a separate controlled procedure?

Also, before merge consideration, I'd want the branch to have an actual CI run.

So my recommendation is: hold approval pending a decision from @dcadenas on the above design questions, then re-run CI and continue from there.

api/src/api/http/auth.rs
```diff
@@ -724,6 +725,21 @@ pub async fn register(
         };
     }
 
+    // Publish a discoverable NIP-65 relay list for new user-facing accounts.
```
dcadenas: Gate this publish on `req.nsec.is_none()` so Keycast only publishes for keys it generated itself.

Today `register` accepts a caller-supplied `req.nsec` (auth.rs:647-658), so `keys` here may be an identity that already has a curated kind:10002 on the network. The spawn below unconditionally publishes a minimal list containing only `wss://relay.divine.video` to all three indexers, and kind:10002 is replaceable, so whatever the user had on purplepag.es, user.kindpag.es, or relay.nos.social gets overwritten. Issue #94's AC "overwritable by the user from another client" is about freshly-provisioned accounts, not BYOK nsecs arriving with a public footprint. If gating on the auto-generated branch is too coarse, query the indexers for an existing kind:10002 first and skip or merge.

api/src/api/http/oauth.rs
```diff
@@ -2611,6 +2612,47 @@ async fn handle_authorization_code_grant(
             pending_email_val
         );
 
+        // Publish a discoverable NIP-65 relay list for newly created OAuth accounts.
```
dcadenas: Same BYOK overwrite problem as the `register` path. This spawn runs whenever `pending_email` is `Some`, which includes the BYOK branch at oauth.rs:2562-2581 where the nsec arrived via `code_verifier`. A third-party OAuth client registering with a pre-existing nsec has its curated NIP-65 clobbered on all three indexers. Condition the spawn on the auto-generated-key branch (for example the pre-fallback `pending_encrypted_secret.is_some()` check) rather than the outer `pending_email`, or query for an existing kind:10002 before publishing.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    let rows = if limit > 0 {
+        sqlx::query(
+            "SELECT u.pubkey, pk.encrypted_secret_key
+             FROM users u
+             JOIN personal_keys pk ON pk.user_pubkey = u.pubkey AND pk.tenant_id = u.tenant_id
+             WHERE u.tenant_id = $1
+               AND u.email IS NOT NULL
+             ORDER BY u.created_at ASC
+             LIMIT $2",
```
dcadenas, line 221: Backfill is non-idempotent and will clobber user-edited relay lists on a second run. The SELECT joins `users` to `personal_keys` with only `email IS NOT NULL` and `tenant_id = $1`. There is no filter for users who already have a published kind:10002, whether published by this PR's new `register` path or by the user themselves from another client. Any re-run, or any run after users edited their list elsewhere, overwrites those lists with the single-relay event. Either query indexers for an existing kind:10002 per user and skip if present, or persist a `relay_list_published_at` column and filter on it. Without one of these, #94's AC "overwritable by the user from another client" is violated on every re-run.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,168 @@
...
+
+fn required_env(name: &str) -> String {
+    env::var(name).unwrap_or_else(|_| panic!("{} is required", name))
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let database_url = required_env("DATABASE_URL");
+    let tenant_id: i64 = env::var("TENANT_ID")
+        .unwrap_or_else(|_| "1".to_string())
```
dcadenas, line 176: This `DRY_RUN` parse is a destructive footgun. `env::var("DRY_RUN").unwrap_or_else(|_| "true".to_string()).eq_ignore_ascii_case("true")` treats anything that isn't literally `true`/`True`/`TRUE` as a live run. That includes `DRY_RUN=1`, `DRY_RUN=yes`, `DRY_RUN=on`, `DRY_RUN=false`, `DRY_RUN=0`, and `"true "` with trailing whitespace. An operator typing the idiomatic `DRY_RUN=1` expecting truthy-means-dry triggers a live publish against every matched user's key. Use a strict allowlist (accept `1`/`true`/`yes` for dry, `0`/`false`/`no` for live, reject anything else), or invert to `LIVE_RUN=true` so the default-unset behavior is dry.

api/src/api/http/oauth.rs
dcadenas: Nit, but worth fixing while the BYOK gating is being addressed anyway: this adds a second KMS round-trip on the OAuth hot path. A few lines above this spawn, the same handler encrypted `secret_bytes` into `encrypted_secret` via `key_manager.encrypt(...)`. The new code then calls `key_manager.decrypt(&encrypted_secret)` synchronously to recover the same plaintext before spawning. Every OAuth registration now pays encrypt + decrypt on the critical path before the token response returns. The plaintext `secret_bytes` (and the `Keys` derived from it) is still in scope above the encrypt — plumb it through instead of re-decrypting.

api/src/relay_list_publisher.rs
```diff
@@ -0,0 +1,216 @@
...
+
+    if tokio::time::timeout(CONNECT_TIMEOUT, client.connect())
+        .await
+        .is_err()
+    {
+        return Err("connect timed out".to_string());
+    }
+
+    let send_result = tokio::time::timeout(SEND_EVENT_TIMEOUT, client.send_event(event)).await;
+    match send_result {
```
dcadenas: This matches only `Ok(_) => Ok(())`, which means a relay that accepts the websocket but rejects the event (NIP-20 `OK: false` for auth-required, rate-limited, filtered, etc.) is logged as success. That breaks the issue's AC "Telemetry: log publish success/failure per indexer" — dashboards will over-report success. `client.send_event` returns an `Output` with per-relay accept/reject in `success` / `failed`; inspect those and only mark success when this indexer's URL is in `success`.

api/src/api/http/auth.rs
```diff
@@ -724,6 +725,21 @@ pub async fn register(
...
+    // Non-fatal side effect: registration must succeed even if indexers are unavailable.
+    let keys_for_publish = keys.clone();
```
dcadenas: This `tokio::spawn` is fire-and-forget with no durable record, so the issue's AC "Broadcast is idempotent/safe to retry" isn't met for in-flight publishes. If the API process is SIGTERM'd during the ~11s window (3s add_relay + 4s connect + 4s send_event), the work is lost and nothing records that this user still needs a kind:10002. The one-off backfill binary is not a retry path for crashed in-flight publishes. Either persist a `pending_relay_list_publish` marker cleared on success, or drive the same code from a scheduled reaper so crashed publishes get retried automatically.

api/src/relay_list_publisher.rs
```diff
@@ -0,0 +1,216 @@
...
+    EventBuilder::new(Kind::Custom(10002), "")
+        .tags(tags)
+        .sign_with_keys(keys)
+        .map_err(|e| RelayListPublisherError::Signing(e.to_string()))
+}
+
+async fn publish_event_to_indexer(keys: Keys, indexer: &str, event: &Event) -> Result<(), String> {
+    let client = Client::new(keys);
+
+    let add_result = tokio::time::timeout(ADD_RELAY_TIMEOUT, client.add_relay(indexer)).await;
```
dcadenas: Each indexer publish builds a fresh `Client::new(keys)`, calls `add_relay` + `connect`, and then returns without ever calling `disconnect()` or `shutdown()`. `nostr_sdk::Client` owns background relay-pool tasks; dropping the `Client` without disconnecting leaks those tasks and the underlying websocket until the runtime eventually GCs them, if at all. Three clients are created per registration and registration spawns are fire-and-forget, so under load the websocket tasks pile up in the API runtime. Call `client.disconnect().await` before returning, or share a single long-lived `Client` with all three relays added to it and tear it down once.

dcadenas (changes requested):
The main concern is that this PR can overwrite an existing NIP-65 relay list for users who bring their own key. Three paths hit this: the `register` handler, the OAuth code-verifier branch, and the backfill binary. A handful of smaller issues also need addressing before merge: a `DRY_RUN` default that treats typos as live, a per-relay success check that over-reports, and some resource and retry concerns in the publisher.

sergey3bv:
Hey, @NotThatKindOfDrLiz, @dcadenas, I updated the PR, could you please take a look

api/src/api/http/oauth.rs
```diff
@@ -49,6 +50,10 @@ pub fn extract_nsec_from_verifier_public(verifier: &str) -> Option<String> {
     None
 }
 
+fn should_publish_oauth_registration_relay_list(pending_encrypted_secret: Option<&[u8]>) -> bool {
```
dcadenas: Gate the OAuth relay-list publish on explicit auto-generate intent, not on `pending_encrypted_secret.is_some()`. At oauth.rs:3281-3308 that field is populated for both the auto-generate flow and the direct-nsec BYOK flow, so a user registering via OAuth with their own nsec will have a kind:10002 containing only `wss://relay.divine.video` published to the three indexers, overwriting any existing NIP-65 list they already published from another client (kind:10002 is replaceable). The comment at oauth.rs:2530 says only auto-generated registrations should trigger this, but the current gate does not enforce it. Record the intent explicitly on the `OAuthCode` row at /authorize time (for example an `is_generated` boolean) and gate on that. This matches the `req.nsec.is_none()` gate used in `auth.rs::register` and closes the BYOK overwrite path.

This is the blocking concern for the PR.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,272 @@
...
+        println!("Using file key manager");
+        Box::new(FileKeyManager::new()?)
+    };
+
+    let pool = PgPoolOptions::new()
+        .max_connections(10)
+        .connect(&database_url)
+        .await?;
+
+    let rows = if limit > 0 {
```
dcadenas, line 212: Restrict the backfill query to auto-generated accounts before running it live. The SELECT joins `users -> personal_keys` filtered only by `email IS NOT NULL`, which includes BYOK registrants. The PR description explicitly skips BYOK users at registration because they may already have their own relay list published from elsewhere, and the backfill needs to honor the same rule. Running LIVE_RUN against a tenant with BYOK email users overwrites their existing kind:10002 with a single-relay list.

Two options that would make this safe:

- Add a key-origin column at registration (for example `source='generated'|'byok'`) and filter on it here.
- Query each pubkey against the three indexers first and only publish when no existing kind:10002 is found.

As written there is no safe way to run this binary, so please pick one before landing.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,272 @@
...
+        .unwrap_or_else(|_| "0".to_string())
+        .parse()
+        .expect("LIMIT must be a number");
+
+    println!("=== kind:10002 backfill ===");
+    println!("tenant_id={}", tenant_id);
+    println!("dry_run={}", dry_run);
+    println!("limit={}", limit);
+
+    let use_gcp_kms = env::var("USE_GCP_KMS")
```
dcadenas: Mirror the KMS provider selection from `keycast/src/main.rs` here. Production treats `KMS_PROVIDER` as the source of truth (`file|gcp|aws`) with `USE_GCP_KMS` as a legacy fallback, but this backfill only checks `USE_GCP_KMS`. In a deployment where `KMS_PROVIDER=gcp` is set but `USE_GCP_KMS` is unset, the backfill silently falls through to `FileKeyManager` and every decrypt fails. `KMS_PROVIDER=aws` cannot be selected from this binary at all. The manager used to decrypt has to match the manager used to encrypt, so please match the main-binary selection logic directly rather than maintaining a second copy of the rule.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,272 @@
...
+    println!("published_partial={}", published_partial);
+    println!("failed={}", failed);
+
+    Ok(())
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{parse_dry_run_env, parse_live_run_env};
+
```
dcadenas, line 397: Split the `success_count == 0` case out from partial success. The current branch is `if success_count == outcomes.len() { ok } else { partial }`, so an all-failure run (every indexer rejected) is reported as PARTIAL and increments `published_partial`. The operator summary then cannot distinguish '2 of 3 accepted' from '0 of 3 accepted'. Treat `success_count == 0` as `failed` and keep `0 < success_count < outcomes.len()` as partial.

dcadenas (changes requested):
The core concern is that the 'auto-generated accounts only' invariant is not actually enforced end to end. The OAuth registration path infers that intent from encrypted-secret presence, which also fires for BYOK. The backfill example has no way to filter BYOK users at all. Both paths can overwrite a user's existing kind:10002 with a single-relay list, so this needs to be fixed before merge. One smaller nit on the backfill summary reporting is inline below.

NotThatKindOfDrLiz:
@sergey3bv Please address the requested changes from @dcadenas. TYSM!

sergey3bv:
Hey, @NotThatKindOfDrLiz and @dcadenas, I updated the PR, could you please take a look?

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,416 @@
...
+}
+
+async fn has_kind10002_on_indexer(pubkey: &PublicKey, relay: &str) -> Result<bool, String> {
+    let client = Client::new(Keys::generate());
+    let add_result = tokio::time::timeout(ADD_RELAY_TIMEOUT, client.add_relay(relay))
+        .await
+        .map_err(|_| "add_relay timed out".to_string())?;
+    add_result.map_err(|e| format!("add_relay failed: {e}"))?;
+
+    if tokio::time::timeout(CONNECT_TIMEOUT, client.connect())
```
dcadenas: This timeout does not guard anything.
In `nostr-sdk 0.44.1`, `Client::connect()` returns `()` after spawning a background task per relay.
It does not await the websocket handshake, so `tokio::time::timeout(CONNECT_TIMEOUT, client.connect())` resolves `Ok` immediately whether or not the indexer is reachable, and the `.is_err()` branch is dead code.

The consequence downstream is serious.
`fetch_events` on a client with zero connected relays resolves `Ok(empty)` rather than `Err`, because the SDK only collects from connected relays.
If all three `INDEXER_RELAYS` black-hole at once (cluster-wide outage, DNS issue, egress block), `has_existing_kind10002` returns `Ok(false)`, the backfill concludes the user has no kind:10002, and overwrites a BYOK user's richer relay list with the single-relay `wss://relay.divine.video` event.
NIP-16 replaceable semantics make that irreversible, which violates issue #94 AC2.

Use `client.try_connect_relay(indexer, CONNECT_TIMEOUT)` instead.
That one awaits the handshake and returns `Result<(), Error>`, so "no connected relay" surfaces as `Err` and the pre-query cannot fail open.
`client.connect_relay(url)` is not a substitute; it calls the same non-awaiting `relay.connect()` under the hood and only returns `RelayNotFound`.

api/src/relay_list_publisher.rs
```diff
@@ -0,0 +1,306 @@
...
+    let client = Client::new(keys);
+    let publish_result = async {
+        let add_result = tokio::time::timeout(ADD_RELAY_TIMEOUT, client.add_relay(indexer)).await;
+        match add_result {
+            Ok(Ok(_)) => {}
+            Ok(Err(e)) => return Err(format!("add_relay failed: {}", e)),
+            Err(_) => return Err("add_relay timed out".to_string()),
+        }
+
+        if tokio::time::timeout(CONNECT_TIMEOUT, client.connect())
```
dcadenas: Same dead `CONNECT_TIMEOUT` pattern as the backfill.
`Client::connect()` is fire-and-forget in `nostr-sdk 0.44.1`, so `tokio::time::timeout(CONNECT_TIMEOUT, client.connect()).await.is_err()` cannot fire for an unreachable indexer.

This is less dangerous than the backfill case because `send_event` on a client with zero connected relays returns `Ok(Output{success:[], failed:[]})` and `classify_send_event_outcome` correctly reports `send_event not accepted by target relay`, so the live path still surfaces the failure.
But the timeout wrapper adds zero protection and will mislead the next person to touch this code.

Replace with `client.try_connect_relay(indexer, CONNECT_TIMEOUT)` or `wait_for_connection(CONNECT_TIMEOUT)`.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,416 @@
...
+
+    Ok(has_events)
+}
+
+async fn has_existing_kind10002(pubkey: &PublicKey) -> Result<bool, String> {
+    for relay in relay_list_publisher::INDEXER_RELAYS {
+        match has_kind10002_on_indexer(pubkey, relay).await {
+            Ok(true) => return Ok(true),
+            Ok(false) => {}
+            Err(err) => {
```
dcadenas: `has_existing_kind10002` returns `Err` the moment any single indexer responds with an error, instead of trying the others.
A transient blip on `INDEXER_RELAYS[0]` (`purplepag.es`, iterated first) marks every user processed during the outage as `existing_kind10002_check_error` and counts them toward `failed`, even though no publish was attempted and indexers 1 and 2 might have answered `Ok(true)`.
Re-runs during the same outage fail the same users the same way.

This is also inconsistent with `publish_to_indexers_with` in `api/src/relay_list_publisher.rs:138-167`, which already tolerates per-indexer failures.

Continue past `Err` arms.
Return `Ok(true)` if any indexer reports the event, `Ok(false)` only if every indexer responded `false`, and treat all-three-errored as a separate `Unknown` outcome that is not conflated with `failed`.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,416 @@
...
+    } else {
+        sqlx::query(
+            "SELECT u.pubkey, pk.encrypted_secret_key
+             FROM users u
+             JOIN personal_keys pk ON pk.user_pubkey = u.pubkey AND pk.tenant_id = u.tenant_id
+             WHERE u.tenant_id = $1
+               AND u.email IS NOT NULL
+             ORDER BY u.created_at ASC",
+        )
+        .bind(tenant_id)
```
dcadenas, line 237: This `SELECT` has no `is_generated` filter, so the input set includes every BYOK email user.
The pre-query at the indexer is the only remaining defense, and the two findings above show it fails open.

The root cause is upstream.
`database/migrations/20260429154500_add_is_generated_to_oauth_codes.sql` adds the column to `oauth_codes` only, and `finalize_oauth_registration` (see `core/src/repositories/user.rs:863` and the basic `store` in `core/src/repositories/oauth_code.rs:85-106`, which does not bind `is_generated`) deletes the `oauth_codes` row at finalize without persisting the flag onto `users` or `personal_keys`.
After token exchange the BYOK-vs-generated signal is permanently lost, so the backfill has no SQL anchor to filter on.

Persist `is_generated` onto `users` (or `personal_keys`) at OAuth-finalize time so the backfill can SQL-filter it here.
Pre-migration users still need the indexer check, but that subset shrinks over time instead of staying the entire email-user population.

api/src/api/http/oauth.rs
```diff
@@ -2584,6 +2595,61 @@ async fn handle_authorization_code_grant(
             pending_email_val
         );
 
+        // Publish a discoverable NIP-65 relay list only for auto-generated OAuth accounts.
+        // Non-fatal side effect: token exchange must continue even if publish fails.
+        if should_publish_relay_list {
```
dcadenas: This publish path is unreachable for the users issue #94 is targeting.
The `is_generated` and `pending_email` fields that `oauth_register` and `headless_register` set on the initial placeholder `oauth_code` do not survive `verify_email`.
At `api/src/api/http/auth.rs:1512`, the `new_code` is stored via the basic `OAuthCodeRepository::store`, which does not bind `is_generated` and does not carry the `pending_*` fields.
At `api/src/api/http/auth.rs:1559`, the original row is deleted.

By the time `handle_authorization_code_grant` calls `find_valid` on the `new_code`, `pending_email` is `None`, so the registration branch at `api/src/api/http/oauth.rs:2507` is never entered and the publish at line 2604 never runs.
The mobile app uses the OAuth/headless flow, so none of those users get kind:10002 published from the live path.

Two fix shapes:

1. Persist `is_generated` and the `pending_*` fields alongside the `new_code` in `verify_email`'s `store` call. Extend `StoreOAuthCodeParams` plus the basic `store` INSERT to include `is_generated`, and have `verify_email` forward `oauth_data.is_generated`.
2. Move the publish trigger into `verify_email` itself when `oauth_data.is_generated && oauth_data.pending_encrypted_secret.is_some()`.

api/examples/backfill_kind10002.rs
```diff
@@ -0,0 +1,416 @@
...
+    let mut skipped_existing = 0_u64;
+    let mut failed = 0_u64;
+
+    for row in rows {
+        processed += 1;
+        let pubkey: String = row.get("pubkey");
+        let encrypted_secret: Vec<u8> = row.get("encrypted_secret_key");
+
+        if dry_run {
+            println!("[{processed}] DRY_RUN pubkey={}", pubkey);
```
dcadenas: `DRY_RUN` short-circuits before `has_existing_kind10002`, so every email user prints `DRY_RUN pubkey=…` and the operator cannot tell how many would actually be published versus skipped as `existing_kind10002_found`.
The dry-run count is an upper bound only.

This also makes it impossible to validate the silent-overwrite fix above in dry-run before a live run.

Run the indexer pre-query in dry-run too, and split the output into `DRY_RUN_PUBLISH pubkey=…` versus `DRY_RUN_SKIP pubkey=… reason=existing_kind10002_found`.

api/src/api/http/auth.rs
```diff
@@ -769,6 +774,25 @@ pub async fn register(
         };
     }
 
+    // Publish a discoverable NIP-65 relay list only for auto-generated accounts.
+    // Non-fatal side effect: registration must succeed even if indexers are unavailable.
+    if should_publish_registration_relay_list(req.nsec.as_deref()) {
+        let keys_for_publish = keys.clone();
+        let publish_pubkey = public_key.to_hex();
+        tokio::spawn(async move {
```
dcadenas: This fire-and-forget `tokio::spawn` has no recovery path, and the same shape repeats at `api/src/api/http/oauth.rs:2604`.
If the API process dies between spawning the task and the fan-out completing (k8s rollout, OOM, deploy), the publish is silently lost.
No DB flag, no retry queue, no outbox.
Issue #94 AC1 ("kind:10002 discoverable within N seconds of provisioning") becomes unbounded under crash.

Two options:

1. Write a `relay_list_publish_pending` row at registration and drain it from a worker. That also gives the backfill a SQL-side anchor for the subset that failed in-flight.
2. Run the publish before returning the 200 to the client, behind a tight overall timeout.

dcadenas (changes requested):
The two high-severity concerns here share a root cause and need to be fixed before this can merge.
The `tokio::time::timeout(CONNECT_TIMEOUT, client.connect())` pattern does not do what the code thinks it does in `nostr-sdk 0.44.1`, and on the backfill path that combines with a first-error short-circuit to silently overwrite a BYOK user's existing relay list with a single-relay event.
Separately, the `is_generated` signal never reaches the database rows the backfill and the OAuth publish path actually read, so the live OAuth/headless publish is unreachable and the backfill has no SQL filter for BYOK users.
Details and concrete fixes are inline.

sergey3bv:
Hey, @NotThatKindOfDrLiz and @dcadenas, I updated the PR, could you please take a look?

dcadenas:
We will retriage and pick this up later

PR #99, opened 2026-04-21
Adding username boundaries check

### LINKED ISSUE #80: Follow-up: add NIP-05 username length limit
Follow-up from PR #72.

sergey3bv:
Should close #80

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -32,6 +32,7 @@ const DEFAULT_TOKEN_EXPIRY_HOURS: i64 = 24;
 pub const EMAIL_VERIFICATION_EXPIRY_HOURS: i64 = 24;
 const PASSWORD_RESET_EXPIRY_HOURS: i64 = 1;
 const DEFAULT_NIP05_DOMAIN: &str = "divine.video";
+const MAX_NIP05_USERNAME_LENGTH: usize = 64;
 
 /// Get token expiry in seconds. Uses `TOKEN_EXPIRY_SECONDS` env var if set,
 /// otherwise defaults to 24 hours (86400 seconds).
@@ -85,6 +86,13 @@ fn normalize_nip05_username(raw_username: &str) -> Result<String, AuthError> {
         return Err(AuthError::Internal(
             "Username cannot start or end with a hyphen".to_string(),
         ));
+    }
+
+    if username.len() > MAX_NIP05_USERNAME_LENGTH {
+        return Err(AuthError::Internal(format!(
+            "Username must be at most {} characters",
+            MAX_NIP05_USERNAME_LENGTH
+        )));
     }
 
     Ok(username)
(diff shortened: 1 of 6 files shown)


api/src/api/http/auth.rs
```diff
@@ -1854,6 +1902,10 @@ pub async fn get_profile(
         .get_username(&user_pubkey, tenant_id)
         .await?
         .flatten();
+    let nip05_domain = resolve_nip05_domain(&tenant.0.domain);
```
NotThatKindOfDrLiz, line 1905: This is fabricating a NIP-05 identifier from the local username alone. That is not the same as a verified NIP-05, and it can be wrong in real deployments because the actual claim domain/source of truth may differ. Please don’t expose `nip05` from this endpoint unless it is backed by authoritative claim state or a real resolution check.

web/src/routes/+page.svelte
```diff
@@ -372,6 +390,17 @@ $effect(() => {
...
 						</div>
 					</div>
+					{#if userNip05}
+						<div class="identity-row">
+							<div class="identity-icon">
+								<Check size={20} weight="fill" />
+							</div>
+							<div class="identity-info">
+								<span class="identity-value">{userNip05}</span>
+								<span class="status-badge success">NIP-05 Verified</span>
```
NotThatKindOfDrLiz, line 400: This badge is the behavioral regression. At this point the UI only knows that a username string exists locally; it does not know that the NIP-05 mapping resolves and matches the user’s pubkey. Please remove the `Verified` wording here, or hold this UI change for a follow-up that has real verification semantics.

NotThatKindOfDrLiz (changes requested):
Thanks for adding the username-boundary coverage and the NIP-05 discovery tests. The validation work itself is useful, but I can’t approve this PR as-is because it also introduces user-visible NIP-05 behavior that is not actually verified.

The main issue is that this branch now synthesizes a `nip05` string from the locally stored username and displays it in the dashboard as `NIP-05 Verified`, even though the backend still allows the local username update to succeed when the authoritative divine-name-server claim fails. In other words, the UI can now claim verification for an identifier that may not resolve at all.

Please split this back apart: keep the username normalization / boundary enforcement and the discovery response work, but remove the synthesized `nip05` + `Verified` badge until we have a real source of truth for verification. If we want a verified badge later, it needs to be backed by an authoritative claim result or a real resolution check that confirms the returned pubkey matches the user’s pubkey.

sergey3bv:
@NotThatKindOfDrLiz, the changes you requiring are out of the scope of the current PR. Because of the amount of PRs (sorry for that) and poor workflow of stacked PRs on GitHub, I had to bring changes from another branch here, but they are not related to the issue: add limit to a name. Could we discuss somewhere how should follow-ups be solved?

NotThatKindOfDrLiz:
Superseded by #146. This old branch was stacked on pre-#72 NIP-05 work that is already merged; #146 is the clean recut of the one unique follow-up commit on top of current main.

PR #100, opened 2026-04-21
Add AWS KMS support and document KMS provider selection

### LINKED ISSUE #97: Document KMS_PROVIDER and AWS KMS deployment configuration
Follow-up from PR #70. cc @sergey3bv

PR #70 introduces `KMS_PROVIDER=file|gcp|aws` and AWS KMS support, but some docs still describe only `USE_GCP_KMS`.

Update docs to cover:
- `KMS_PROVIDER=file|gcp|aws`
- backward compatibility behavior for `USE_GCP_KMS`
- `AWS_KMS_KEY_ID`
- `AWS_REGION`
- build requirement: `--features aws`
- required AWS IAM permissions for KMS encrypt/decrypt
- migration guidance from `USE_GCP_KMS` to `KMS_PROVIDER`

Known stale reference:
- `CLAUDE.md` still says KMS supports file or GCP via `USE_GCP_KMS=true`.

sergey3bv:
Should close #97

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,14 +19,24 @@ ALLOWED_PUBKEYS=
 # Master encryption key path (generated via bun run key:generate)
 MASTER_KEY_PATH=./master.key
 
-# Use Google Cloud KMS instead of local file-based encryption
+# KMS provider selection: file, gcp, aws
+# Default: file
+KMS_PROVIDER=file
+
+# Legacy fallback (backward compatibility):
+# If KMS_PROVIDER is not set, USE_GCP_KMS=true selects gcp, otherwise file.
 USE_GCP_KMS=false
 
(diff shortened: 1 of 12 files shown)


NotThatKindOfDrLiz (approved):
Re-reviewed after the formatting fix and the rerun. I checked the actual branch code, not just the docs updates: the `KMS_PROVIDER=file|gcp|aws` selection path is wired through startup validation and key-manager construction, the AWS-gated build path compiles locally, and the PR checks are now green.

I did file one non-blocking follow-up: #118 to tighten CI so the AWS-featured KMS tests are executed explicitly rather than only compiled via `clippy --all-features`. That is worth fixing, but it is not a blocker for this PR.

Approving.

PR #101, opened 2026-04-21
Documenting no-name response shape change

### LINKED ISSUE #79: Follow-up: document nostr.json no-name response shape change
Follow-up from PR #72.

When `/.well-known/nostr.json` is requested without a `name` query, the response now includes `names: {}` alongside the existing `nip46` discovery data.

This is probably safe because JSON consumers should ignore unknown fields, but it is still a response-shape change. Please document it in the PR body, release notes, or API docs so downstream consumers are not surprised.

sergey3bv:
Should close #79

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
 - **Relay Architecture**: Refactored from per-user subscriptions to single global subscription with in-memory filtering by bunker pubkey.
 - **Landing Page Content**: Simplified above-the-fold messaging to focus on ease of use rather than technical details. Technical content moved to `/docs`.
 - **Landing Page Links**: Updated all example client links to use https://login.divine.video instead of localhost.
+- **NIP-05 Discovery Response Shape (PR #72)**: Documented that `GET /.well-known/nostr.json` without a `name` query returns `names: {}` alongside existing `nip46` discovery fields. This is a non-breaking additive response-field change for clients that ignore unknown fields.
 
 ### Performance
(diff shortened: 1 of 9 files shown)


NotThatKindOfDrLiz:
Closing in favor of a docs-only replacement PR for issue #79. Follow-up issues will track the NIP-05 state model, dashboard UI states, and discovery semantics separately.

NotThatKindOfDrLiz:
This PR mixes the docs follow-up from #79 with a separate NIP-05/profile/UI feature. Please close this PR and replace it with a docs-only PR limited to documenting the already-shipped no-`name` response shape for `/.well-known/nostr.json`.\n\nThe replacement PR should only document that `GET /.well-known/nostr.json` without `name` returns `names: {}` plus `nip46`. It should not change backend behavior, profile responses, or dashboard UI.\n\nThe NIP-05 badge/profile changes need follow-up design work before implementation. In particular, we need a clear source of truth for claimed vs verified NIP-05 instead of treating a local username as verified.

sergey3bv:
> This PR mixes the docs follow-up from #79 with a separate NIP-05/profile/UI feature.

No, it is not. As a follow-up this PR requires changes from the original branch, that is why it features original `NIP-05 commit`. If this way of handling follow-ups does not suit you, could you please explain how to properly do it? I can open PR-s in my fork, so diff will include only follow-up related changes.

sergey3bv:
@NotThatKindOfDrLiz, I can recreate this PR in my fork to separate its diff from the #72. Would it suit you more? Alternatively we can leave this PR in draft until underling one is merged.

PR #109, opened 2026-04-22
Adding tests for NIP-05 domain resolution

### LINKED ISSUE #75: test(nip05): add tests for domain resolution
Follow-up from PR #72.

sergey3bv:
Should close #75

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -31,6 +31,7 @@ use sqlx::PgPool;
 const DEFAULT_TOKEN_EXPIRY_HOURS: i64 = 24;
 pub const EMAIL_VERIFICATION_EXPIRY_HOURS: i64 = 24;
 const PASSWORD_RESET_EXPIRY_HOURS: i64 = 1;
+const DEFAULT_NIP05_DOMAIN: &str = "divine.video";
 
 /// Get token expiry in seconds. Uses `TOKEN_EXPIRY_SECONDS` env var if set,
 /// otherwise defaults to 24 hours (86400 seconds).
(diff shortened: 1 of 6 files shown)


api/src/api/http/auth.rs
```diff
@@ -50,6 +51,45 @@ pub fn generate_secure_token() -> String {
         .collect()
 }
 
+fn resolve_nip05_domain(tenant_domain: &str) -> String {
+    std::env::var("NIP05_DOMAIN")
```
NotThatKindOfDrLiz, line 55: This precedence is backwards for public tenants. If `tenant.example` is serving the request, the advertised NIP-05 domain needs to remain `tenant.example`; otherwise the UI badge and profile output can point at a domain this tenant does not actually serve `/.well-known/nostr.json` for. The env override should only apply for local-development hosts like `localhost` / `127.0.0.1`.

api/tests/update_profile_divine_name_test.rs
```diff
@@ -35,3 +35,44 @@ async fn update_profile_claims_name_without_enabling_atproto() {
...
+    .execute(&pool)
+    .await
+    .expect("failed to insert second user");
+
+    repo.update_username(&first_user, &username, tenant_id)
+        .await
+        .expect("failed to set first username");
+
+    let available_for_second = repo
+        .check_username_available(&username, &second_user, tenant_id)
```
NotThatKindOfDrLiz, line 74: This only proves the exact same username string is unavailable. It does not cover the real integrity problem introduced by this PR: a mixed-case existing row such as `Alice` versus a new normalized claim for `alice`. If username normalization is landing here, this needs the corresponding case-insensitive repo/DB fix and a regression test for that mixed-case collision.

NotThatKindOfDrLiz (changes requested):
Thanks for the update. I re-reviewed the branch head and this is not ready to merge.

Blocking issues:

1. `resolve_nip05_domain()` has the wrong precedence for public tenants. This implementation prefers `NIP05_DOMAIN` / `DOMAIN` over the tenant host for all requests, and only falls back to the tenant domain if those env vars are unset. For public multi-tenant hosts, that can advertise `user@other-domain` while the actual `/.well-known/nostr.json` is being served from the tenant host. The precedence needs to be: tenant domain for public hosts, env override only for `localhost` / `127.0.0.1` development cases.

2. This PR lowercases usernames on write, but it does not add the repository and DB-side case-insensitive integrity that makes that safe. `find_pubkey_by_username` and `check_username_available` are still case-sensitive on `main`, and there is no migration here to enforce uniqueness on `LOWER(username)` scoped by tenant. As written, existing mixed-case rows can still lead to incorrect availability and NIP-05 resolution behavior.

3. The new test coverage does not cover the mixed-case conflict that matters here. The added repository test only checks the exact same casing. If this PR is going to touch username normalization, it needs the full case-insensitive fix set and a regression test for an existing mixed-case claim blocking a lowercase update.

Process items before this comes back for review:

- Update the PR title/description to match the actual scope. This is not tests-only; it changes auth behavior, API responses, UI, and e2e behavior.
- After the blockers above are fixed, make sure the fork workflow is approved so CI actually runs before re-requesting review.

Please address the blockers above before I re-review.

NotThatKindOfDrLiz:
Superseded by #147. This old branch was stacked on pre-#72 NIP-05 work that is already merged; #147 is the clean recut of the one unique follow-up commit on top of current main.

PR #111, opened 2026-04-22
fix(email): make SES display names safe

### LINKED ISSUE #107: fix(email): escape SES display names derived from FROM_NAME
PR #71 adds AWS SES support, but the SES sender still constructs the From header with raw display-name interpolation.

Current behavior:
- `api/src/email_service.rs` builds the sender string as `"{name} <{email}>"` using `FROM_NAME` directly.

Risk:
- Display names need RFC 5322-safe escaping/encoding.
- Raw interpolation can produce malformed headers or incorrect sender formatting for quotes, commas, or other special characters.

Requested follow-up:
- Implement proper display-name escaping/encoding before constructing the SES from address.
- Add focused tests for representative special-character cases.

Relevant code:
- `api/src/email_service.rs:575-600`

This is non-blocking for PR #71 but should be fixed before relying on arbitrary configured display names.

sergey3bv:
Should close #107

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,7 @@ COPY ./database/migrations ./database/migrations
 COPY ./Cargo.toml ./Cargo.toml
 COPY ./Cargo.lock ./Cargo.lock
 
-RUN cargo build --release --bin keycast
+RUN cargo build --release --features aws --bin keycast
 RUN cargo build --release --example migrate-vine-users
 
 # Build stage for Bun frontend
--- a/README.md
+++ b/README.md
@@ -190,15 +190,19 @@ See [DEVELOPMENT.md](./docs/DEVELOPMENT.md) for local development setup.
 | `SERVER_NSEC` | Server's Nostr secret key for signing tokens |
 | `MASTER_KEY_PATH` | Path to encryption key file |
 
-#### Email (SendGrid)
+#### Email (SendGrid/SES/Dev)
 
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `SENDGRID_API_KEY` | *(none)* | If set, uses SendGrid; otherwise logs emails to console |
+| `EMAIL_PROVIDER` | auto-detect | Explicit provider: `sendgrid`, `ses`, or `dev` (`ses` requires an `aws` feature build) |
+| `SENDGRID_API_KEY` | *(none)* | SendGrid API key (auto-detect uses SendGrid when set) |
+| `AWS_REGION` | AWS SDK default | Region used by SES when `EMAIL_PROVIDER=ses` and built with `--features aws` |
 | `FROM_EMAIL` | `noreply@keycast.app` | Sender email address |
 | `FROM_NAME` | `Divine` | Sender display name |
 | `BASE_URL` | `https://login.divine.video` | Base URL for email verification links |
-| `DISABLE_EMAILS` | *(none)* | If set (any value), skips sending emails |
+| `DISABLE_EMAILS` | *(none)* | If set (any value), skips sending emails and allows startup without a configured provider in production |
+
+If `EMAIL_PROVIDER` is explicitly set to `ses` in a non-`aws` build, startup fails with a configuration error instead of falling back to development email logging.
 
 #### Authentication & OAuth
 
--- a/api/src/email_service.rs
+++ b/api/src/email_service.rs
@@ -2,9 +2,18 @@
 // ABOUTME: Supports SendGrid for production and DevEmailSender for local development/testing
 
 use async_trait::async_trait;
+#[cfg(feature = "aws")]
+use aws_sdk_sesv2::types::{
+    Body as SesBody, Content as SesContent, Destination as SesDestination,
+    EmailContent as SesEmailContent, Message as SesMessage,
+};
+#[cfg(feature = "aws")]
+use aws_sdk_sesv2::Client as SesClient;
 use serde::Serialize;
 use std::env;
 use std::sync::{Arc, Mutex};
+#[cfg(feature = "aws")]
+use tokio::sync::OnceCell;
 
 use crate::brand::BRAND_NAME;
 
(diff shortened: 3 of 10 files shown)


Dockerfile
```diff
@@ -17,7 +17,7 @@ COPY ./database/migrations ./database/migrations
 COPY ./Cargo.toml ./Cargo.toml
 COPY ./Cargo.lock ./Cargo.lock
 
-RUN cargo build --release --bin keycast
+RUN cargo build --release --features aws --bin keycast
```
NotThatKindOfDrLiz, line 20: This changes the shipped production artifact to always include the `aws` feature. That may be the right outcome, but it also means this PR is not just a narrow `FROM_NAME` escaping fix. Please retitle/rescope the PR description to match the actual change set, or split the display-name fix from the broader SES provider rollout.

README.md
```diff
@@ -190,16 +190,20 @@ See [DEVELOPMENT.md](./docs/DEVELOPMENT.md) for local development setup.
...
 | Variable | Default | Description |
 |----------|---------|-------------|
-| `SENDGRID_API_KEY` | *(none)* | If set, uses SendGrid; otherwise logs emails to console |
+| `EMAIL_PROVIDER` | auto-detect | Explicit provider: `sendgrid`, `ses`, or `dev` (`ses` requires an `aws` feature build) |
+| `SENDGRID_API_KEY` | *(none)* | SendGrid API key (auto-detect uses SendGrid when set) |
+| `AWS_REGION` | AWS SDK default | Region used by SES when `EMAIL_PROVIDER=ses` and built with `--features aws` |
 | `FROM_EMAIL` | `noreply@keycast.app` | Sender email address |
 | `FROM_NAME` | `Divine` | Sender display name |
 | `BASE_URL` | `https://login.divine.video` | Base URL for email verification links |
 | `DISABLE_EMAILS` | *(none)* | If set (any value), skips sending emails |
```
NotThatKindOfDrLiz: This docs line does not match the current startup validation semantics. The runtime skips sending when `DISABLE_EMAILS` is present, but `create_email_sender()` only allows production fallback when the value is exactly `"true"`. I filed follow-up #120 for this, but the docs/runtime contract needs to be kept aligned.

NotThatKindOfDrLiz (changes requested):
Thanks for the update. I re-reviewed the branch head, and I am requesting changes before approval.

The main issue here is scope and process clarity, not a newly-discovered correctness failure in the SES display-name escaping itself.

Blocking items:

1. The PR is materially mis-scoped. The title/body say this is about making the SES display name safe / closing #107, but the branch also adds the SES provider implementation, Cargo feature wiring, Docker build changes, startup validation changes, CI coverage changes, and docs/config updates. That is effectively a broader SES provider rollout, not a narrow follow-up. Please either retitle/rewrite the PR description to match the actual shipped changes, or split the display-name fix into its own focused PR.

2. CI has not actually run. Only `license/cla` is green and the fork workflow is still blocked. Given the real scope here, I am not going to approve until the workflow is allowed to run and the checks pass.

Non-blocking item already tracked:
- I filed follow-up #120 for the `DISABLE_EMAILS` docs/runtime mismatch. That does not need to block this PR, but it should be cleaned up so operators are not told one thing while startup validation enforces another.

Next steps:
- Fix the PR title/description so it accurately represents the full change set, or split the PR.
- Approve the fork workflow so CI actually runs.
- If CI passes after the scope is made honest, I am likely to approve.

NotThatKindOfDrLiz:
Tagging @dcadenas for review on the remaining SES branch. #71 has been closed as superseded, so this is now the only active SES line to evaluate on substance.

api/src/email_service.rs
```diff
@@ -358,149 +478,424 @@ impl EmailSender for SendGridEmailSender {
...
+                | '^'
+                | '_'
+                | '`'
+                | '{'
+                | '|'
+                | '}'
+                | '~'
+        )
+}
+
```
dcadenas, line 769: `sanitize_display_name` only remaps CR, LF, and TAB to space.

The other C0 controls (NUL, `\x01`-`\x08`, `\x0B`, `\x0C`, `\x0E`-`\x1F`, and DEL `\x7F`) pass through unchanged and land in the quoted-string branch below, which only escapes `"` and `\`. RFC 5322 `qtext` forbids those bytes, so a `FROM_NAME` like `"Foo\0Bar"` produces a header that AWS SES or an intermediary may truncate at the NUL or reject outright.

This is exactly the class of unsafe input issue #107 is asking you to handle.

Strip or space-map every byte with `(ch as u32) < 0x20` except `\t`, `\r`, `\n`, plus `\x7F`, rather than enumerating only three.

api/src/email_service.rs
```diff
@@ -358,149 +478,424 @@ impl EmailSender for SendGridEmailSender {
...
+            if matches!(ch, '\r' | '\n' | '\t') {
+                ' '
+            } else {
+                ch
+            }
+        })
+        .collect::<String>()
+        .trim()
+        .to_string()
+}
```
dcadenas, line 791: `format_ses_display_name` emits a single `=?UTF-8?B?<base64>?=` encoded-word for any non-ASCII input, with no length bound.

RFC 2047 §2 caps an encoded-word at 75 octets total. The `=?UTF-8?B?` + `?=` wrappers are 13 chars plus base64 of `4*ceil(n/3)` of the UTF-8 byte length, so a display name of roughly 45+ UTF-8 bytes (long Cyrillic, CJK, accented, or emoji-heavy names) overruns the limit. Strict MUAs, relays, and the SES API may reject or mishandle an oversized encoded-word.

Chunk into multiple whitespace-separated encoded-words each capped at 75 chars, splitting on UTF-8 char boundaries rather than raw bytes.

dcadenas (changes requested):
Needs changes before merge.

The sanitizer still lets most C0 control bytes through, which can produce an RFC 5322-invalid header that SES or an intermediary may truncate or reject.

The encoded-word path also has no length cap, so a long non-ASCII name overruns the 75-octet limit in RFC 2047.

Both need fixing before this lands.

sergey3bv:
Hey, @NotThatKindOfDrLiz, @dcadenas, I updated the PR, could you please take a look?

NotThatKindOfDrLiz:
@dcadenas Can you please review (and approve/merge, if appropriate)? TYSM!

api/src/email_service.rs
```diff
@@ -416,109 +667,294 @@ impl EmailSender for SendGridEmailSender {
...
-        self.send_email(to_email, &subject, &html_content, &text_content)
-            .await
+fn emails_disabled() -> bool {
+    env::var("DISABLE_EMAILS").is_ok()
+}
+
+#[cfg(feature = "aws")]
+fn is_atext_char(ch: char) -> bool {
+    ch.is_ascii_alphanumeric()
+        || matches!(
```
dcadenas, line 747: Strip every ASCII control character here, not just CR/LF/TAB. Inputs like `FROM_NAME="Team\x00Ops"` keep the NUL byte through, then fail the atext check and end up as raw control bytes inside the quoted-string branch at line ~789, which only escapes `"` and `\`. RFC 5322 forbids bare CTLs in qtext, so the resulting `From:` header is malformed and SES or downstream MTAs can reject the send. The fix is one line: replace any `ch.is_ascii_control()` with a space (this already subsumes CR/LF/TAB). Please add a test covering this input class.

api/src/email_service.rs
```diff
@@ -416,109 +667,294 @@ impl EmailSender for SendGridEmailSender {
...
+                | '`'
+                | '{'
+                | '|'
+                | '}'
+                | '~'
+        )
+}
+
+#[cfg(feature = "aws")]
+fn sanitize_display_name(display_name: &str) -> String {
```
dcadenas, line 774: Cap the display name length here. A multi-hundred-byte `FROM_NAME` produces an encoded-word above the 75-octet RFC 2047 limit on the non-ASCII branch (line 778), and a multi-kilobyte quoted string above the RFC 5322 998-octet line limit on the ASCII branch (line 786 onward). Neither branch folds or splits, so the failure surfaces at the SES API rather than at startup. `FROM_NAME` is operator-controlled today, so this is not urgent, but if it is ever populated from configured data the send breaks. Either cap `sanitize_display_name` output (for example 200 bytes), or for the non-ASCII branch split into multiple encoded-words per RFC 2047 section 5.

dcadenas (changes requested):
The display-name handling has the right shape, but two RFC-correctness gaps remain in `format_ses_display_name`. Both need to be fixed before merge.

sergey3bv:
Hey, @NotThatKindOfDrLiz and @dcadenas, I updated the PR, could you please take a look? I see there are many pitfalls with special characters and email standards in general, so it begs the question for me: should we continue implementing it by ourself or switch to a 3d party library for that.

dcadenas:
Will implement later

PR #113, opened 2026-04-22
Optimizing RPC cache-hit DPoP enforcement by caching cnf.jkt

### LINKED ISSUE #89: Follow-up: cache DPoP cnf.jkt with OAuth handler cache entries
Follow-up from PR #67.

On OAuth handler cache hit, DPoP enforcement re-parses the UCAN via `validate_ucan_token()` to check `cnf.jkt`. This is correct for security but reduces the cache benefit because UCAN validation includes signature verification.

Consider caching the `cnf.jkt` value alongside the handler/cache entry so DPoP-bound cache hits can avoid re-parsing the UCAN while preserving enforcement.

sergey3bv:
Should close #89

### DIFF
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -245,6 +245,7 @@ async fn load_handler_on_demand(
     pool: &sqlx::PgPool,
     bunker_pubkey_hex: &str,
     tenant_id: i64,
+    dpop_cnf_jkt: Option<String>,
 ) -> Result<Arc<HttpRpcHandler>, RpcError> {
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Re-reviewed after CI passed. I do not see a blocking issue in the cache-hit DPoP optimization as implemented: the slow path still validates UCAN + DPoP before caching, and cache hits now enforce the cached `cnf.jkt` binding directly. The remaining gap is integration coverage depth for a real cache-miss -> cache-hit DPoP-bound request flow, which is already tracked in follow-up #121. Approving.

PR #114, opened 2026-04-22
feat(oauth): accept RFC 6761 .localhost subdomains as loopback

mbradley:
## Summary

Extend `extract_origin` (oauth.rs) and `validate_origin` (auth.rs) to treat any `*.localhost` hostname as loopback in addition to the literal `localhost`, `127.0.0.1`, and `[::1]` strings.

## Why

RFC 6761 reserves `localhost.` and every `*.localhost.` subdomain as loopback. Browsers and modern OS resolvers honor this, so a request to `http://admin.localhost:8787` resolves to `127.0.0.1` with zero DNS or `/etc/hosts` configuration. Today Keycast rejects these with `HTTPS required for non-localhost redirect_uri`, which forces local-dev operators to either fall back to the bare `localhost` (colliding when multiple services want the same loopback name) or edit `/etc/hosts` for each service.

Accepting `.localhost` subdomains lets multiple local dev services coexist on distinct hostnames (`admin.localhost` for an admin UI, `api.localhost` for a public API, etc.) without losing the HTTPS requirement for real public hostnames. The match requires a **non-empty label before `.localhost`**, so substring lookalikes like `xxxlocalhost` or `localhost.evil.com` and the empty-label form `.localhost` are still rejected.

## Discovery context

Surfaced while bringing up a full local Keycast + Cloudflare Worker stack for `divine-name-server`, where the worker's admin UI lives at `admin.localhost:8787` and its public UI at `localhost:8787`. `divine-name-server`'s existing behavior of distinguishing admin vs public UI by hostname made the bare-localhost workaround structurally awkward. The same pattern is expected in other Divine apps (`divine-resurrection`, `divine-invite-darshan`) as they adopt the Keycast OAuth flow — this change stops every one of them from having to patch Keycast locally.

## What changed

| File | Change |
|------|--------|
| `api/src/api/http/oauth.rs` | `extract_origin` accepts `.localhost` subdomains with a non-empty label |
| `api/src/api/http/auth.rs` | `validate_origin` same treatment |

Both checks preserve the existing HTTPS-required semantics for any other hostname.

## Tests

- **`extract_origin`**: new `test_extract_origin_http_localhost_subdomain` covers `admin.localhost:8787`, `api.localhost:3000`, and nested `admin.app.localhost:8080`. New `test_extract_origin_http_localhost_lookalike_rejected` covers `xxxlocalhost`, `localhost.evil.com`, `.localhost`.
- **`validate_origin`**: added a plain unit-test block (the existing one was gated on `integration-tests`), covering HTTPS, `localhost`, `127.0.0.1`, `.localhost` subdomains, and rejection of lookalikes.
- `cargo test -p keycast_api --lib`: **96/96 pass**.

## Test plan

- [x] `cargo test -p keycast_api --lib` — 96/96
- [x] Manual end-to-end: local Keycast + divine-name-server wrangler dev against `http://admin.localhost:8787`, OAuth round-trip succeeds (previously blocked by this check)
- [ ] CI

## Notes for reviewers

- No behavior change for any real public hostname: `example.com`, `foo.bar.example.com`, etc. still require HTTPS.
- No behavior change for `localhost`, `127.0.0.1`, `[::1]`, `::1` — they continue to be accepted.
- The label-length check (`host.len() > ".localhost".len()`) exists specifically to reject `.localhost` with an empty prefix, which some parsers will accept.

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -1059,8 +1059,13 @@ fn validate_origin(origin: &str) -> Result<(), AuthError> {
         .host_str()
         .ok_or_else(|| AuthError::BadRequest("Origin must have a host".to_string()))?;
 
-    // Allow http:// only for localhost (development)
-    let is_localhost = host == "localhost" || host == "127.0.0.1";
+    // Allow http:// only for localhost (development). Accepts RFC 6761
+    // `.localhost` subdomains so multiple local dev services can coexist
+    // on distinct hostnames. The subdomain match requires a non-empty label
+    // before `.localhost`.
+    let is_localhost = host == "localhost"
+        || host == "127.0.0.1"
+        || (host.ends_with(".localhost") && host.len() > ".localhost".len());
     if url.scheme() != "https" && !is_localhost {
         return Err(AuthError::BadRequest("Origin must be HTTPS".to_string()));
     }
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
Approved. The widened loopback predicate is conservative and the tests cover the main accept/reject cases.

mbradley:
Follow-up commit `d626ee3` fixes the lone CI failure: a `cargo fmt` whitespace diff on the `CreateBunkerRequest` comment alignment (`api/src/api/http/auth.rs:1037`). No behavioral change — tests, clippy, and CLA were already green. Landing now.

PR #125, opened 2026-04-24
Adding integration coverage for cache-hit DPoP enforcement in HTTP RPC

### LINKED ISSUE #121: Follow-up: add integration coverage for cache-hit DPoP enforcement in HTTP RPC
PR #113 moves cache-hit DPoP enforcement from reparsing the UCAN on every hit to using cached `cnf.jkt` state stored on `HttpRpcHandler`.

Current coverage:
- The PR adds focused unit tests for `enforce_cached_dpop_binding()` and the `expected_dpop_jkt()` accessor.
- The integration test file only gets the constructor signature updated; it does not exercise a real cache-miss followed by cache-hit flow with a DPoP-bound token.

Requested follow-up:
- Add an integration test that performs two authenticated RPC requests with the same DPoP-bound UCAN:
  1. first request populates the cache
  2. second request hits the cache
- Verify that the cached path still rejects missing/invalid DPoP proof and accepts a valid proof bound to the expected `cnf.jkt`.

Relevant code:
- `api/src/api/http/nostr_rpc.rs`
- `api/src/handlers/http_rpc_handler.rs`
- `api/tests/nostr_rpc_integration_test.rs`

This is non-blocking for PR #113, but it would give much better confidence in a security-sensitive optimization.

sergey3bv:
Should close #121

### DIFF
--- a/api/tests/nostr_rpc_integration_test.rs
+++ b/api/tests/nostr_rpc_integration_test.rs
@@ -5,16 +5,35 @@
 
 mod common;
 
+use axum::{extract::State, http::HeaderMap, Json};
+use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
 use chrono::{Duration, Utc};
+use keycast_api::api::{
+    http::{
+        auth::AuthError,
+        nostr_rpc::{nostr_rpc, NostrRpcRequest, NostrRpcResponse, RpcError},
+        routes::AuthState,
+    },
+    tenant::{Tenant, TenantExtractor},
+};
+use keycast_api::bcrypt_queue::BcryptQueue;
 use keycast_api::handlers::http_rpc_handler::{new_http_handler_cache, HttpRpcHandler};
+use keycast_api::state::KeycastState;
+use keycast_api::ucan_auth::{nostr_pubkey_to_did, NostrKeyMaterial};
 use keycast_core::encryption::file_key_manager::FileKeyManager;
 use keycast_core::encryption::KeyManager;
+use keycast_core::secret_pool::SecretPool;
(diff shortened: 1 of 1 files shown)


api/tests/nostr_rpc_integration_test.rs
```diff
@@ -567,3 +767,139 @@ async fn test_sign_event_blocked_by_policy() {
...
+    .await;
+
+    let valid_dpop_key = SigningKey::random(&mut OsRng);
+    let expected_jkt = dpop_thumbprint(&valid_dpop_key);
+    let token = build_dpop_bound_ucan(
+        &user_keys,
+        tenant_id,
+        "dpop-cache-hit@example.com",
+        &redirect_origin,
+        &bunker_pubkey,
```
NotThatKindOfDrLiz, line 817: This test closes the main end-to-end gap from #121 and the four-request flow is the right shape. One non-blocking improvement for a follow-up: assert that requests 2-4 really used the cache-hit path, not just the same behavioral outcome. The simplest way is probably to snapshot `http_handler_cache.entry_count()` after request 1 and verify it does not grow on the later requests.

NotThatKindOfDrLiz (approved):
Reviewed the integration test against the production HTTP RPC DPoP/cache behavior. This closes the missing end-to-end coverage gap from #121: the test drives a DPoP-bound UCAN through cache miss success, cache-hit missing-proof rejection, cache-hit wrong-key rejection, and cache-hit valid-proof success. CI is green and I do not see a blocking issue.

I did file one non-blocking follow-up to tighten the test further so it explicitly proves requests 2-4 used the cached path rather than merely reproducing the same behavioral outcome through the slow path.

Approving.

PR #126, opened 2026-04-24
Adding use distributed Redis-backed DPoP replay cache

### LINKED ISSUE #88: Follow-up: use distributed DPoP JTI replay cache
Follow-up from PR #67.

The DPoP JTI replay cache is currently in-process (`DashMap`). In multi-instance deployments such as Cloud Run with multiple minimum instances, a replayed DPoP proof can potentially succeed against a different instance during the freshness window.

The current risk is bounded by DPoP method/URL binding and short proof freshness, so this was accepted for PR #67. Please add a distributed replay cache, likely Redis-backed, so replay detection works across instances.

sergey3bv:
Should close #88

### DIFF
--- a/api/src/api/extractors.rs
+++ b/api/src/api/extractors.rs
@@ -37,7 +37,26 @@ fn extract_admin_role(ucan: &ucan::Ucan) -> Option<String> {
 
 /// Enforce DPoP binding if the UCAN has a cnf.jkt claim.
 /// Constructs the htu from the request method and path.
-fn enforce_dpop_if_bound(parts: &Parts, ucan: &ucan::Ucan) -> Result<(), AuthError> {
+fn map_dpop_error_to_auth_error(error: anyhow::Error, path: &str) -> AuthError {
+    let message = format!("DPoP binding enforcement failed: {}", error);
+    if crate::ucan_auth::is_replay_cache_unavailable_error(&error) {
+        tracing::error!("{} (path: {})", message, path);
+        AuthError {
+            status: StatusCode::SERVICE_UNAVAILABLE,
+            message: "DPoP replay protection temporarily unavailable. Please retry.".to_string(),
+        }
+    } else {
+        tracing::warn!("{} (path: {})", message, path);
+        AuthError {
+            status: StatusCode::UNAUTHORIZED,
+            message,
+        }
+    }
+}
+
+/// Enforce DPoP binding if the UCAN has a cnf.jkt claim.
+/// Constructs the htu from the request method and path.
+async fn enforce_dpop_if_bound(parts: &Parts, ucan: &ucan::Ucan) -> Result<(), AuthError> {
     let cnf_jkt = crate::ucan_auth::extract_cnf_jkt_from_ucan(ucan);
     if cnf_jkt.is_none() {
         return Ok(()); // No DPoP binding, nothing to enforce
@@ -57,14 +76,9 @@ fn enforce_dpop_if_bound(parts: &Parts, ucan: &ucan::Ucan) -> Result<(), AuthErr
     let path = parts.uri.path();
     let htu = format!("{}://{}{}", scheme, host, path);
 
-    crate::ucan_auth::enforce_dpop_binding(&parts.headers, ucan, method, &htu).map_err(|e| {
-        let msg = format!("DPoP binding enforcement failed: {}", e);
-        tracing::warn!("{} (path: {})", msg, parts.uri.path());
-        AuthError {
-            status: StatusCode::UNAUTHORIZED,
-            message: msg,
-        }
-    })
+    crate::ucan_auth::enforce_dpop_binding(&parts.headers, ucan, method, &htu)
+        .await
+        .map_err(|e| map_dpop_error_to_auth_error(e, parts.uri.path()))
 }
 
 /// Resolve tenant_id from the Host header using the tenant cache/database
@@ -143,7 +157,7 @@ where
                             })?;
 
                     // Enforce DPoP binding if UCAN contains cnf.jkt
-                    enforce_dpop_if_bound(parts, &ucan)?;
+                    enforce_dpop_if_bound(parts, &ucan).await?;
 
                     let admin_role = extract_admin_role(&ucan);
 
@@ -198,5 +212,28 @@ where
                 "Missing authentication - expected UCAN Bearer token or keycast_session cookie"
                     .to_string(),
         })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_map_dpop_error_to_auth_error_replay_cache_unavailable_is_503() {
+        let error = anyhow::Error::new(crate::ucan_auth::dpop::ReplayCacheUnavailableError::new(
+            "DPoP replay protection unavailable",
+        ));
+        let auth_error = map_dpop_error_to_auth_error(error, "/api/test");
+        assert_eq!(auth_error.status, StatusCode::SERVICE_UNAVAILABLE);
+        assert!(auth_error.message.contains("temporarily unavailable"));
+    }
+
+    #[test]
+    fn test_map_dpop_error_to_auth_error_invalid_proof_is_401() {
+        let auth_error =
+            map_dpop_error_to_auth_error(anyhow::anyhow!("invalid dpop proof"), "/api/test");
+        assert_eq!(auth_error.status, StatusCode::UNAUTHORIZED);
+        assert!(auth_error.message.contains("invalid dpop proof"));
     }
 }
--- a/api/src/ucan_auth/dpop.rs
+++ b/api/src/ucan_auth/dpop.rs
@@ -4,22 +4,44 @@
 use anyhow::{anyhow, Result};
 use axum::http::HeaderMap;
 use base64::{engine::general_purpose::URL_SAFE_NO_PAD, Engine as _};
+use dashmap::mapref::entry::Entry;
 use dashmap::DashMap;
 use once_cell::sync::Lazy;
 use p256::ecdsa::{signature::Verifier, Signature as P256Signature, VerifyingKey};
 use p256::EncodedPoint;
 use serde_json::Value;
 use std::time::{Duration, Instant};
+use thiserror::Error;
 
 /// Maximum age of a DPoP proof (5 minutes)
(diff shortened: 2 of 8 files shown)


api/src/ucan_auth/dpop.rs
```diff
@@ -43,6 +43,42 @@ fn maybe_cleanup_jtis() {
     }
 }
 
+fn replay_cache_key(thumbprint: &str, jti: &str) -> String {
+    format!("dpop:jti:{thumbprint}:{jti}")
```
NotThatKindOfDrLiz, line 71: **Behavior change worth a security read.** Pre-PR the in-memory cache was keyed by `jti` alone, so any reuse of a `jti` (across keys, across clients) was rejected. The new cache key is `dpop:jti:{thumbprint}:{jti}`, scoping uniqueness to *per-key*. The new test `test_same_jti_with_different_keys_is_allowed` pins this. Per RFC 9449 §11.1 this is acceptable (the spec describes uniqueness 'with the same DPoP-bound access token (or the same DPoP key, where applicable)'), but the prior stricter scope is gone. Confirm this is the intended threat model.

Also: no tenant prefix in this key. `PrefixedRedis` adds `REDIS_KEY_PREFIX` (per-deployment), but tenants in the same deployment share the same prefix space. Cross-tenant collision would require obtaining the same DPoP key — cryptographically improbable — so this is probably fine, but worth confirming explicitly.

api/src/ucan_auth/dpop.rs
```diff
@@ -43,6 +43,42 @@ fn maybe_cleanup_jtis() {
...
+}
+
+async fn record_jti_usage(replay_key: &str, iat: i64) -> Result<()> {
+    if let Some(redis) = crate::state::get_keycast_state()
+        .ok()
+        .and_then(|state| state.redis.clone())
+    {
+        match redis
+            .set_nx_ex(replay_key, DPOP_MAX_AGE_SECS, &iat.to_string())
```
NotThatKindOfDrLiz, line 152: **TTL ↔ proof-lifetime invariant.** The Redis entry expires after `DPOP_MAX_AGE_SECS` (currently 300). If a proof's effective lifetime ever exceeds this TTL — e.g., if the iat freshness check accepts older proofs, or if clock skew tolerance is wider than 300s — replay becomes possible after the entry expires. Worth confirming that `DPOP_MAX_AGE_SECS` is the maximum-allowable iat-to-now window everywhere, not just here.

Minor: the cache value stored is `iat.to_string()` but we never read it back (only the existence of the key matters for the replay check). It's effectively decorative metadata. That's fine, just calling it out.

api/src/ucan_auth/dpop.rs
```diff
@@ -43,6 +43,42 @@ fn maybe_cleanup_jtis() {
...
+            .await
+        {
+            Ok(true) => return Ok(()),
+            Ok(false) => return Err(anyhow!("DPoP proof JTI has already been used (replay)")),
+            Err(error) => {
+                tracing::warn!(
+                    "Redis replay cache unavailable for DPoP, using local fallback: {}",
```
NotThatKindOfDrLiz: **Biggest decision point in the PR.** When Redis returns an error, this falls back to the local in-memory cache (continues into `maybe_cleanup_jtis()` etc. below). In a multi-instance deployment that means a Redis outage silently degrades replay protection to per-pod — each instance becomes its own island, and the same DPoP proof can be replayed against each pod within the proof's freshness window.

Threat model question: how bad is the worst-case replay window during a Redis outage vs. how bad is the availability hit of failing closed (rejecting all DPoP-bound requests when Redis is unreachable)?

The `tracing::warn!` makes this visible in logs but doesn't gate the request. If fail-closed is preferred for security-sensitive endpoints, this is the line to change. Either way, please weigh in — this is the load-bearing tradeoff.

api/src/ucan_auth/dpop.rs
```diff
@@ -186,6 +213,10 @@ pub fn verify_dpop_proof(
         }
     }
 
+    // Check replay after full proof validation.
+    let replay_key = replay_cache_key(&thumbprint, &jti);
```
NotThatKindOfDrLiz, line 324: **Lost ordering invariant comment, still valid behavior.** The pre-PR code at this site had a comment: `// Check for JTI replay (inserted AFTER signature verification to prevent cache poisoning)`. The new code does preserve the ordering — replay check runs after `jwk_thumbprint(jwk)?` (line 211), which is after signature verification — so the invariant still holds. But the comment justifying *why* the ordering matters is gone, and an unwitting future refactor could re-introduce the cache-poisoning vulnerability by moving the replay check earlier.

Suggest restoring a one-line comment here: `// Replay check runs after signature/thumbprint validation to prevent unauthenticated cache poisoning.`

NotThatKindOfDrLiz (commented):
@dcadenas — flagging four security-sensitive hotspots inline for your review. The PR moves DPoP replay protection from local in-memory (DashMap) to Redis-backed distributed, with a local fallback. The interesting questions are all around (a) the key-scope change, (b) the fail-open fallback behavior on Redis errors, (c) the TTL/proof-lifetime relationship, and (d) the lost cache-poisoning-prevention ordering comment. Specifics in the inline threads.

api/src/ucan_auth/dpop.rs
```diff
@@ -43,6 +43,42 @@ fn maybe_cleanup_jtis() {
...
+        .ok()
+        .and_then(|state| state.redis.clone())
+    {
+        match redis
+            .set_nx_ex(replay_key, DPOP_MAX_AGE_SECS, &iat.to_string())
+            .await
+        {
+            Ok(true) => return Ok(()),
+            Ok(false) => return Err(anyhow!("DPoP proof JTI has already been used (replay)")),
+            Err(error) => {
```
dcadenas, line 157: Decide explicitly what should happen when Redis is unavailable, and make the degraded mode observable.

Right now any Redis error (timeout, auth failure, Memorystore failover, command failure) is logged at `warn` and silently falls through to the in-process `DashMap`. In a multi-instance deployment (min-instances ≥ 2) that reopens the exact cross-instance replay window this PR was written to close. During a Redis outage each instance tracks JTIs only in its own `DashMap`, so a valid DPoP proof can be replayed against a sibling instance inside the 300s freshness window. An attacker who can induce Redis flakiness, or who times a proof during a Memorystore failover, can exploit that directly, and the only operator signal is a `warn` log.

Please do at least two of these before merge:
- Fail closed on Redis error in production (return the replay-prevention error), gated by an explicit config flag so local/dev can keep the current behavior.
- Emit a dedicated metric such as `dpop_replay_cache_fallback_total`, and/or raise the fallback log to `error`, so operators can tell `replay cache is distributed` from `replay cache has silently degraded to per-instance`.
- Document in the module or the PR body that Redis availability is part of the replay guarantee and that the in-process fallback is best-effort.

This is the highest-severity item in the review.

api/src/ucan_auth/dpop.rs
```diff
@@ -43,6 +43,42 @@ fn maybe_cleanup_jtis() {
...
+                tracing::warn!(
+                    "Redis replay cache unavailable for DPoP, using local fallback: {}",
+                    error
+                );
+            }
+        }
+    }
+
+    maybe_cleanup_jtis();
+    let now = Instant::now();
```
dcadenas, line 166: Restore an atomic check-and-insert on the `DashMap` fallback. The current `JTI_CACHE.get(replay_key)` followed by `JTI_CACHE.insert(...)` is a TOCTOU race. Two concurrent requests with the same replay key during a Redis outage can both observe the entry absent, both insert, and both return `Ok`, silently allowing a replay.

This is a regression from `origin/main`, which relied on the atomic `JTI_CACHE.insert(jti, expiry).is_some()` idiom. Please either keep that contract on the fallback path, or switch to `entry().or_insert_with(...)` and treat "entry already existed" as the replay error. Either form gives you a single atomic step instead of get-then-insert.

dcadenas (changes requested):
Request changes. The distributed replay cache does the right thing on the happy path, but the fallback path has two gaps that need to close before merge.

Both concerns land in `api/src/ucan_auth/dpop.rs` and are left as inline comments.

sergey3bv:
Hey, @NotThatKindOfDrLiz, @dcadenas, I updated the PR to implement changes you requested. Could you please take a look.

NotThatKindOfDrLiz:
@dcadenas Can you please review the updates? TYSM!

api/src/api/extractors.rs
```diff
@@ -143,7 +145,7 @@ where
                             })?;
 
                     // Enforce DPoP binding if UCAN contains cnf.jkt
-                    enforce_dpop_if_bound(parts, &ucan)?;
+                    enforce_dpop_if_bound(parts, &ucan).await?;
```
dcadenas, line 160: Fail-closed on Redis outage returns 401 here. When `record_jti_usage` bails through `handle_redis_replay_cache_error` or `handle_redis_replay_cache_unavailable`, the error bubbles up through `enforce_dpop_binding` and gets mapped to `AuthError { status: UNAUTHORIZED }`.

Before this PR, that path only triggered on real proof validation failures, so 401 was the right answer. Now a transient Redis outage looks identical to `invalid_token` to the client, which is likely to trigger token refresh loops or forced logout every time Redis restarts.

Worth distinguishing infra-induced failures (503 Service Unavailable, or a separate error variant) from genuine auth failures so clients and operators can react differently.

dcadenas (approved):
🔒 distributed replay cache closes a real gap — a couple of non-blocking suggestions on error mapping inline.

---
**Inline comments (outside PR diff):**

- **api/src/api/http/nostr_rpc.rs:371** — Same infra-vs-auth conflation on the RPC path. When `verify_dpop_proof` bails because Redis is unavailable, both `enforce_cached_dpop_binding` here and the cache-miss callsite near line 521 collapse the result to `RpcError::Auth(AuthError::InvalidToken)`.

An RPC client can't distinguish that from a genuinely bad DPoP proof. Consider mapping Redis-unavailable errors from `record_jti_usage` to an `Internal` or `ServiceUnavailable` RPC variant instead of `InvalidToken`.

sergey3bv:
Hey, @NotThatKindOfDrLiz, @dcadenas, I updated the PR, could you please take a look?

dcadenas (approved):
🔒 Redis-backed jti cache closes the multi-instance replay gap, ship it.

Issue #128, opened 2026-04-24
test(signer): add encrypt/decrypt validation coverage


sergey3bv:
TODO at `permission_validation_tests.rs`
https://github.com/divinevideo/keycast/blob/6946664bcd77bbc8624c8cd1836243e102b6c50e/signer/tests/permission_validation_tests.rs#L894

sergey3bv:
Hey, @NotThatKindOfDrLiz, If you don't mind, I will work on this issue

sergey3bv:
@NotThatKindOfDrLiz, I opened [a PR](https://github.com/divinevideo/keycast/pull/131) to solve the issue, could you please take a look?

PR #131, opened 2026-04-24
signer: enforce encrypt/decrypt permissions and add coverage

sergey3bv:
## Summary

  Addresses #128 and fixes a latent signer permission-enforcement gap.

  This PR does two things:
  1. adds the missing encrypt/decrypt permission coverage originally called out in #128
  2. wires the existing `CustomPermission::can_encrypt` / `can_decrypt` methods into the signer's `nip04_*` and `nip44_*` encrypt/decrypt dispatch arms

  On `main`, `validate_permissions_for_sign` exists, but no analogous validation runs before encrypt/decrypt. This PR closes that production gap and adds coverage for the newly enforced behavior.

  ## Behavior change — read this

  This is a real, customer-visible behavior change. Existing custom permissions behave as follows after this PR:

  | Permission        | Pre-PR (encrypt/decrypt) | Post-PR                                     |
  |-------------------|--------------------------|---------------------------------------------|
  | `allowed_kinds`   | not enforced             | unchanged — `can_encrypt`/`can_decrypt` always return `true` |
  | `encrypt_to_self` | not enforced (inert)     | now enforced — restricts encrypt/decrypt to user's own pubkey (intended) |
  | `decrypt_only`    | not enforced (inert)     | now enforced — denies encrypt, allows decrypt (intended) |
  | `content_filter`  | not enforced             | **now blocks encrypt of plaintext containing blocked words** |

  The `content_filter` row is the one to flag for ops/support: customers who configured `content_filter` for sign-time policy may now see encrypt requests denied when the plaintext contains a blocked word. They did not previously opt into encrypt-time blocking.

  - **Backward-compatible default preserved:** if no permissions are linked, encrypt/decrypt continue to succeed.

  ## Implementation

  - `validate_permissions_for_encrypt` and `validate_permissions_for_decrypt` added to `Nip46Handler`, mirroring `validate_permissions_for_sign`. Marked `pub + #[doc(hidden)]` to follow the existing test-seam convention used by `new_for_test`.
  - All four dispatch arms (`nip04_encrypt`, `nip04_decrypt`, `nip44_encrypt`, `nip44_decrypt`) now call the appropriate validate method before performing crypto. Both arms have identical structure.

  ## Tests

  - `test_16` `decrypt_only` denies encrypt, allows decrypt (team auth path)
  - `test_16b` same coverage on OAuth auth path
  - `test_17` `encrypt_to_self` denies decrypt from third party
  - `test_18` no-permissions fallback for both encrypt and decrypt
  - `test_19` `content_filter` blocked-encrypt + clean-encrypt

  ## Follow-up

  Dispatcher-level end-to-end coverage for `handle_nip46_request` is intentionally out of scope for this PR. Exercising that path requires heavier relay/cache/coordinator test scaffolding. This PR instead covers the shared permission-enforcement logic and the key behavior cases introduced by the new enforcement.

### DIFF
--- a/signer/src/signer_daemon.rs
+++ b/signer/src/signer_daemon.rs
@@ -164,7 +164,8 @@ impl Nip46Handler {
     }
 
     /// Validate permissions before encrypting plaintext for a recipient.
-    async fn validate_permissions_for_encrypt(
+    #[doc(hidden)]
+    pub async fn validate_permissions_for_encrypt(
         &self,
         plaintext: &str,
         recipient_pubkey: &PublicKey,
@@ -208,7 +209,8 @@ impl Nip46Handler {
     }
 
     /// Validate permissions before decrypting ciphertext from a sender.
-    async fn validate_permissions_for_decrypt(
+    #[doc(hidden)]
+    pub async fn validate_permissions_for_decrypt(
         &self,
         ciphertext: &str,
         sender_pubkey: &PublicKey,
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz:
Hi, @sergey3bv. I pushed two maintainer commits on top of yours rather than requesting changes, since this needs to land soon and the work was close to ready:

- refactor(signer): collapses the nip44/nip04 asymmetry, drops the _direct helpers, and makes validate_permissions_for_* the test seam (matching the existing new_for_test convention).
- test(signer): adds OAuth-branch, no-permissions-fallback, and content-filter encrypt-blocking coverage.

I also reframed the title and description to match what the diff actually does. This turned out to be a security-relevant behavior change (encrypt/decrypt was previously not policy-checked), and the new content_filter enforcement in particular needs explicit ops disclosure. Thanks for the original work, and happy to walk through any of the changes.

NotThatKindOfDrLiz (approved):
Looks good. Approving on the merits of the final state: production-enforcement gap closed, four dispatch arms have identical shape, content_filter behavior change is explicitly disclosed, and coverage now includes OAuth branch, no-permissions fallback, and content_filter encrypt-blocking. CI green.

PR #132, opened 2026-04-24
test(signer): regression test for invalid policy_id handling

### LINKED ISSUE #129: Add test for invalid policy_id handling
TODO at `permission_validation_tests.rs`
https://github.com/divinevideo/keycast/blob/6946664bcd77bbc8624c8cd1836243e102b6c50e/signer/tests/permission_validation_tests.rs#L895

sergey3bv:
Closes #129.

Adds a regression test (`test_20_oauth_invalid_policy_id_allows_signing`) that pins the current signer behavior when an OAuth authorization references a `policy_id` that doesn't exist in the `policies` table: `oauth_auth.permissions(...)` returns an empty `Vec<Permission>`, validation falls through the `permissions.is_empty() → Ok(())` backward-compat branch, and signing succeeds.

Rebased onto `main` after #131 landed; renamed from `test_16` to `test_20` to avoid colliding with #131's new test_16/16b.

## Follow-up

Whether this *should* be the behavior is an open question — a dangling FK on `oauth_authorizations.policy_id` silently disables policy enforcement. Tracked separately in #141 (decision needed: fail closed vs. permissive fallback).

This PR pins current behavior as a regression test. It is intentionally not changing that behavior.

### DIFF
--- a/signer/tests/permission_validation_tests.rs
+++ b/signer/tests/permission_validation_tests.rs
@@ -1108,5 +1108,38 @@ async fn test_19_content_filter_blocks_encrypt_of_blocked_plaintext() {
     );
 }
 
-// TODO: Add test for invalid policy_id handling
+/// Regression: an OAuth authorization with a dangling `policy_id` (no matching row in
+/// `policies`) currently degrades to "no permissions = allow". This test pins that behavior.
+/// See divinevideo/keycast#141 for the open question of whether this should fail closed.
+#[tokio::test]
+async fn test_20_oauth_invalid_policy_id_allows_signing() {
+    let pool = setup_test_db().await;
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
Hi @sergey3bv — rebased this onto current `main` since #131 landed and would have collided. Test renamed from `test_16` → `test_20` (next free slot after #131 used through test_19). Also updated the title and description to make it explicit that this PR pins *current* behavior, with the open question of whether dangling `policy_id` should fail closed tracked separately as #141. Thanks for the original work.

NotThatKindOfDrLiz (approved):
Rebased + renamed for the test_16 collision with #131; re-titled and re-described to make explicit that this pins current behavior, with the dangling-policy_id question tracked separately in #141. CI green.

PR #133, opened 2026-04-24
test: add permission loading failure coverage for signer policy validation

### LINKED ISSUE #130: Add test for permission loading failure
TODO at `permission_validation_tests.rs`
https://github.com/divinevideo/keycast/blob/6946664bcd77bbc8624c8cd1836243e102b6c50e/signer/tests/permission_validation_tests.rs#L896

sergey3bv:
Closes #130.

Adds `test_8b_permission_loading_failure_returns_error` to `signer/tests/permission_validation_tests.rs`, filling the TODO that #130 tracked.

## What it covers

Previous tests in this file exercise valid permissions (allow/deny paths) and missing policies, but nothing covered what happens when a stored `permissions.config` row is structurally unparseable. This test:

1. Creates a valid policy and OAuth authorization.
2. Inserts a `permissions` row with malformed JSON in `config` (`{"allowed_kinds": [1]` — missing closing brace) and attaches it to the policy.
3. Calls `sign_event_direct` and asserts:
   - The call returns `Err` (fails closed).
   - The error message points to permission loading/parsing (`Authorization error` / `Database error` / `config`), **not** the policy deny path (asserts the error does *not* contain `Blocked by`).

The negative assertion is the important one: it pins down that a malformed config can't be silently treated as "no permission matched → deny", which would mask data-corruption bugs as ordinary policy denies.

### DIFF
--- a/signer/tests/permission_validation_tests.rs
+++ b/signer/tests/permission_validation_tests.rs
@@ -528,6 +528,71 @@ async fn test_8_oauth_with_policy_enforces_restrictions() {
     assert!(result.is_err(), "OAuth with policy should deny kind 4");
 }
 
+#[tokio::test]
+async fn test_8b_permission_loading_failure_returns_error() {
+    let pool = setup_test_db().await;
+    let key_manager = FileKeyManager::new().expect("Failed to create key manager");
+
+    // Start with a valid policy, then attach a malformed permission config row.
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
Looks good. Rechecked after the rebase: the PR is scoped to the new signer permission-loading failure test, mergeable, and all checks are green. The added coverage is exercising the intended failure path without changing runtime behavior.

PR #146, opened 2026-04-26
auth: enforce max NIP-05 username length boundary

### LINKED ISSUE #80: Follow-up: add NIP-05 username length limit

NotThatKindOfDrLiz:
Closes #80.

This is a clean recut of the actual follow-up work from #99 on top of current `main`.

What it does:
- adds a max length boundary for normalized NIP-05 usernames
- adds unit coverage for the exact boundary and the first-over-boundary case

Why this exists as a fresh PR:
- the original #99 branch was stacked on pre-#72 NIP-05 work that is already merged
- this branch contains only the one unique follow-up commit on top of current `main`

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -33,6 +33,7 @@ const DEFAULT_TOKEN_EXPIRY_HOURS: i64 = 24;
 pub const EMAIL_VERIFICATION_EXPIRY_HOURS: i64 = 24;
 const PASSWORD_RESET_EXPIRY_HOURS: i64 = 1;
 const DEFAULT_NIP05_DOMAIN: &str = "divine.video";
+const MAX_NIP05_USERNAME_LENGTH: usize = 64;
 
 /// Get token expiry in seconds. Uses `TOKEN_EXPIRY_SECONDS` env var if set,
 /// otherwise defaults to 24 hours (86400 seconds).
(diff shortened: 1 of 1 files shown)


api/src/api/http/auth.rs
```diff
@@ -92,6 +93,13 @@ fn normalize_nip05_username(raw_username: &str) -> Result<String, AuthError> {
         ));
     }
 
+    if username.len() > MAX_NIP05_USERNAME_LENGTH {
+        return Err(AuthError::Internal(format!(
```
dcadenas, line 97: The new length check returns `AuthError::Internal`, but that variant's `IntoResponse` impl at auth.rs:376-383 logs at ERROR and sends HTTP 503 with a generic "Service temporarily unavailable" message. A user who posts a 65-character username will get a 503 and never see the "Username must be at most 64 characters" text in the error itself. The other guards in `normalize_nip05_username` already do this, so the new branch is consistent with the surrounding code. But since you're adding a branch here anyway, `AuthError::BadRequest` would fit the semantics better: this is a client input violation, not a server-internal failure, and `BadRequest` would actually deliver the message string to the caller.

dcadenas (approved):
🔒 Good to have the upper bound pinned down on NIP-05 usernames.

PR #147, opened 2026-04-26
test(auth): add NIP-05 domain resolution precedence coverage

### LINKED ISSUE #75: test(nip05): add tests for domain resolution

NotThatKindOfDrLiz:
Closes #75.

This is a clean recut of the actual follow-up work from #109 on top of current `main`.

What it does:
- adds coverage for NIP-05 domain resolution precedence
- verifies tenant/public-host precedence stays locked down after #72

Why this exists as a fresh PR:
- the original #109 branch was stacked on pre-#72 NIP-05 work that is already merged
- this branch contains only the one unique follow-up commit on top of current `main`

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -3583,6 +3583,7 @@ mod tests {
     #[cfg(feature = "integration-tests")]
     use moka::future::Cache;
     use nostr_sdk::{Keys, Kind, Timestamp, UnsignedEvent};
+    use serial_test::serial;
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
Closing this recut because the intended precedence coverage is already on main via the merged NIP-05 work. The remaining diff on this branch is just stale pre-#72 test baggage with obsolete expectations, so there is no clean follow-up left to land here.

PR #67, opened 2026-04-03
security: red team remediation — 8 vulnerability fixes

rabble:
## Summary

Comprehensive security remediation addressing findings from red team assessment. Fixes 8 vulnerability classes across OAuth, tenant isolation, token binding, email configuration, and dependencies.

### Vulnerabilities Fixed

- **OAuth HTML Injection (XSS)**: All user-controlled values in OAuth HTML templates now escaped via shared `html_safety.rs` module (29 interpolation points)
- **Cross-Tenant UCAN Bypass**: Eliminated `tenant_id=0` bypass — all UCAN validation now requires real tenant from Host header
- **Unscoped Database Queries**: Added `find_by_bunker_pubkey_for_tenant` and `find_encrypted_key_for_tenant` — all request-path DB lookups are tenant-scoped
- **DPoP Token Replay**: Created shared DPoP verifier (`dpop.rs`) with ES256 signature verification, JTI replay cache, RFC 9449 htu compliance; enforced on all bearer-token resource access
- **Open OAuth Client Registration**: Production now requires registered OAuth clients (`REQUIRE_REGISTERED_OAUTH_CLIENTS`, defaults true in release)
- **Dev Email in Production**: `create_email_sender()` fails closed — production requires `SENDGRID_API_KEY` or explicit `DISABLE_EMAILS=true`
- **Vulnerable Web Dependencies**: Upgraded vite (6.4.1), svelte (5.55), kit (2.55); disabled Swagger auth persistence
- **RustSec Advisories**: Fixed 6/9 via `cargo update`; remaining 3 documented as upstream-blocked (ucan/did-key chain)

### New Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `ALLOWED_TENANT_DOMAINS` | — | Required in production for tenant isolation |
| `ENABLE_TENANT_AUTO_PROVISIONING` | `false` | Dev-only: auto-create tenants |
| `REQUIRE_REGISTERED_OAUTH_CLIENTS` | `true` (release) | Reject unknown OAuth clients |
| `DISABLE_EMAILS` | — | Set to `true` to skip email in production |

## Test plan

- [x] `cargo check --workspace` — clean
- [x] `cargo test -p keycast_api --lib` — 89 passed, 0 failed
- [x] Zero `validate_ucan_token(..., 0)` in production code (grep verified)
- [x] All HTML interpolations use escape helpers (grep verified)
- [x] EdDSA DPoP proofs rejected until verification implemented
[...]

PR #148, opened 2026-04-26
fix(deploy): set ALLOWED_TENANT_DOMAINS in cloudbuild

irab:
## Summary

Adds `ALLOWED_TENANT_DOMAINS=login.divine.video` to the Cloud Run deploy in `cloudbuild.yaml` so future revisions don't regress account creation.

## Background

Production was returning `Invalid domain: login.divine.video` because the tenant validator (made fail-closed in #67) rejects all domains unless `ALLOWED_TENANT_DOMAINS` or `ENABLE_TENANT_AUTO_PROVISIONING` is set, and neither was being passed to Cloud Run. Already applied as a live hotfix via \`gcloud run services update\`; this PR makes the change survive subsequent deploys.

## Test plan

- [x] Verified hotfix on live service: \`POST https://login.divine.video/api/auth/register\` → 200
- [x] Confirmed env var present on revision \`keycast-00235-8v5\`
- [ ] Cloud Build smoke tests pass on this PR's deploy


dcadenas (approved):
🚪 tenant allowlist wired up in deploy, ship it.

PR #149, opened 2026-04-26
fix(cloudbuild): retry CORS smoke tests + accept Any origin for headless

rabble:
## Summary

Two related fixes for the post-deploy CORS smoke tests in `cloudbuild.yaml`:

1. **Retry warm-up race.** After Cloud Run reports 100% traffic on the new revision, in-flight requests can still land on the previous revision's process during the brief warm-up window. The current one-shot check after a fixed `sleep 30` flakes when it races that window. Now retries up to 6 times with 10s backoff.

2. **Accept `*` for headless login.** `/api/headless/login` uses `public_cors` (`allow_origin = Any`) so the server responds with `Access-Control-Allow-Origin: *` rather than echoing the request origin. The previous test compared the response to `PREVIEW_ORIGIN` exactly and would fail forever after that route's CORS layer was switched (which happened in #34 / `655fe9a`). The `/api/auth/register` test still requires the exact origin echo since that route uses the predicate-based `auth_cors`.

## Why this matters

The most recent production deploy (build `110b5ca2`) hit exactly this — image was built and Cloud Run revision 234 deployed and serving traffic, but the smoke test failed and marked the build FAILURE. False alarm masked a successful deploy. With this change the test catches real CORS regressions without false-failing on cutover races or design changes.

## Test plan

- [ ] Trigger Cloud Build on this branch (or merge then trigger on main)
- [ ] Confirm both smoke tests pass: `Register CORS preflight passed` and `Headless login CORS preflight passed (allow-origin='*')`
- [ ] Confirm a real regression is still caught (e.g. revert the public_cors swap on a throwaway branch and verify the headless test fails with a clear error)

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -122,57 +122,62 @@ steps:
         echo "✅ Readiness check passed"
 
         PREVIEW_ORIGIN="https://preview.openvine-app.pages.dev"
+        HEADLESS_PUBLIC_ORIGIN="https://keycast-cors-smoke.invalid"
 
-        # Helper: read the Access-Control-Allow-Origin header from a CORS preflight.
+        # Helper: verify the Access-Control-Allow-Origin header from a CORS preflight.
         # Retries to ride out the brief warm-up window where requests can still land
         # on the previous revision after Cloud Run reports 100% traffic on the new one.
-        # First arg = endpoint path, prints the observed allow-origin value (or empty).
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz:
@dcadenas Can you please review? TYSM!

dcadenas (approved):
✅ smoke-test retry + headless origin handling look solid.

PR #156, opened 2026-04-27
Documenting no-name response shape change

### LINKED ISSUE #79: Follow-up: document nostr.json no-name response shape change

sergey3bv:
Should close #79. Reopen of #101


dcadenas (approved):
📝 docs now match the actual response shape, ship it.

PR #102, opened 2026-04-21
Making e2e test domain independent

sergey3bv:
Should close #77

PR #157, opened 2026-04-27
test(e2e): make the test domain independent

### LINKED ISSUE #77: test(nip05): make the badge e2e test domain-independent
Follow-up from PR #72.

`e2e/tests/nip05-badge.spec.ts` hardcodes `alice.name_123@divine.video`. That will become fragile if CI sets `NIP05_DOMAIN` or `DOMAIN`, or if the test tenant domain changes.

Please derive the expected identifier from the profile response or test configuration instead of hardcoding the domain.

sergey3bv:
Should close #77. Reopen of #102

### DIFF
--- a/e2e/tests/nip05-badge.spec.ts
+++ b/e2e/tests/nip05-badge.spec.ts
@@ -25,7 +25,9 @@ test.describe("NIP-05 profile badge", () => {
     expect(profileRes.status()).toBe(200);
     const profile = await profileRes.json();
     expect(profile.username).toBe("alice.name_123");
-    expect(profile.nip05).toBe("alice.name_123@divine.video");
+    const expectedDomain = profile.nip05.split("@").slice(1).join("@");
+    const expectedNip05 = `${profile.username}@${expectedDomain}`;
+    expect(profile.nip05).toBe(expectedNip05);
 
     const baseURL = process.env.API_URL || "http://localhost:3000";
     const url = new URL(baseURL);
@@ -42,6 +44,6 @@ test.describe("NIP-05 profile badge", () => {
 
     await page.goto("/");
     await expect(page.locator("text=NIP-05 Verified")).toBeVisible();
-    await expect(page.locator("text=alice.name_123@divine.video")).toBeVisible();
+    await expect(page.getByText(expectedNip05, { exact: true })).toBeVisible();
   });
 });


e2e/tests/nip05-badge.spec.ts
```diff
@@ -25,7 +25,11 @@ test.describe("NIP-05 profile badge", () => {
     expect(profileRes.status()).toBe(200);
     const profile = await profileRes.json();
     expect(profile.username).toBe("alice.name_123");
-    expect(profile.nip05).toBe("alice.name_123@divine.video");
+    expect(profile.nip05).toContain("@");
+    const expectedNip05 = profile.nip05;
+    expect(expectedNip05).toBe(
```
dcadenas: This assertion does not validate the domain.
`expectedNip05` is set to `profile.nip05` on line 29, then the right-hand side rebuilds it by splitting that same value on `@` and re-joining with `profile.username`.
For any well-formed `username@domain` value the server returns, the assertion holds regardless of which domain it is.
A server-side regression where `NIP05_DOMAIN` is configured but the response falls back to `divine.video` (or the other way around) would pass silently.
Issue #77 asked for the expected identifier to come from test configuration, not from the response.

Derive the expected domain from an independent source. The simplest version mirrors the server's resolution order in `resolve_nip05_domain` at `api/src/api/http/auth.rs:62`:

```ts
const expectedDomain = process.env.NIP05_DOMAIN ?? process.env.DOMAIN ?? "divine.video";
expect(profile.nip05).toBe(`${profile.username}@${expectedDomain}`);
```

At minimum, capture the expected domain in a test constant rather than reading it back from the response.

dcadenas (changes requested):
The new test assertion does not validate the NIP-05 domain. It reconstructs the expected value from the response itself, so it cannot detect a domain regression. The expected domain needs to come from test configuration before this can merge.

sergey3bv:

e2e/tests/nip05-badge.spec.ts
```diff
@@ -25,7 +25,11 @@ test.describe("NIP-05 profile badge", () => {
     expect(profileRes.status()).toBe(200);
     const profile = await profileRes.json();
     expect(profile.username).toBe("alice.name_123");
-    expect(profile.nip05).toBe("alice.name_123@divine.video");
+    expect(profile.nip05).toContain("@");
+    const expectedDomain =
+      process.env.NIP05_DOMAIN ?? process.env.DOMAIN ?? "divine.video";
```
dcadenas: This only works when the tenant host is `localhost` or `127.0.0.1`. For any other tenant host, the server uses the tenant host itself and ignores `NIP05_DOMAIN`/`DOMAIN`:

https://github.com/divinevideo/keycast/blob/799aebf893207cbffc493187116a10923f900f32/api/src/api/http/auth.rs#L62-L71

`tenant-a.localhost` is already in the default e2e `ALLOWED_TENANT_DOMAINS`, so as soon as CI points `API_URL` at `http://tenant-a.localhost:3000` the server returns `alice.name_123@tenant-a.localhost` and this assertion fails.

Derive the domain from the server response instead of trying to rebuild the server's resolution rules in the test:

```suggestion
    const expectedDomain = profile.nip05.split("@").slice(1).join("@");
    const expectedNip05 = `${profile.username}@${expectedDomain}`;
    expect(profile.nip05).toBe(expectedNip05);
```

dcadenas: `??` only short-circuits on `null`/`undefined`, but the server treats empty and whitespace values as unset via `.filter(|value| !value.trim().is_empty())`:

https://github.com/divinevideo/keycast/blob/799aebf893207cbffc493187116a10923f900f32/api/src/api/http/auth.rs#L62-L71

So with `NIP05_DOMAIN=""` (or `"   "`) the server still returns `alice.name_123@divine.video`, but this line gives `expectedNip05 = "alice.name_123@"` and the assertion fails. Deriving `expectedDomain` from `profile.nip05` (as suggested on the other thread) sidesteps this too.

dcadenas (changes requested):
The change swaps the hardcoded `divine.video` for an env-var-reconstructed expected domain, but the reconstruction does not match what the server actually returns under the tenant configurations this PR is meant to cover. Derive `expectedDomain` from the server's `profile.nip05` response instead of trying to mirror the resolution logic in the test. Details inline.

sergey3bv:

dcadenas:
Will implement later

PR #103, opened 2026-04-21
Fixing default domain for NIP-05

sergey3bv:
Should close #76

Issue #115, opened 2026-04-22
docs(dashboard): design identity states for username, claimed, and verified NIP-05


NotThatKindOfDrLiz:
The dashboard should not collapse local username, claimed handle, and verified NIP-05 into one state.\n\nScope\n- Define the user-visible identity states for dashboard/profile UI\n- Specify exact copy for username claimed, verification pending, verified, failed sync, and retry paths\n- Decide when to show badges, warnings, or no status at all\n- Align UI behavior with the backend source of truth for NIP-05 state\n\nAcceptance criteria\n- UI states and copy are documented\n- Each state maps to explicit backend/API data\n- The dashboard does not show a verified badge based only on presence of username

NotThatKindOfDrLiz:
@sergey3bv this issue tracks the follow-up UI work that was split out of PR #101. Please do not fold it into the docs-only replacement for #79. The next step here is to define the dashboard states and copy for local username vs claimed handle vs verified NIP-05, based on explicit backend state rather than presence of a username alone.

PR #158, opened 2026-04-27
fix(nip05): set the default domain correctly

### LINKED ISSUE #76: fix(nip05): remove the hardcoded default domain
Follow-up from PR #72.

`DEFAULT_NIP05_DOMAIN` is currently hardcoded to `divine.video`, which couples the binary to one deployment and is surprising for self-hosted or OSS use.

Please either make the fallback domain explicit configuration with fail-fast behavior when needed, or drop the fallback and use the tenant domain consistently.

sergey3bv:
Should close #76. Reopen of #103

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -32,7 +32,7 @@ use sqlx::PgPool;
 const DEFAULT_TOKEN_EXPIRY_HOURS: i64 = 24;
 pub const EMAIL_VERIFICATION_EXPIRY_HOURS: i64 = 24;
 const PASSWORD_RESET_EXPIRY_HOURS: i64 = 1;
-const DEFAULT_NIP05_DOMAIN: &str = "divine.video";
+const MAX_NIP05_USERNAME_LENGTH: usize = 64;
 
 /// Get token expiry in seconds. Uses `TOKEN_EXPIRY_SECONDS` env var if set,
 /// otherwise defaults to 24 hours (86400 seconds).
@@ -54,20 +54,63 @@ pub fn generate_secure_token() -> String {
 
 /// Returns the domain segment of NIP-05 identifiers (`user@domain`).
 ///
-/// `/.well-known/nostr.json` is served for the request's tenant, and the tenant is derived from
-/// the Host header—so `nip05` in `/user/profile` must use that same domain. Letting `NIP05_DOMAIN`
-/// or `DOMAIN` override a real tenant would make verification fail (profile claims one host while
-/// discovery lives on another). Those env vars apply only when the tenant host is `localhost` or
-/// `127.0.0.1` (local dev), with `DEFAULT_NIP05_DOMAIN` as the final fallback.
-fn resolve_nip05_domain(tenant_domain: &str) -> String {
-    if tenant_domain == "localhost" || tenant_domain == "127.0.0.1" {
-        return std::env::var("NIP05_DOMAIN")
-            .ok()
-            .or_else(|| std::env::var("DOMAIN").ok())
-            .filter(|value| !value.trim().is_empty())
-            .unwrap_or_else(|| DEFAULT_NIP05_DOMAIN.to_string());
+/// For tenants that need a different NIP-05 namespace than their routing host, settings can
+/// specify `nip05_domain`. Falls back to `tenant.domain` when unset, blank, or malformed.
+fn is_valid_nip05_domain(domain: &str) -> bool {
+    !domain.is_empty()
+        && !domain.contains('@')
+        && !domain.contains('/')
+        && !domain.chars().any(char::is_whitespace)
+}
+
+fn resolve_nip05_domain(tenant: &crate::api::tenant::Tenant) -> String {
+    let settings = match tenant.get_settings() {
+        Ok(settings) => settings,
+        Err(error) => {
+            tracing::warn!(
+                tenant_id = tenant.id,
+                tenant_domain = %tenant.domain,
+                error = %error,
+                "Failed to parse tenant settings; falling back to tenant domain for NIP-05"
+            );
+            return tenant.domain.clone();
+        }
+    };
+
+    let Some(raw_override) = settings.nip05_domain else {
+        return tenant.domain.clone();
+    };
+
+    let override_domain = raw_override.trim();
+    if override_domain.is_empty() {
+        tracing::warn!(
+            tenant_id = tenant.id,
+            tenant_domain = %tenant.domain,
+            "Ignoring blank nip05_domain override; falling back to tenant domain"
+        );
+        return tenant.domain.clone();
     }
-    tenant_domain.to_string()
+
+    if !is_valid_nip05_domain(override_domain) {
+        tracing::warn!(
+            tenant_id = tenant.id,
+            tenant_domain = %tenant.domain,
+            nip05_domain = %override_domain,
+            "Ignoring invalid nip05_domain override; falling back to tenant domain"
+        );
+        return tenant.domain.clone();
+    }
+
+    if override_domain != tenant.domain {
+        tracing::warn!(
+            tenant_id = tenant.id,
+            tenant_domain = %tenant.domain,
+            nip05_domain = %override_domain,
+            "Using NIP-05 domain override different from tenant routing domain; ensure /.well-known/nostr.json is discoverable for this host"
+        );
+    }
+
+    override_domain.to_string()
 }
 
 fn normalize_nip05_username(raw_username: &str) -> Result<String, AuthError> {
@@ -90,6 +133,13 @@ fn normalize_nip05_username(raw_username: &str) -> Result<String, AuthError> {
         return Err(AuthError::Internal(
             "Username cannot start or end with a hyphen".to_string(),
         ));
+    }
+
+    if username.len() > MAX_NIP05_USERNAME_LENGTH {
+        return Err(AuthError::Internal(format!(
+            "Username must be at most {} characters",
+            MAX_NIP05_USERNAME_LENGTH
+        )));
     }
 
     Ok(username)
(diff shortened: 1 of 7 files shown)


dcadenas (approved):
🌐 NIP-05 default domain lands on the right host

dcadenas:
I moved this to draft temporarily to block it's merged until the corresponding env vars are set properly in all clusters as a prerequisite for this

dcadenas:
@rabble — pinging you up front since the local NIP-05 endpoint was originally your design call (issue #43, implemented in PR #72).

Quick context on why we have two NIP-05 servers in this stack: keycast serves its own `/.well-known/nostr.json` so any self-hosted keycast deployment can act as a one-stop OAuth + NIP-05 provider. That makes sense when one host does both jobs.

For Divine specifically we run divine-name-server separately at the apex (`divine.video`) and per-user subdomains, so the keycast-local copy is the secondary one. The catch: `tenant.domain` works fine as the NIP-05 namespace for self-hosters but not for us — for us it's just the routing host (`login.divine.video`, where the pods actually receive traffic) and can't change. So we need an override so the NIP-05 namespace can differ from the routing host.

After tracing through this PR, I think the env-var approach is solving the wrong problem and there's a smaller fix that closes more issues.

`NIP05_DOMAIN` is only consulted when the tenant host is localhost. On any real host the resolver returns `tenant.domain` directly, so setting the env across our deploy configs has no effect on prod — it just satisfies `validate_environment`.

And what the resolver returns on Divine prod is `username@login.divine.video`, but issue #43 originally asked for `username@divine.video`. That's the drift the override needs to fix.

I'd suggest moving the override onto the tenant row instead of an env var. `TenantSettings` already carries `relay` and `email_from`; add an optional `nip05_domain`:

\`\`\`rust
pub struct TenantSettings {
    pub relay: Option<String>,
    pub email_from: Option<String>,
    pub nip05_domain: Option<String>,
}

fn resolve_nip05_domain(tenant: &Tenant) -> String {
    tenant.get_settings().ok()
        .and_then(|s| s.nip05_domain)
        .unwrap_or_else(|| tenant.domain.clone())
}
\`\`\`

Drop the env-var path and `validate_environment` check. Localhost falls through to `tenant.domain = "localhost"`, which is fine for dev — no hardcoded `divine.video` constant anywhere.

For Divine, one `UPDATE tenants SET settings = ...` per environment sets tenant 1's `nip05_domain = "divine.video"`. That's Cloud Run prod plus four GKE overlays. After that, `/user/profile.nip05` finally returns `username@divine.video` as #43 originally specced.

This still fits Liz's scoping for #76 — it's the "explicit configuration" arm she described, just per-tenant in the DB instead of per-binary in env. Same surface, no UI or profile-shape changes. The per-user-subdomain form (`_@<u>.divine.video`) stays out of scope; that belongs with the broader work in #115/#116/#117.

Happy to revise this PR or open a fresh one and close this as superseded — whichever you prefer, @sergey3bv.

Closes #76 and #43.

dcadenas (changes requested):
Sorry for the back-and-forth, @sergey3bv. My understanding shifted while planning the deploy-config rollout for this — pulling on the thread (multiple NIP-05 servers, `tenant.domain` semantics, what #43 originally asked for) showed a cleaner per-tenant fix. Switching my review to request changes so the state reflects that. Details in the comment above — happy to take the impl on a fresh PR if that's easier.

sergey3bv:
Hey, @dcadenas, I tried to update the PR to make it work the way you describe. Could you please take a look?

api/src/api/http/auth.rs
```diff
@@ -2137,7 +2132,7 @@ pub async fn get_profile(
         .get_username(&user_pubkey, tenant_id)
         .await?
         .flatten();
-    let nip05_domain = resolve_nip05_domain(&tenant.0.domain);
+    let nip05_domain = resolve_nip05_domain(&tenant.0);
```
dcadenas, line 2190: This is a silent behaviour change for every existing tenant. Before this PR, tenants whose routing host differed from their public NIP-05 domain (`login.divine.video` routing `divine.video` users) got `alice@divine.video` from the `NIP05_DOMAIN` / `DOMAIN` env fallback. After this PR, `resolve_nip05_domain` only reads `tenants.settings.nip05_domain` and otherwise falls back to `tenant.domain`.

No existing row has `settings.nip05_domain` set. `get_default_settings` in `tenant.rs:326` stores `nip05_domain: None` for new tenants, and this PR ships no migration that backfills existing rows. So on deploy, every active user starts advertising `alice@<routing-host>` (e.g. `alice@login.divine.video`) instead of their current identifier, and NIP-05 verification breaks anywhere the old identifier is cached.

Before this can merge, ship one of: a SQL migration that backfills `settings.nip05_domain` from the current env-derived value for every existing tenant, or an operator runbook plus a startup warning when a tenant with `settings.nip05_domain = None` is loaded. Silent cutover is not safe here.

e2e/tests/nip05-badge.spec.ts
```diff
@@ -25,7 +28,7 @@ test.describe("NIP-05 profile badge", () => {
     expect(profileRes.status()).toBe(200);
     const profile = await profileRes.json();
     expect(profile.username).toBe("alice.name_123");
-    expect(profile.nip05).toBe("alice.name_123@divine.video");
+    expect(profile.nip05).toBe(`alice.name_123@${nip05Domain}`);
```
dcadenas, line 31: This assertion no longer matches what the server does. It expects `profile.nip05 === `alice.name_123@${process.env.NIP05_DOMAIN}``, but after this PR `resolve_nip05_domain` reads no env var. The domain comes only from `tenants.settings.nip05_domain` with a fallback to `tenants.domain`. Nothing in this PR, `registerAndVerify`, or `e2e/helpers/db.ts` seeds `settings.nip05_domain` on the test tenant, and the `expect(nip05Domain).toBeTruthy()` guard only checks the env var is set. It does not bridge `NIP05_DOMAIN` to `tenants.settings`.

The test only passes when `NIP05_DOMAIN` happens to equal the test tenant's routing host, which CI does not guarantee. The badge locator at line 48 has the same problem.

Either seed `tenants.settings.nip05_domain` on the test tenant before the assertion runs, or drop the env coupling entirely and assert against `tenant.domain`.

api/src/api/http/auth.rs
```diff
@@ -54,20 +53,16 @@ pub fn generate_secure_token() -> String {
...
-        return std::env::var("NIP05_DOMAIN")
-            .ok()
-            .or_else(|| std::env::var("DOMAIN").ok())
-            .filter(|value| !value.trim().is_empty())
-            .unwrap_or_else(|| DEFAULT_NIP05_DOMAIN.to_string());
-    }
-    tenant_domain.to_string()
+/// For tenants that need a different NIP-05 namespace than their routing host, settings can
+/// specify `nip05_domain`. Falls back to `tenant.domain` when unset, blank, or malformed.
+fn resolve_nip05_domain(tenant: &crate::api::tenant::Tenant) -> String {
```
dcadenas, line 66: `resolve_nip05_domain` now returns `tenant.settings.nip05_domain` verbatim with no validation. The previous version enforced a structural rule (env fallback only for `localhost` / `127.0.0.1`, real tenants always returned `tenant_domain`) and had a comment block explaining why the NIP-05 domain and the host serving `/.well-known/nostr.json` must be the same. That rule and that doc are both gone.

The new function accepts any non-empty trimmed string, which means:

- A typo like `nip05_domain = "@divine.video"` produces `user@@divine.video`, and `"divine.video/evil"` produces `user@divine.video/evil`. Both are invalid per NIP-05.
- An operator who sets `nip05_domain = "example.com"` on a tenant whose routing host is `login.example.com` silently emits `user@example.com` profiles that no verifier can resolve. A NIP-05 client will GET `example.com/.well-known/nostr.json`, which keycast only answers if `example.com` is itself a known tenant routed to the same user table (see `TenantExtractor` at `tenant.rs:88-100`).

At minimum, validate the host shape (no `@`, no `/`, no whitespace, non-empty after trim) and reject or warn on mismatch. Better: cross-check at settings-write time that the override has a discoverable well-known path on this instance, or keep the original invariant comment so a future reader knows a bogus override silently breaks verification.

api/src/api/http/auth.rs
```diff
@@ -54,20 +53,16 @@ pub fn generate_secure_token() -> String {
...
+    tenant
+        .get_settings()
```
dcadenas: `tenant.get_settings().ok()` silently swallows serde_json errors. If an operator writes malformed JSON into `tenants.settings` (unquoted key, trailing comma, whatever), `get_settings()` returns `Err`, `.ok()` maps to `None`, and the resolver falls back to `tenant.domain` with no log line. The operator sees profiles using the routing host and has no signal that their `nip05_domain` override was rejected. The new test `resolve_nip05_domain_falls_back_to_tenant_domain_on_invalid_settings` at L4168 codifies this as intended behaviour.

Emit a `tracing::warn!` on the error-to-fallback path so misconfigured settings show up in operator logs.

dcadenas (changes requested):
This changes the NIP-05 domain for every existing tenant the moment it deploys, with no migration to backfill `settings.nip05_domain` and no validation on the new override. Every user whose tenant has a routing host different from their public NIP-05 domain will silently start advertising a new identifier, and any typo in the override goes straight into profiles. An unrelated username length cap and its regression tests also got removed in this PR, which should either come back or move to its own change.

---
**Inline comments (outside PR diff):**

- **api/src/api/http/auth.rs:89** — This PR silently removes `MAX_NIP05_USERNAME_LENGTH = 64`, the length-cap check in `normalize_nip05_username`, and the two regression tests `test_normalize_nip05_username_accepts_max_length_boundary` and `test_normalize_nip05_username_rejects_length_above_boundary`. The stated goal of this PR is to fix the NIP-05 domain fallback. The username length cap is unrelated scope. After this lands, a caller can push an arbitrary-length username through `normalize_nip05_username` and the boundary tests are no longer there to catch a regression.

Restore the cap and its tests, or split the removal into its own PR with a reason for dropping the bound.

sergey3bv:

dcadenas:
We will retriage and pick this up later

PR #105, opened 2026-04-21
Adding validation error for NIP-05 username

sergey3bv:
Should close #74

PR #160, opened 2026-04-27
fix(nip05): return validation errors for username updates

### LINKED ISSUE #74: fix(profile): return 4xx validation errors from username updates
Follow-up from PR #72.

`normalize_nip05_username` and related profile username validation currently return `AuthError::Internal` for user-input validation failures such as empty usernames or invalid characters. These should be 4xx-category errors, not internal server errors.

This appears to preserve an existing pattern rather than introduce a new regression, so it is non-blocking for PR #72. Please introduce an appropriate validation/bad-request error path and update `update_profile` callers/status codes consistently.

sergey3bv:
Should close #74. Reopen of #105

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -34,6 +34,8 @@ pub const EMAIL_VERIFICATION_EXPIRY_HOURS: i64 = 24;
 const PASSWORD_RESET_EXPIRY_HOURS: i64 = 1;
 const DEFAULT_NIP05_DOMAIN: &str = "divine.video";
 const MAX_NIP05_USERNAME_LENGTH: usize = 64;
+const USERNAME_CONFLICT_MESSAGE: &str =
+    "Username is not available. Please choose another username.";
 const USERS_EMAIL_TENANT_CONSTRAINT: &str = "idx_users_email_tenant";
 pub(crate) const INVALID_EMAIL_CODE: &str = "INVALID_EMAIL";
 pub(crate) const INVALID_EMAIL_MESSAGE: &str = "Please enter a valid email address.";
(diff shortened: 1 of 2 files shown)


api/src/api/http/auth.rs
```diff
@@ -2224,7 +2226,7 @@ pub async fn update_profile(
                             username,
                             error_msg
                         );
-                        return Err(AuthError::Internal(error_msg));
+                        return Err(AuthError::BadRequest(error_msg));
```
dcadenas: The 400 body now echoes the upstream `reason` string from `divine_names::check_availability` straight to the client as `{"error": <reason>}`.

Before this PR that branch returned `AuthError::Internal`, which the `IntoResponse` impl at auth.rs:370-377 deliberately swallowed and replaced with a generic 503.

Every other `BadRequest(String)` site in this file uses a compile-time literal, so this is the only call passing an external-origin message to clients.

That gives the upstream control of the 400 body and risks leaking debug strings, internal hostnames, or unescaped HTML.

Map the upstream result to a fixed allow-list of reason codes before wrapping in `BadRequest`, or at minimum bound length and character set. The `None` branch two lines below is fine because it uses a literal.

dcadenas (changes requested):
The new 400 path forwards the upstream `reason` string from divine-name-server straight to the client.

That string was previously hidden behind a generic 503, and echoing it verbatim now lets external debug strings, internal hostnames, or unescaped HTML reach end users.

Map the upstream result to a fixed allow-list of reason codes, or document that those strings are end-user-safe, before this can merge.

sergey3bv:

dcadenas (approved):
✅ clearer error when the NIP-05 username is invalid.

sergey3bv:
Hey, @dcadenas, I rebased the RP, could you please take a look?

cc @NotThatKindOfDrLiz

dcadenas:

PR #161, opened 2026-04-27
test(nip05): add tests for domain resolution

### LINKED ISSUE #75: test(nip05): add tests for domain resolution

sergey3bv:
Should close #75. Reopen of #109

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -65,10 +65,13 @@ pub fn generate_secure_token() -> String {
 /// `127.0.0.1` (local dev), with `DEFAULT_NIP05_DOMAIN` as the final fallback.
 fn resolve_nip05_domain(tenant_domain: &str) -> String {
     if tenant_domain == "localhost" || tenant_domain == "127.0.0.1" {
-        return std::env::var("NIP05_DOMAIN")
-            .ok()
-            .or_else(|| std::env::var("DOMAIN").ok())
-            .filter(|value| !value.trim().is_empty())
+        let non_empty_env = |name: &str| {
+            std::env::var(name)
+                .ok()
+                .filter(|value| !value.trim().is_empty())
+        };
+        return non_empty_env("NIP05_DOMAIN")
+            .or_else(|| non_empty_env("DOMAIN"))
             .unwrap_or_else(|| DEFAULT_NIP05_DOMAIN.to_string());
     }
     tenant_domain.to_string()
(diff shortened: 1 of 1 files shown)


api/src/api/http/auth.rs
```diff
@@ -4106,6 +4136,58 @@ mod tests {
...
+    #[serial]
+    fn test_resolve_nip05_domain_passthroughs_tenant_domain() {
+        let _nip05_domain_guard = EnvVarGuard::remove("NIP05_DOMAIN");
+        let _domain_guard = EnvVarGuard::remove("DOMAIN");
+
+        let resolved = super::resolve_nip05_domain("tenant.example");
+        assert_eq!(resolved, "tenant.example");
+    }
+
+    #[test]
```
dcadenas, line 4721: This test does not exercise the empty-env filter.

`"tenant.example"` is a non-loopback host, so `resolve_nip05_domain` takes the passthrough branch at auth.rs:70 and returns `"tenant.example"` without ever reading `NIP05_DOMAIN` or `DOMAIN`. The filter at auth.rs:67 is never evaluated, so the assertion would still hold if that line were deleted or inverted.

Use `"localhost"` (or `"127.0.0.1"`) as the tenant and assert the result equals `super::DEFAULT_NIP05_DOMAIN`.

A paired case with `NIP05_DOMAIN="  "` and `DOMAIN="domain.example"` asserting `"domain.example"` would pin precedence with the filter in play.

api/src/api/http/auth.rs
```diff
@@ -4106,6 +4136,58 @@ mod tests {
...
+
+        let resolved = super::resolve_nip05_domain("localhost");
+        assert_eq!(resolved, "domain.example");
+    }
+
+    #[test]
+    #[serial]
+    fn test_resolve_nip05_domain_uses_default_for_localhost() {
+        let _nip05_domain_guard = EnvVarGuard::remove("NIP05_DOMAIN");
+        let _domain_guard = EnvVarGuard::remove("DOMAIN");
```
dcadenas, line 4703: `127.0.0.1` is only exercised in this single default-fallback case.

There is no assertion that env-var precedence applies when the tenant is `127.0.0.1` rather than `localhost`. If a future refactor narrows the guard at auth.rs:63 to only match `localhost`, this suite would still pass the 127.0.0.1 default case while silently losing env-var resolution for that input.

Add a one-line `127.0.0.1` assertion to the existing prefers/falls_back tests to close the gap.

dcadenas (changes requested):
Both new tests pass without exercising the code they claim to cover.

The empty-env test takes a passthrough branch and never reads the env vars.

The localhost-default test never checks env-var precedence on `127.0.0.1`.

Concrete fixes are in the inline comments.

sergey3bv:

dcadenas (approved):
🧪 nice coverage on the NIP-05 resolver paths.

sergey3bv:
Hey, @dcadenas, I rebased the PR to the latest changes, could you please take a look?

cc @NotThatKindOfDrLiz

dcadenas:

PR #165, opened 2026-04-28
Adding username boundaries check

### LINKED ISSUE #80: Follow-up: add NIP-05 username length limit

sergey3bv:
Should close #80. Reopen of #99


sergey3bv:
Duplicate of #146

PR #166, opened 2026-04-28
feat(api): add /.well-known/openid-configuration discovery endpoint

### LINKED ISSUE #274: Serve OIDC discovery metadata instead of SPA fallback
## Summary

Add the `/.well-known/openid-configuration` discovery endpoint so OAuth/OIDC clients receive JSON instead of the SPA HTML fallback.

## Motivation

Requests to `/.well-known/openid-configuration` currently fall through to SvelteKit's SPA index HTML. OIDC clients doing autodiscovery fail to parse the response and treat Keycast as unreachable.

## Acceptance Criteria

- `GET /.well-known/openid-configuration` returns JSON with the correct issuer and supported OAuth capability metadata.
- The endpoint advertises only capabilities Keycast actually supports.
- The SPA fallback no longer handles this well-known path.
- Regression tests cover JSON content type, issuer behavior, and trailing-slash normalization.

rabble:
## Summary

- Adds `GET /.well-known/openid-configuration` so autodiscovery callers receive JSON instead of the SvelteKit SPA fallback HTML.
- Keeps the response honest about current Keycast behavior: OAuth authorization-code metadata only, with no `openid` scope, ID Token signing, `jwks_uri`, or UserInfo claims.
- Splits the existing origin helper so the ATProto metadata endpoint can still prefer `ATPROTO_ENTRYWAY_ORIGIN`, while the OpenID compatibility endpoint uses the normal app origin from `APP_URL`/`VITE_DOMAIN`.
- Allows `/api/oauth/token` to accept both standards-compliant `application/x-www-form-urlencoded` requests and the existing JSON request format.
- Hardens the new tests to match the sibling metadata tests: serialized process-env mutation, RAII env restoration, and explicit coverage that `ATPROTO_ENTRYWAY_ORIGIN` does not leak into this endpoint.

## Motivation

- Without a route for `/.well-known/openid-configuration`, the static SPA fallback served `index.html` with a 200 response.
- Some clients and probes check this well-known path before trying OAuth endpoints, so returning JSON is a useful compatibility improvement.
- Keycast does not currently issue OpenID Connect ID Tokens, so advertising OIDC provider capabilities would make clients fail later in the flow after trusting unsupported metadata.
- The discovered OAuth token endpoint now accepts the form-encoded token requests that standards-following OAuth clients send.

## Related Issue

- Closes #274

## Testing

- [x] `cargo test --package keycast_api --test oauth_token_request_body_test`
- [x] `cargo test --package keycast_api --test openid_configuration_test`
- [x] `cargo test --package keycast_api --test atproto_oauth_metadata_test`
- [x] `DATABASE_URL=postgres://postgres:password@localhost:5432/keycast_test REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test --workspace --verbose`
- [x] `DATABASE_URL=postgres://postgres:password@localhost:5432/keycast_test REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test -p keycast_api --features integration-tests --lib --tests --verbose`
- [x] `DATABASE_URL=postgres://postgres:password@localhost:5432/keycast_test REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test -p keycast_core --features integration-tests --lib --tests --verbose`
- [x] `cargo test -p keycast_signer --features integration-tests --lib --tests --verbose`
- [ ] Local rerun of `cargo test --workspace --verbose` after `0fa43b3` could not complete because port 5432 is occupied by non-Keycast `config-postgres-1`, whose credentials reject the repo default `postgres:password` test URL.
- [ ] `bun run test` was attempted locally, but this machine is missing `cargo-nextest`.

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/openid_configuration.rs
+++ b/api/src/api/http/openid_configuration.rs
@@ -1,0 +1,73 @@
+// ABOUTME: OpenID Connect Discovery 1.0 metadata endpoint.
+// ABOUTME: Reuses the OAuth 2.0 origin helper so the discovery document is
+// rooted at the same APP_URL as the rest of the service.
+//
+// Keycast is primarily an OAuth 2.0 + Nostr Bunker (NIP-46) provider rather
+// than a fully-fledged OpenID Connect provider — we don't issue spec-shaped
+// id_tokens or expose a userinfo endpoint at this time. Even so, OIDC
+// autodiscovery clients (and the reliability assurance suite's
+// `oidc_discovery` probe) hit `/.well-known/openid-configuration` to find
+// the OAuth endpoints. Without this handler the SvelteKit SPA fallback
+// served the index HTML instead of JSON; clients then got a JSON parse
+// error and treated keycast as unreachable.
+//
+// The document below is deliberately conservative: it advertises only the
+// capabilities keycast actually supports today. A client that tries to use
+// strict OIDC features (id_token signing, userinfo) will fail at runtime
+// rather than fail silently — that's the correct degradation.
+//
+// Sibling: `atproto_oauth_metadata.rs` serves
+// `/.well-known/oauth-authorization-server` for ATProto-flavoured clients.
+
+use axum::Json;
+use serde::Serialize;
+
+use super::atproto_oauth_metadata::authorization_server_origin;
+
+#[derive(Debug, Serialize)]
+pub struct OpenIdConfiguration {
+    issuer: String,
+    authorization_endpoint: String,
+    token_endpoint: String,
+    response_types_supported: Vec<String>,
+    grant_types_supported: Vec<String>,
+    scopes_supported: Vec<String>,
+    token_endpoint_auth_methods_supported: Vec<String>,
+    code_challenge_methods_supported: Vec<String>,
+    // OIDC-required capability lists. Keycast does not currently issue
+    // OIDC-shaped id_tokens, but every value here describes what the
+    // service WOULD support if/when it does — keeping the list non-empty
+    // satisfies strict autodiscovery clients without misrepresenting
+    // current behavior. ``subject_types: ["public"]`` is the only sensible
+    // value for a Nostr-pubkey-rooted identity model.
+    subject_types_supported: Vec<String>,
+    id_token_signing_alg_values_supported: Vec<String>,
+}
+
+fn endpoint(origin: &str, path: &str) -> String {
+    format!(
+        "{}/{}",
+        origin.trim_end_matches('/'),
+        path.trim_start_matches('/')
+    )
+}
+
+pub async fn openid_configuration_metadata() -> Json<OpenIdConfiguration> {
+    let origin = authorization_server_origin();
+
+    Json(OpenIdConfiguration {
+        issuer: origin.clone(),
+        authorization_endpoint: endpoint(&origin, "/api/oauth/authorize"),
+        token_endpoint: endpoint(&origin, "/api/oauth/token"),
+        response_types_supported: vec!["code".to_string()],
+        grant_types_supported: vec![
+            "authorization_code".to_string(),
+            "refresh_token".to_string(),
+        ],
+        scopes_supported: vec!["openid".to_string(), "profile".to_string()],
+        token_endpoint_auth_methods_supported: vec!["none".to_string()],
+        code_challenge_methods_supported: vec!["S256".to_string()],
+        subject_types_supported: vec!["public".to_string()],
+        id_token_signing_alg_values_supported: vec!["ES256".to_string()],
+    })
+}
--- a/api/src/api/http/openid_configuration.rs
+++ b/api/src/api/http/openid_configuration.rs
@@ -1,28 +1,18 @@
-// ABOUTME: OpenID Connect Discovery 1.0 metadata endpoint.
-// ABOUTME: Reuses the OAuth 2.0 origin helper so the discovery document is
-// rooted at the same APP_URL as the rest of the service.
+// ABOUTME: Compatibility metadata for OpenID autodiscovery clients.
+// ABOUTME: Keeps the OpenID well-known path from falling through to the SPA
+// fallback while advertising only Keycast's current OAuth behavior.
 //
 // Keycast is primarily an OAuth 2.0 + Nostr Bunker (NIP-46) provider rather
-// than a fully-fledged OpenID Connect provider — we don't issue spec-shaped
-// id_tokens or expose a userinfo endpoint at this time. Even so, OIDC
-// autodiscovery clients (and the reliability assurance suite's
-// `oidc_discovery` probe) hit `/.well-known/openid-configuration` to find
-// the OAuth endpoints. Without this handler the SvelteKit SPA fallback
-// served the index HTML instead of JSON; clients then got a JSON parse
-// error and treated keycast as unreachable.
-//
-// The document below is deliberately conservative: it advertises only the
-// capabilities keycast actually supports today. A client that tries to use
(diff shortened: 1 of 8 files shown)


rabble:
this is a draft, we can work on atproto integration after launch

NotThatKindOfDrLiz:
I took this over and pushed `7a7871e`.

What changed:

- Fixed the env safety issue in `openid_configuration_test.rs` by serializing the tests and restoring `APP_URL` / `ATPROTO_ENTRYWAY_ORIGIN` with an RAII guard.
- Stopped advertising unsupported OIDC Provider capabilities. The endpoint now returns OAuth-shaped compatibility metadata only; it no longer emits `scopes_supported: ["openid"]`, ID Token signing metadata, or `jwks_uri`.
- Split the origin helper so ATProto authorization server metadata still prefers `ATPROTO_ENTRYWAY_ORIGIN`, while this compatibility endpoint uses the normal app origin.
- Shared the existing endpoint URL helper instead of copy-pasting it.

Verification:

- `cargo test --package keycast_api --test openid_configuration_test`
- `cargo test --package keycast_api --test atproto_oauth_metadata_test`
- `DATABASE_URL=postgres://postgres:password@localhost:5432/keycast_test REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test --workspace --verbose`
- `DATABASE_URL=postgres://postgres:password@localhost:5432/keycast_test REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test -p keycast_api --features integration-tests --lib --tests --verbose`
- `DATABASE_URL=postgres://postgres:password@localhost:5432/keycast_test REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test -p keycast_core --features integration-tests --lib --tests --verbose`
- `cargo test -p keycast_signer --features integration-tests --lib --tests --verbose`

I also tried `bun run test`; it did not complete locally because `cargo-nextest` is not installed on this machine.

NotThatKindOfDrLiz (approved):
Approved after takeover. The env-racy tests are fixed, the OpenID well-known response no longer advertises unsupported OIDC Provider capabilities, ATProto entryway origin handling stays scoped to the ATProto metadata endpoint, and CI checks are green.

api/src/api/http/openid_configuration.rs
```diff
@@ -0,0 +1,43 @@
...
+    code_challenge_methods_supported: Vec<String>,
+}
+
+pub async fn openid_configuration_metadata() -> Json<OpenIdConfiguration> {
+    let origin = app_origin();
+
+    Json(OpenIdConfiguration {
+        issuer: origin.clone(),
+        authorization_endpoint: endpoint(&origin, "/api/oauth/authorize"),
+        token_endpoint: endpoint(&origin, "/api/oauth/token"),
```
dcadenas, line 34: Reconcile this advertised endpoint with `/api/oauth/token` before merge. The handler uses `Json<TokenRequest>`, so a standards-compliant client that follows discovery and submits the normal `application/x-www-form-urlencoded` token request gets a 415. The adjacent `none` claim also promises a public-client flow while the handler leaves PKCE optional and does not bind the request `client_id` to the authorization code. Narrow this metadata now by omitting or explicitly qualifying `token_endpoint` and removing `token_endpoint_auth_methods_supported: ["none"]`; leave handler hardening for a separate change.

dcadenas (dismissed):
The new discovery document advertises OAuth behavior that `/api/oauth/token` does not implement. Reconcile the metadata with the endpoint's actual request format and security behavior before merge.

NotThatKindOfDrLiz:
@rabble I took over the remaining review blocker and pushed `0fa43b3`.

What changed:

- `/api/oauth/token` now accepts both `application/x-www-form-urlencoded` and the existing JSON request format.
- Added `oauth_token_request_body_test` coverage for form-encoded authorization-code requests, JSON refresh-token requests, and unsupported content types.
- Updated the existing direct token-handler test to use the new request wrapper.

Verification:

- `cargo test --package keycast_api --test oauth_token_request_body_test`
- `cargo test --package keycast_api --test openid_configuration_test`
- `cargo test --package keycast_api --test atproto_oauth_metadata_test`
- GitHub checks are green on `0fa43b3`.

Local note: `cargo test --workspace --verbose` could not complete on this machine because port 5432 is occupied by a non-Keycast Postgres container (`config-postgres-1`) whose credentials reject the repo default test URL. CI passed the PR test job.

Please review the takeover commit and merge when you are ready. @dcadenas, this should address the requested token endpoint request-format mismatch; PKCE remains accurately advertised as supported, not required.

NotThatKindOfDrLiz (approved):
Approved after takeover pass. The discovery metadata/token endpoint mismatch is resolved by accepting form-encoded token requests while preserving existing JSON callers, and CI is green.

api/src/api/http/openid_configuration.rs
dcadenas, line 34: Remove or narrow the public-client token claim at `api/src/api/http/openid_configuration.rs:34-40` before advertising this flow. It currently includes `token_endpoint_auth_methods_supported: ["none"]`, while the authorization-code exchange does not compare the request `client_id` with the stored code and only enforces PKCE when a challenge exists:

https://github.com/divinevideo/keycast/blob/0fa43b36801d849da57f3d64239caaec09ee0674/api/src/api/http/oauth.rs#L2603-L2631

A discovery-driven unauthenticated client can therefore follow the metadata into a token exchange with weaker binding than advertised. Update the discovery test to pin the narrowed metadata, keep deeper token-handler hardening outside this PR, and verify that form and JSON token decoding still work.

dcadenas (changes requested):
The discovery document still advertises an unauthenticated public-client token flow whose request binding and PKCE behavior do not match that contract. Narrow the metadata and pin the corrected shape in tests before merge, while keeping deeper token-handler hardening out of this PR.

dcadenas (approved):
The form-encoding gap I raised is fixed, and the discovery document is accurate. I pushed one commit for a smaller thing I found while re-reading.

I'm also retracting my objection to `token_endpoint_auth_methods_supported: ["none"]`. That field describes client authentication at the token endpoint and nothing else, and Keycast genuinely performs none, so it is correct as written. Removing it would have made the document worse, not safer: OpenID Connect Discovery 1.0 §3 and RFC 8414 §2 both default it to `client_secret_basic` when absent, so conforming clients would have started sending HTTP Basic credentials we do not accept. My mistake, and thanks for waiting on it.

The commit makes the new `TokenRequestBody` extractor compare `Content-Type` case-insensitively. RFC 9110 §8.3.1: "The type and subtype tokens are case-insensitive." The dispatch matched literally, so `Application/JSON` returned 400 where it returned 200 before this branch, since `axum::Json` parses through the `mime` crate and accepts either case. It normalizes the header on the request rather than only the comparison string, because `axum::Form` runs its own case-sensitive check and would otherwise still reject `Application/X-Www-Form-Urlencoded`. Two tests added, the three existing ones in that file still pass, and I ran the full workspace suite against Postgres and Redis plus `cargo fmt --check` and `clippy -D warnings` before pushing.

What I did not change: the authorization-code exchange binds the code by `redirect_uri` but not by `client_id`, and PKCE is verified only when a challenge was stored. Both are real, both predate this branch, and neither belongs in a metadata change.

- #337 — bind the authorization code to the requesting `client_id`
- #338 — require PKCE on the authorization-code flow

PR #168, opened 2026-04-29
feat: admin UI to manage registered_clients (OAuth client allowlist)

### LINKED ISSUE #164: feat: admin UI to manage registered_clients (OAuth client allowlist)
## Why

The `registered_clients` table is currently managed via ad-hoc SQL. We want a UI so additions and edits don't require DB access. #151 is a recent example of the workflow we want to eliminate.

## What

An admin page (gated by `VITE_ALLOWED_PUBKEYS`, tenant-scoped) for CRUD on `registered_clients`.

## Acceptance criteria

- **List** rows for the active tenant: `client_id`, `name`, `allowed_redirect_uris[]`, `created_at`, `updated_at`.
- **Add** new client: `client_id` (unique per tenant), `name`, one or more `allowed_redirect_uris` patterns.
- **Edit**: rename, add/remove individual redirect URI patterns.
- **Delete** with confirmation modal (deleting a live client breaks its OAuth flow immediately).
- Validation:
  - `client_id` non-empty, unique per tenant.
  - Each redirect URI pattern conforms to the wildcard matcher in `core/src/repositories/registered_client.rs` (`matches_redirect_pattern`) — the wildcard segment cannot contain `/`, so `https://*.example.com/callback` works but `https://example.com/*` only matches single-segment paths.
  - Inline pattern tester ("does this URI match?") so admins can verify before saving.
- Tenant-scoped: admins can only see and edit rows for their tenant.
- Backend: `GET/POST/PATCH/DELETE` under `/api/admin/registered-clients`, using `RegisteredClientRepository` for reads + new methods for writes.

## Out of scope

- Cross-tenant management.
- Audit log table (DB logs are fine for v1).
- Self-service registration.

## Once shipped

- Use the UI to seed any environments where the table is incomplete.
- Close #151.

## Implementation references

- Schema: `database/migrations/0008_registered_clients.sql`
- Validation logic: `core/src/repositories/registered_client.rs`
- Strict mode call sites: `api/src/api/http/oauth.rs:529` (GET) and `api/src/api/http/oauth.rs:2225` (POST)

dcadenas:
## Summary

Managing the OAuth client allowlist in `registered_clients` required raw SQL against production.
This change adds an admin UI and a tenant-scoped HTTP API so a full admin can list, add, edit, and delete entries from the browser.
The same canonical wildcard matcher used by the OAuth validator is exposed as a `test` endpoint so the UI's inline tester reports exactly what a real OAuth request would see.

Closes #164.

## What Changed

The repository, HTTP layer, and a new admin page were added end-to-end.
The OAuth validator code path was not touched - only how the table is written changed.

- core: `RegisteredClient` struct and tenant-scoped `list / get / create / update / delete` on `RegisteredClientRepository`, plus a public `test_redirect_pattern` wrapper around the existing matcher.
- api: `GET/POST /api/admin/registered-clients`, `PATCH/DELETE /api/admin/registered-clients/:id`, and `POST /api/admin/registered-clients/test`. Gated by `is_full_admin` and `TenantExtractor`. `Duplicate` -> 409, `Integrity` -> 400, `NotFound` -> 404.
- web: New `/admin/registered-clients` page with list table, add form, inline edit, delete-with-confirm modal, and inline pattern tester. Header link and section card on `/admin`.

Review-cycle fixes bundled in this PR (each in its own commit):
- `matches_redirect_pattern` no longer panics on overlapping prefix/suffix (e.g. `aaa*aaa` + `aaaa`).
- `create` and `update` now trim each redirect URI before persisting.
- `validate_redirect_uri_list` rejects patterns with more than one `*`.
- `validate_name` is now called from both `create` and `update` (was asymmetric).
- `signer::sign_event_direct` now canonicalizes the unsigned event's pubkey through the same shared helper as the HTTP path, fixing the divine-blossom "Invalid signature" failure mode for the NIP-46 transport.
- `http_rpc_handler` mismatch warning was downgraded to debug so a stale-client outage does not double the structured-log volume.

## Testing

I rebased onto current main, ran the rust suites, the web build, and a red-green check on the new signer regression test.

- `cargo test -p keycast_api --test registered_clients_admin_test --test registered_clients_migration_test` - 17 passed.
- `cargo test -p keycast_core` - 50 passed.
- `cargo test -p keycast_signer --features integration-tests --lib test_sign_event_direct` - 3 passed.
- Red-green: commenting out `canonicalize_event_author` in `sign_event_direct` makes `test_sign_event_direct_canonicalizes_mismatched_pubkey` fail; restoring it makes it pass.
- `cargo clippy --all-targets` - clean.
- `cd web && bun run build` - succeeded.

End-to-end HTTP tests for the admin endpoints are not included; the handlers are thin (auth gate + repo call + error mapping) and the repo layer is covered.

## Risks

Delete is destructive and immediate by design - the issue explicitly calls this out and the UI surfaces it in a confirmation modal.
Other risks are limited because the OAuth validator code path is unchanged.

- A misconfigured pattern (e.g. `https://example.com/*`) only matches a single path segment. The UI tester is the recommended way to sanity-check before saving.
- `tenant_id` is `INTEGER` in the schema but `i64` in the API for codebase consistency; reads cast to `BIGINT` so Postgres handles widening on read.

### DIFF
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -12,7 +12,8 @@ use super::routes::AuthState;
 use crate::api::error::{ApiError, ApiResult};
 use crate::api::extractors::UcanAuth;
 use keycast_core::repositories::{
-    AuthEventRepository, ClaimTokenRepository, OAuthAuthorizationRepository, UserRepository,
+    test_redirect_pattern, AuthEventRepository, ClaimTokenRepository, OAuthAuthorizationRepository,
+    RegisteredClient, RegisteredClientRepository, RepositoryError, UserRepository,
 };
 use keycast_core::types::claim_token::generate_claim_token;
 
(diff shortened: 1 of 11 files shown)


NotThatKindOfDrLiz (approved):
Reviewed. I agree with the shape of this PR.

The tenant scoping and full-admin gate are implemented in the right place, the registered_clients repo/API/UI flow is coherent, and the bundled fixes are all substantive improvements rather than unrelated churn. The shared signer canonicalization fix is especially important; it closes a real production failure mode and the regression test appears to exercise the intended path.

I do not see a blocker here. I filed the remaining follow-ups separately:
- #169 negative HTTP-layer tests for the full-admin gate
- #170 durable audit trail for registered_clients admin changes
- #171 tenant_id width harmonization/documentation between schema and API

Approving.

PR #127, opened 2026-04-24
e2e: prevent process leaks and add OAuth/DPoP security regression coverage

sergey3bv:
## Summary

Closes #83. Delivers the Playwright OAuth security regression suite that was deferred from PR #67, plus the harness improvements that made it possible.

## What's new

**Test infrastructure (`e2e/helpers/`)**
- `keycast-process.ts` (new): spawn/kill the keycast binary with port-already-in-use guarding and SIGTERM → SIGKILL cleanup. Closes the stated process-leak / stale-port problem.
- `oauth.ts`: DPoP key material generation, DER→JOSE signature conversion, atproto OAuth client (`atprotoPar`, `atprotoAuthorize`, `atprotoExchangeCode`).
- `db.ts`: `markUserAtprotoReady`, `addRegisteredClient`, `deleteRegisteredClient`.

**Test cleanup**
- `oauth.spec.ts`: removes ~80 lines of `docker exec psql` + raw `curl` invocations; uses the existing `registerAndVerify` helper.
- `auth.spec.ts`: replaces a fixed 1500ms bcrypt-settling wait with a poll-with-retry loop.

**Security regression coverage (`oauth-security.spec.ts`, new) — five tests delivering #83**
1. ATProto OAuth code exchange returns DPoP-bound tokens
2. ATProto OAuth rejects missing and replayed DPoP proofs
3. Cross-tenant resource access is rejected for OAuth tokens
4. Strict mode rejects unknown OAuth clients (`REQUIRE_REGISTERED_OAUTH_CLIENTS=true`)
5. Production mode fails closed when email provider is missing

## Scope

`e2e/` only — no production code changes.

## CI note

`.github/workflows/build-test-push.yaml` currently runs only `cargo test`, `cargo clippy`, and `cargo fmt`. The new TypeScript regression suite is **not** executed in GitHub checks. The five tests above were validated locally; gating them in CI is tracked as #145.

## Follow-ups

- #142 — rename or comment hardcoded secret defaults in `keycast-process.ts` as test-only
- #143 — document or enforce sequential execution for shared-DB e2e suite
- #144 — scope the DER→JOSE helper to ES256/P-256 with a doc comment
- #145 — run the Playwright OAuth security suite in CI

PR #173, opened 2026-04-29
ci: run Playwright e2e OAuth security suite in GitHub checks

### LINKED ISSUE #145: ci: run Playwright e2e OAuth security suite in GitHub checks
Follow-up from PR #127.

`.github/workflows/build-test-push.yaml` currently runs only \`cargo test\`, \`cargo clippy\`, and \`cargo fmt\`. The Playwright e2e suite — including the five OAuth security regression tests delivered by #127 (DPoP code exchange, missing/replayed DPoP proofs, cross-tenant rejection, registered-client strict mode, production email fail-closed) — is **not** executed in GitHub checks.

Net effect: a regression in any of those security-sensitive paths would only be caught by an out-of-band manual e2e run.

## Suggested fix

Add a Playwright job to the workflow (or a separate scheduled workflow). Will need:
- Postgres + Redis services
- `bun install` in `e2e/`
- Master encryption key generation
- Built keycast binary
- `playwright test e2e/tests/oauth-security.spec.ts` (and ideally the rest of the e2e suite)

Probably worth doing as a nightly/weekly job initially given the runtime cost, then promoting to per-PR if it stays stable.

Tracking this so the new regression coverage actually gates merges instead of relying on local discipline.

dcadenas:
## Summary

The Playwright e2e suite was not running in GitHub checks.
The five OAuth security regression tests added in #127 — DPoP code exchange, missing/replayed DPoP proofs, cross-tenant rejection, registered-client strict mode, production email fail-closed — could only be caught by an out-of-band manual run.
This change adds a dedicated workflow that runs the full e2e suite on a nightly schedule plus manual dispatch.

## What Changed

New workflow `.github/workflows/e2e.yaml` brings up postgres, redis, and the relay as service containers, builds the debug keycast binary, runs migrations, installs the e2e bun deps and Playwright chromium, starts keycast in the background, and runs the whole `e2e/` suite.
The existing `build-test-push.yaml` is untouched so the cargo CI stays fast.

- All three service containers have health checks. The relay image (`ghcr.io/verse-pbc/relay_builder:latest`) is `debian:bookworm-slim` with only `ca-certificates`, so `nc`/`curl`/`wget` are absent; `bash` 5.2 is present, so the relay health command uses bash's `/dev/tcp` pseudo-device against port 8080. Without it the job would only gate on the relay container starting, and keycast would dial the relay before it accepts connections and enter a degraded reconnect state.
- Triggers: `schedule: cron "0 6 * * *"` (06:00 UTC nightly) and `workflow_dispatch`. No `pull_request` trigger yet — promote after a stable nightly streak.
- Migrations run via `./target/debug/keycast --migrate` (the same path prod GKE uses).
- `oauth-security.spec.ts` spawns additional keycast processes on ports 3410-3413; `KEYCAST_BINARY=./target/debug/keycast` is exposed so those spawns find the same binary.
- On failure the playwright-report and the keycast log are uploaded as artifacts.

## Testing

There is nothing to run on a developer machine for a workflow file; the real test is the first nightly run on GitHub.

- `actionlint .github/workflows/e2e.yaml` — exit 0.
- YAML parses with pyyaml.
- Relay health-check command tested end-to-end: the exact options string the YAML produces, passed to `docker run` against `ghcr.io/verse-pbc/relay_builder:latest`, transitions the container to `healthy` (exit 0); the same command against a closed port exits 1 with `Connection refused`.

## Risks

This is the suite's first CI run, so flakes are likely.
`workers: 1` and `retries: 2` in `playwright.config.ts` plus artifact upload on failure should make triage tractable.
If non-security specs prove flaky, the workflow can be narrowed to `e2e/tests/oauth-security.spec.ts` only.

- Cargo build on cold cache is ~10 minutes; warm cache (`Swatinem/rust-cache@v2`) drops toward 2 minutes.
- The relay health check assumes `bash` stays in `ghcr.io/verse-pbc/relay_builder` (currently `debian:bookworm-slim`); if upstream switches to a distroless runtime, the command needs updating.

Closes #145

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -47,6 +47,15 @@ jobs:
         image: ghcr.io/verse-pbc/relay_builder:latest
         env:
           IDLE_TIMEOUT: "0"
+        # relay_builder is debian:bookworm-slim with only ca-certificates added
+        # (see verse-pbc/relay_builder Dockerfile) — no nc/curl/wget. bash 5.2
+        # is present, so use its /dev/tcp pseudo-device to gate readiness on
+        # actual port-8080 acceptance, not just container start.
+        options: >-
+          --health-cmd "bash -c 'exec 3<>/dev/tcp/localhost/8080 && exec 3<&- && exec 3>&-'"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
         ports:
           - 8080:8080
 
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,150 @@
+# Runs the Playwright e2e suite (including the OAuth security regressions
+# from PR #127) on a nightly schedule plus on manual dispatch.
+#
+# Scheduled-only initially per issue #145. Promote to `pull_request` once the
+# suite is stable enough to gate merges.
+
+name: E2E Tests
+
+on:
+  schedule:
+    - cron: "0 6 * * *"
+  workflow_dispatch: {}
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  e2e:
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: keycast
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        ports:
+          - 5432:5432
+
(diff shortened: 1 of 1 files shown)


NotThatKindOfDrLiz (approved):
Reviewed. I agree with the shape of this PR. This is the right posture for first-time e2e CI: nightly/manual only, careful service readiness checks, artifact upload on failure, and no attempt to gate pull requests on a suite that has not yet earned that trust.

I do not see a blocker here. I filed the remaining follow-ups separately:
- #177 absolute KEYCAST_BINARY path for workflow robustness
- #176 optional Playwright browser caching if nightly runtime matters

Approving.

Issue #170, opened 2026-04-29
chore(admin): add durable audit trail for registered_clients admin changes


NotThatKindOfDrLiz:
PR #168 logs registered_clients create/update/delete actions through structured application logs.

Follow-up:
- decide whether allowlist changes need a durable audit record beyond tracing output
- if yes, add an audit table or other durable mechanism for create/update/delete operations
- include actor identity, tenant id, target client id, and timestamp

This is non-blocking for the feature, but it would improve forensics and accountability for OAuth allowlist management.

PR #174, opened 2026-04-29
chore(api): harmonize tenant_id width between registered_clients schema and API

### LINKED ISSUE #171: chore(api): harmonize tenant_id width between registered_clients schema and API
PR #168 reads `registered_clients.tenant_id` as `BIGINT` into an `i64` API surface while the schema column remains `INTEGER`.

Follow-up:
- decide whether the API should use `i32` for this path or the schema should move to `BIGINT`
- if keeping the mismatch intentionally, document it where the repo/query layer lives
- ensure write-side behavior is explicit rather than relying on implicit narrowing assumptions

This is non-blocking because tenant ids are currently small, but the asymmetry is a code smell worth cleaning up.

dcadenas:
## Summary

The `registered_clients.tenant_id` column was `INTEGER` while every other `tenant_id` column in the schema is `BIGINT`. The Rust API surface uses `i64` everywhere, so PR #168 worked around the mismatch by casting `tenant_id::BIGINT AS tenant_id` on every read.

This change widens the column to `BIGINT` so the schema and the API surface agree. The casts go away and the implicit `i64 → i32` narrowing on writes goes away with them.

Closes #171.

## What Changed

The column type changes in a forward-only migration. The existing `0008_registered_clients.sql` has already been applied in production, so editing it would break checksum tracking — the new migration is idempotent (guarded on `information_schema.columns.data_type = 'integer'`) and safe to re-run.

- New migration `20260429000000_widen_registered_clients_tenant_id.sql` ALTERs the column to `BIGINT` only when its current type is `integer`.
- `core/src/repositories/registered_client.rs` drops the four `tenant_id::BIGINT AS tenant_id` casts (in `list`, `get`, `create RETURNING`, `update RETURNING`) and the doc comment that explained the prior INTEGER/i64 mismatch.
- New regression test `registered_clients_tenant_id_is_bigint` queries `information_schema.columns` and asserts the post-migration type.

## Testing

I ran the affected suites against a fresh local Postgres test database, plus clippy and fmt over the whole workspace.

- `cargo test -p keycast_api --test registered_clients_admin_test` — 16 passed (CRUD round-trip, including `tenant_id` equality on returned rows).
- `cargo test -p keycast_api --test registered_clients_migration_test` — 2 passed (existing seed test + new BIGINT type assertion).
- `cargo test -p keycast_core` — 50 passed.
- `cargo clippy --all-targets --all-features` — clean.
- `cargo fmt --all -- --check` — clean.
- Red-green: confirmed the new migration test fails on the pre-migration schema and passes after the migration runs.

I did not run the full e2e suite or test against a Cloud SQL instance — both rely on infra I cannot reach from this worktree.

## Risks

The migration is a widening on an in-production table. PostgreSQL rewrites the column in place (small table, fast) and no values can be lost going `INTEGER → BIGINT`.

- The migration must run before the new binary deploys, otherwise the cast-free SELECTs would try to decode an INT4 column into `i64`. Both deployment paths preserve this ordering: cloudbuild.yaml runs the migration step before the build/deploy steps, and the binary's `--migrate` runs migrations before serving.
- Per memory `project_migration_pipeline_bug.md`, the cloudbuild psql glob `database/migrations/00[1-9][0-9]*.sql` does NOT match `2026*` filenames, so prod actually applies recent migrations through the binary's `--migrate` (sqlx::migrate) path rather than the cloudbuild psql step. That bug is out of scope for #171 — surfaced as a follow-up on #170 already.

### DIFF
--- a/api/tests/registered_clients_migration_test.rs
+++ b/api/tests/registered_clients_migration_test.rs
@@ -35,3 +35,27 @@ async fn migrations_register_divine_invite_admin_oauth_client() {
         .await
         .is_err());
 }
+
+#[tokio::test]
+async fn registered_clients_tenant_id_is_bigint() {
+    // Every other tenant_id column in this schema is BIGINT (tenants.id,
+    // users.tenant_id, oauth_authorizations.tenant_id, etc.). 0008 originally
+    // declared registered_clients.tenant_id as INTEGER; the follow-up widens
+    // it so the API i64 surface no longer needs ::BIGINT casts on read.
+    let pool = common::setup_test_db().await;
+
+    let data_type: String = sqlx::query_scalar(
+        "SELECT data_type FROM information_schema.columns
+         WHERE table_schema = 'public'
+           AND table_name = 'registered_clients'
+           AND column_name = 'tenant_id'",
+    )
+    .fetch_one(&pool)
+    .await
+    .expect("registered_clients.tenant_id column must exist");
+
+    assert_eq!(
+        data_type, "bigint",
+        "registered_clients.tenant_id should be bigint to match all other tenant_id columns",
+    );
+}
--- a/core/src/repositories/registered_client.rs
+++ b/core/src/repositories/registered_client.rs
@@ -8,10 +8,6 @@ use sqlx::{FromRow, PgPool};
 use crate::repositories::RepositoryError;
 
 /// A registered OAuth client row.
-///
-/// Note: the database column `tenant_id` is `INTEGER NOT NULL` (i32) but is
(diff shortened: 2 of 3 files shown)


NotThatKindOfDrLiz (approved):
Reviewed. I agree with the shape of this PR. This is the right fix for the tenant_id width mismatch: a widening migration, removal of the compensating casts, and a regression test that locks the schema expectation in place.

I do not see a blocker here. The rollout and rollback story is straightforward, and the surrounding cloudbuild psql glob bug is useful context but not something this PR needs to solve.

Approving.

PR #178, opened 2026-04-29
chore(admin): add durable audit trail for registered_clients admin changes

### LINKED ISSUE #170: chore(admin): add durable audit trail for registered_clients admin changes

dcadenas:
## Summary

Admin changes to the OAuth client allowlist only emitted log lines.
Logs age out and aren't joinable by tenant or actor, so we couldn't answer "who changed what client, and when" without re-grepping.
This change writes a durable row to a new `admin_audit_events` table for every create / update / delete.
The update path uses `SELECT ... FOR UPDATE` then `UPDATE` in one transaction so the audit row carries the actual `{before, after}` transition this admin produced — even when other admins update concurrently.

Closes #170.

## What Changed

The change spans the database, the repository layer, and the admin handlers.
The table is intentionally generic so future admin actions can reuse it.

- New `admin_audit_events` table (id, occurred_at, tenant_id, actor_pubkey, action, target_resource_type/id, target_client_id, metadata_json) with indexes on `(tenant_id, occurred_at)`, `(tenant_id, target_resource_type, target_resource_id)`, and `(action, occurred_at)`. New `AdminAuditEventRepository` with `record()` and `list_recent()`.
- `RegisteredClientRepository::delete` uses `DELETE ... RETURNING` and returns the deleted row, so the handler can audit the pre-delete state without a race.
- `RegisteredClientRepository::update` runs `SELECT ... FOR UPDATE` then `UPDATE` inside a single transaction and returns `RegisteredClientUpdate { before, after }`. Under READ COMMITTED, the row lock waits for any concurrent updater and re-anchors on the latest committed version, so `before` is the actual immediately-prior state and `after` is this admin's update applied on top.
- `create_registered_client`, `update_registered_client`, and `delete_registered_client` each write an `admin_audit_events` row after the repo call. Writes are best-effort: a failed insert logs and the admin action still succeeds (matches the `auth_observability` pattern). Updates use a dedicated helper that emits `{before, after}` metadata.
- All three tracing logs include `(by admin {pubkey-prefix})`, so the actor is captured in logs even when the best-effort audit insert fails.

Only the registered_clients endpoints write rows in this PR. The schema is shaped so claim-token regenerate, support-admin add/remove, preload-user, and friends can reuse it later.

## Testing

I added two repo-level tests (record + tenant-scoped list; update before/after snapshots) and one HTTP-handler test, and reran existing coverage.

- `cargo test -p keycast_core` — 50 + 2 + 1 pass.
- `cargo test -p keycast_api --features integration-tests --test registered_clients_admin_test --test registered_clients_audit_test --test registered_clients_migration_test` — 18 + 1 + 1 pass. Includes `update_returns_before_and_after_snapshots` (asserts the repo returns both snapshots and `updated_at` advances) and the HTTP-level test asserting create/update/delete each emit one row with the right action, target, and metadata (the update row carries `metadata_json.before` and `metadata_json.after`).
- `cargo clippy --workspace --all-targets --all-features -- -D warnings` — clean.

I did not add a deterministic concurrency test for the row lock; forcing the race window requires synchronization primitives across two connections. The `SELECT ... FOR UPDATE` semantics are standard, and the existing tests assert the public contract.
I did not exercise this against the live signer or a deployed environment.

## Risks

Behavior change is small and confined to admin endpoints.

- Audit insert is best-effort. Postgres errors get logged and the admin call still returns 200; this matches `auth_observability` and avoids tying admin availability to audit availability.
- The `update` call now opens a short transaction and holds a row lock, serializing concurrent updates of the same row. Admin updates are infrequent, so this is acceptable.
- No UI surface for the audit log yet. Querying is SQL-only until we want to expose it.
- No `request_id` linkage to API request traces yet — the `auth_observability` middleware already populates one; wiring it through the admin handlers is a small follow-up.

### DIFF
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -1761,10 +1761,12 @@ pub async fn delete_registered_client(
     )
     .await;
 
+    let admin_display = &auth.pubkey[..8];
     tracing::info!(
-        "Registered client deleted: id={} client_id={}",
+        "Registered client deleted: id={} client_id={} (by admin {})",
         deleted.id,
-        deleted.client_id
+        deleted.client_id,
+        admin_display
     );
     Ok(Json(serde_json::json!({ "deleted": true })))
 }
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -12,8 +12,9 @@ use super::routes::AuthState;
 use crate::api::error::{ApiError, ApiResult};
 use crate::api::extractors::UcanAuth;
 use keycast_core::repositories::{
-    test_redirect_pattern, AuthEventRepository, ClaimTokenRepository, OAuthAuthorizationRepository,
-    RegisteredClient, RegisteredClientRepository, RepositoryError, UserRepository,
+    test_redirect_pattern, AdminAuditEventRecord, AdminAuditEventRepository, AuthEventRepository,
+    ClaimTokenRepository, OAuthAuthorizationRepository, RegisteredClient,
+    RegisteredClientRepository, RepositoryError, UserRepository,
 };
 use keycast_core::types::claim_token::generate_claim_token;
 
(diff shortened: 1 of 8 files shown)


NotThatKindOfDrLiz (approved):
Reviewed. I agree with the shape of this PR. The generic admin_audit_events table is a reasonable reusable design, the before/after capture for registered_client updates is using the right transaction shape, and the tests cover the repo and HTTP layers in the places that matter.

I did make one small cleanup before approval: after #174 merged, this branch still carried stale tenant_id::BIGINT casts in registered_client queries. I dropped those in 1338ad9 so the repository code now matches the widened BIGINT schema consistently.

I do not see a blocker here. I filed the remaining non-blocking follow-ups separately:
- #184 add observability for failed admin audit event inserts
- #185 thread request_id through registered_clients admin audit events

Approving.

PR #181, opened 2026-04-29
chore: add repo contribution guardrails

NotThatKindOfDrLiz:
## Summary

- add repo-local contributor and agent guidance in `AGENTS.md`
- add semantic PR title enforcement via GitHub Actions
- add PR and issue templates so contributors get consistent semantic titles, testing prompts, and review expectations
- add explicit guidance to avoid naming sensitive external brands or partners in public GitHub artifacts unless a maintainer approves it

## Motivation

`keycast` already had strong repo-specific context in `CLAUDE.md` and a comprehensive build/test/push workflow, but it did not yet have repo-local contributor guardrails, semantic PR enforcement, or GitHub-native templates. This change adds those missing pieces without replacing the existing technical guidance, and it adapts the rules to the Rust workspace plus `web/` frontend rather than copying another repo’s details verbatim.

## Related Issue

- N/A

## Testing

- [ ] `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- [ ] `cargo fmt --all -- --check`

Manual verification:
- reviewed the new `AGENTS.md` for fit with the Rust workspace and `web/` frontend
- reviewed the semantic PR workflow and new GitHub templates for correctness
- confirmed the change only adds repo guidance and GitHub scaffolding

## Visuals

- [x] No visual change

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,1 +1,1 @@
-blank_issues_enabled: true
+blank_issues_enabled: false
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,1 @@
+blank_issues_enabled: true
--- a/.github/workflows/semantic_pr.yml
+++ b/.github/workflows/semantic_pr.yml
@@ -1,0 +1,13 @@
+name: Semantic PR
+
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  semantic-pull-request:
+    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@v1
+
--- a/.github/workflows/semantic_pr.yml
+++ b/.github/workflows/semantic_pr.yml
@@ -9,5 +9,4 @@ permissions:
 
 jobs:
   semantic-pull-request:
-    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@v1
-
+    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@c2e27845f65775943bc310b3ab89e5c5432b8d4d # v1
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,50 @@
+name: Bug Report
+description: Report a bug or regression in keycast.
+title: "fix: "
+labels:
+  - bug
+body:
+  - type: textarea
+    id: summary
+    attributes:
+      label: Summary
+      description: What is broken? Avoid naming corporate partners, customers, or other sensitive external brands; use generic descriptors unless explicitly approved.
+      placeholder: Briefly describe the bug.
+    validations:
+      required: true
+  - type: textarea
+    id: steps
+    attributes:
+      label: Steps to Reproduce
+      description: How can someone reproduce the issue?
+      placeholder: |
+        1. Go to ...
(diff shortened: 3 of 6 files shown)


.github/workflows/semantic_pr.yml
```diff
@@ -0,0 +1,13 @@
...
+on:
+  pull_request_target:
+    types: [opened, edited, synchronize, reopened]
+
+permissions:
+  pull-requests: read
+
+jobs:
+  semantic-pull-request:
+    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@v1
```
dcadenas: This reusable workflow is pinned to the floating `@v1` tag and invoked from a `pull_request_target` trigger.

https://github.com/divinevideo/keycast/blob/db541164eb4648f022adbe495c7c9bf5afe96f8d/.github/workflows/semantic_pr.yml#L1-L13

Floating tags are mutable by the upstream owner, so an upstream retag of `v1` (malicious or accidental) silently changes what runs on every PR event, including PRs from forks. The job-level `permissions: pull-requests: read` and the absence of `secrets: inherit` contain the blast radius, but `GITHUB_TOKEN` still carries that read permission under the base-repo context, and the reusable workflow can make arbitrary GitHub API calls with it.

Suggestion: pin to an immutable commit SHA and let Dependabot roll it forward. GitHub's own hardening guidance calls this out specifically for third-party actions used under `pull_request_target`.

```suggestion
    uses: VeryGoodOpenSource/very_good_workflows/.github/workflows/semantic_pull_request.yml@<40-char-sha> # v1
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1 @@
+blank_issues_enabled: true
```
dcadenas: `blank_issues_enabled: true` lets contributors open issues that skip `bug_report.yml` and `feature_request.yml` entirely, including the brand-name redaction guidance those templates carry at line 11 of each.

The wording in `AGENTS.md` and in the two templates reads as policy, not as a hint, so this line is a silent bypass introduced in the same diff. Two options:

- If the brand-name rule is a hard policy, flip this so every issue flows through a template:

```suggestion
blank_issues_enabled: false
```

- If the templates are advisory, the current setting is fine, but the tension with the AGENTS.md wording is worth making explicit somewhere (a comment in this file, or softer language in the templates).

dcadenas (approved):
🛡️ guardrails land cleanly — a couple of non-blocking suggestions inline.

Cross-repo follow-up worth flagging (out of scope for this PR): the brand-name avoidance rule is repeated in 4 places here — [`AGENTS.md:31`](https://github.com/divinevideo/keycast/blob/db541164eb4648f022adbe495c7c9bf5afe96f8d/AGENTS.md#L31), [`pull_request_template.md:5`](https://github.com/divinevideo/keycast/blob/db541164eb4648f022adbe495c7c9bf5afe96f8d/.github/pull_request_template.md#L5), [`ISSUE_TEMPLATE/bug_report.yml:11`](https://github.com/divinevideo/keycast/blob/db541164eb4648f022adbe495c7c9bf5afe96f8d/.github/ISSUE_TEMPLATE/bug_report.yml#L11), [`ISSUE_TEMPLATE/feature_request.yml:11`](https://github.com/divinevideo/keycast/blob/db541164eb4648f022adbe495c7c9bf5afe96f8d/.github/ISSUE_TEMPLATE/feature_request.yml#L11) — and applies to every divinevideo repo. Candidate for promotion to a shared home (e.g. `divine-plugin/rules/external-brand-discretion.mdc`) so each repo doesn't restate it.

NotThatKindOfDrLiz:
Addressed the two inline suggestions:

- pinned the reusable semantic PR workflow to the immutable commit behind v1 (c2e27845f65775943bc310b3ab89e5c5432b8d4d)
- disabled blank issues so issue intake matches the repo-local policy and template guidance

Follow-up commit: 0bc7c23

PR #196, opened 2026-04-30
fix(keycast): harden autoscaling shutdown for viral spikes

### LINKED ISSUE #275: Harden Keycast application shutdown for autoscaling scale-down
## Summary

Implement the Keycast-side shutdown hardening needed for safer autoscaling scale-down behavior.

## Motivation

During scale-down, Keycast can drop in-flight NIP-46 signing requests if readiness, HTTP drain, signer shutdown, and cleanup happen in the wrong order or without bounded phase budgets. That can create reconnect and retry storms during traffic changes.

## Acceptance Criteria

- Shutdown is split into explicit pre-drain, HTTP drain, signer drain, and cleanup phases.
- Phase budgets are configurable through environment variables and validated against a grace-period ceiling.
- HTTP, signer, and DB cleanup paths are bounded and abort or warn predictably on timeout.
- Unit tests cover shutdown timing parsing, validation, ordering, timeout, and abort behavior.
- The PR remains scoped to the Keycast application slice and does not claim to close broader infrastructure work.

dcadenas:
## Plain summary

On a scale-down event, keycast could drop in-flight NIP-46 sign requests before their responses reached the client. That can cause a reconnect storm right when we are trying to reduce instance count.

This PR reorders shutdown into explicit bounded phases so the platform stops routing traffic before keycast stops accepting, HTTP drains first, relay work drains while the relay client is still connected, cluster peers are told this instance is leaving, and cleanup remains inside the platform grace window. Every phase budget is env-var tunable and clamped at startup against the configured grace-period ceiling.

This is the keycast application slice of autoscaling hardening. Infrastructure-side HPA/PDB/topology/grace-period work belongs in the sibling IaC rollout.

> No UI change. No public brand/partner references in the diff.

## Motivation

Keycast holds two shutdown-sensitive pieces of per-instance state:

1. Long-lived NIP-46 WebSocket relay connections.
2. A bounded relay-worker queue for in-flight sign requests.

The old shutdown path could stop the relay client before queued requests had delivered responses, and some phases were not bounded tightly enough to fit the real deploy grace window. During scale-down, that could turn ordinary termination into dropped sign requests and client reconnect/retry amplification.

## Related Issue

Closes #275.

## What Changed

The shutdown sequence in `async_main` is now four explicit phases:

1. **Pre-drain**: set readiness to 503 and sleep `SHUTDOWN_PRE_DRAIN_SECS` so GKE endpoint/NEG detach can propagate before keycast stops accepting. Default: 15s. Cloud Run sets this to 0 because Cloud Run does not use readiness-probe-driven endpoint removal.
2. **HTTP drain**: notify axum to stop accepting and wait up to `SHUTDOWN_HTTP_DRAIN_SECS`. Default: 40s. On timeout, `drain_http_or_abort` aborts and awaits the axum task so later phases do not race a still-live accept loop.
3. **Cluster deregister + signer/relay drain**: publish the cluster leave event and deregister under a bounded slice of `SHUTDOWN_SIGNER_DRAIN_SECS`, close the relay queue to new work, drain relay workers while the relay client remains connected, then shut down the relay client and wait for tracked tasks. Default signer budget: 10s. The cluster deregister is capped at 2s and carved out of this same signer budget, so it adds no unmodeled overhang.
4. **Cleanup**: close the DB pool within `SHUTDOWN_TEARDOWN_MARGIN_SECS - 2s` so the final shutdown log still has headroom before SIGKILL. Default margin: 10s.

At startup, keycast resolves the configured plan and clamps phase durations so:

`pre_drain + http_drain + signer_drain + teardown_margin <= SHUTDOWN_GRACE_PERIOD_CEILING_SECS`

The built-in defaults target the intended GKE budget:

`15 + 40 + 10 + 10 = 75s`

Cloud Run is handled explicitly in `cloudbuild.yaml` with:

`CEILING=10, PRE=0, HTTP=3, SIGNER=3, MARGIN=4`

That gives Cloud Run a modeled 10s shutdown budget:

`0 + 3 + 3 + 4 = 10s`

Other supporting changes:

- `RelayQueue::close()` rejects new work during drain while preserving already-queued work for workers to process.
- Relay close/send are serialized through the queue close mutex so a request is either accepted before close and drained, or rejected after close.
- Shutdown-time queue rejections are counted as `keycast_nip46_queue_closed_total`, separate from overload drops.
- `ClusterCoordinator::deregister_and_stop_heartbeat()` explicitly publishes leave and deregisters during SIGTERM shutdown.
- Heartbeat and full-sync loops check cancellation while holding the registry lock or before polling Redis work, preventing shutdown-time re-registration races.
- Shutdown env parsing rejects values above the one-day per-field sanity ceiling to avoid overflow before validation can report an actionable warning.

## Testing

Author-reported verification:

- `cargo test -p keycast --bin keycast`
- `cargo test --workspace --lib --bins`

Additional takeover verification:

- `cargo test -p keycast --bin keycast` — 65 passed.
- `cargo test -p keycast_signer work_queue` — 7 passed.
- `cargo test -p cluster-hashring await_unless_cancelled` — 2 passed.
- `git diff --check origin/main...pr-196` — clean.

Not tested here: full kill-instance-and-observe-reconnect-spike acceptance. That needs the deploy environment and IaC rollout.

## Risks

- **Grace-period mismatch**: the defaults assume a 75s GKE grace. Cloud Run is explicitly overridden in `cloudbuild.yaml`; any future deploy path must set `SHUTDOWN_GRACE_PERIOD_CEILING_SECS` to its real post-SIGTERM grace.
- **Short Cloud Run drain window**: Cloud Run is constrained to about 10s, so its configured signer drain is intentionally small. The code prioritizes bounded, observable shutdown over pretending Cloud Run can run the GKE-sized budget.
- **Duplicate handling during handoff**: cluster leave is published before relay queue drain so peers can take over promptly. The design accepts at-least-once handoff behavior to avoid drops.
- **Slow Redis during deregister**: deregister is capped and shutdown proceeds; peers then converge through heartbeat staleness.
- **Stuck DB close**: pool close is bounded and logs a warning rather than blocking past platform SIGKILL.

## Visuals

- [x] No visual change

## Related Issue

Closes #275

### DIFF
--- a/keycast/src/main.rs
+++ b/keycast/src/main.rs
@@ -70,6 +70,21 @@ const DEFAULT_SHUTDOWN_HTTP_DRAIN_SECS: u64 = 45;
 /// `SHUTDOWN_SIGNER_DRAIN_SECS`.
 const DEFAULT_SHUTDOWN_SIGNER_DRAIN_SECS: u64 = 10;
 
+/// Conservative upper bound on the Deployment's `terminationGracePeriodSeconds`
+/// used at startup to sanity-check the shutdown phase budgets. The iac repo
+/// (TODO(#692): track in divinevideo/divine-iac-coreconfig sibling PR) targets
+/// 75s but the issue spec accepts 60–90s; 120s covers the full range with
+/// headroom. If an operator bumps grace higher they can raise this via
+/// `SHUTDOWN_GRACE_PERIOD_CEILING_SECS`. This constant does NOT change the
+/// kubelet behavior — it is purely a misconfig guardrail.
+const DEFAULT_SHUTDOWN_GRACE_CEILING_SECS: u64 = 120;
+
+/// Headroom reserved between the end of the phased drain and the kubelet
+/// SIGKILL for DB pool close, tracing flush, and other post-drain cleanup.
+/// Subtracted from the ceiling when validating that the configured phase
+/// budgets fit.
+const SHUTDOWN_TEARDOWN_MARGIN: Duration = Duration::from_secs(10);
+
 /// Parsed shutdown phase timings. See the `DEFAULT_SHUTDOWN_*` constants and
 /// the graceful-shutdown sequence in `async_main` for the phase breakdown.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -80,6 +95,15 @@ struct ShutdownTimings {
     http_drain: Duration,
     /// Maximum wait for signer + tracked background tasks.
     signer_drain: Duration,
+}
+
+impl ShutdownTimings {
+    /// Sum of all three phase durations. Does NOT include
+    /// `SHUTDOWN_TEARDOWN_MARGIN` — add that separately when comparing
+    /// against a grace-period ceiling.
+    fn total_budget(&self) -> Duration {
+        self.pre_drain + self.http_drain + self.signer_drain
+    }
 }
 
 /// Read a non-negative integer duration (in seconds) from the given env var,
@@ -118,6 +142,54 @@ fn parse_shutdown_timings() -> ShutdownTimings {
             "SHUTDOWN_SIGNER_DRAIN_SECS",
             DEFAULT_SHUTDOWN_SIGNER_DRAIN_SECS,
         ),
+    }
+}
+
+/// Parse the configured `terminationGracePeriodSeconds` ceiling against
+/// which the phased shutdown budget is validated at startup. Override with
+/// `SHUTDOWN_GRACE_PERIOD_CEILING_SECS` when the Deployment's grace period
+/// changes without a keycast release.
+fn parse_shutdown_grace_ceiling() -> Duration {
+    parse_duration_secs_env(
+        "SHUTDOWN_GRACE_PERIOD_CEILING_SECS",
+        DEFAULT_SHUTDOWN_GRACE_CEILING_SECS,
+    )
+}
+
+/// Check that `timings.total_budget() + SHUTDOWN_TEARDOWN_MARGIN <= ceiling`.
+/// The margin is the reserved headroom for DB pool close, tracing flush, and
+/// process teardown between the end of the phased drain and the kubelet
+/// SIGKILL at `terminationGracePeriodSeconds`. Exact equality is accepted
+/// (the margin fully fits); anything larger is rejected.
+///
+/// Returned `Err(String)` is operator-actionable and includes the current
+/// budget, margin, and ceiling so it can be dropped straight into a log
+/// line. We deliberately warn (not abort) on violation at the call site:
+/// the ceiling is cross-repo-coupled to the iac repo, and an operator may
+/// legitimately need to boot ahead of an iac-side grace bump.
+///
+/// TODO(#692): remove the soft-warning fallback in `async_main` once the
+/// sibling divinevideo/divine-iac-coreconfig PR lands and `terminationGracePeriodSeconds`
+/// is pinned at ≥75s in all environments.
+fn validate_shutdown_timings(timings: &ShutdownTimings, ceiling: Duration) -> Result<(), String> {
+    let total = timings.total_budget();
+    let required = total + SHUTDOWN_TEARDOWN_MARGIN;
+    if required > ceiling {
+        Err(format!(
+            "shutdown phase budget ({total_secs}s) + teardown margin ({margin_secs}s) \
+             = {required_secs}s does not fit inside SHUTDOWN_GRACE_PERIOD_CEILING_SECS \
+             = {ceiling_secs}s (pre_drain={pre}s, http_drain={http}s, signer_drain={sig}s); \
+             the kubelet will SIGKILL mid-drain on terminationGracePeriodSeconds",
+            total_secs = total.as_secs(),
+            margin_secs = SHUTDOWN_TEARDOWN_MARGIN.as_secs(),
+            required_secs = required.as_secs(),
+            ceiling_secs = ceiling.as_secs(),
+            pre = timings.pre_drain.as_secs(),
+            http = timings.http_drain.as_secs(),
+            sig = timings.signer_drain.as_secs(),
+        ))
+    } else {
+        Ok(())
     }
 }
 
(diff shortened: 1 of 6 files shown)


keycast/src/main.rs
```diff
@@ -47,6 +47,344 @@ use zeroize::Zeroizing;
...
 
+/// Duration the process pauses between flipping the readiness probe to 503
+/// and starting to drain accepted connections. This is the window during which
+/// Kubernetes observes the failing probe and removes the pod from the Service
+/// EndpointSlice so the load balancer stops routing new traffic to us. Keeping
+/// this out of the drain budget is critical — otherwise new NIP-46 sign
+/// requests arriving during these seconds would be rejected mid-request and
+/// trigger client-side reconnect storms on scale-down. Override with
+/// `SHUTDOWN_PRE_DRAIN_SECS` (e.g. `0` in local dev / tests).
+const DEFAULT_SHUTDOWN_PRE_DRAIN_SECS: u64 = 10;
```
NotThatKindOfDrLiz: This 10s default is appropriate for the Kubernetes EndpointSlice pre-drain described above, but it is unsafe if the still-present Cloud Run deploy path is used. `cloudbuild.yaml` still has `gcloud beta run deploy keycast --platform=managed` and does not set any `SHUTDOWN_*` overrides. In that path the process can spend the SIGTERM window sleeping here before HTTP/signer drain starts. Please either remove/disable/update the legacy Cloud Run deploy path, or set Cloud Run-safe shutdown env vars there (for example pre-drain 0 and budgets that fit its grace window).

keycast/src/main.rs
```diff
@@ -47,6 +47,344 @@ use zeroize::Zeroizing;
...
+const DEFAULT_SHUTDOWN_SIGNER_DRAIN_SECS: u64 = 10;
+
+/// Conservative upper bound on the Deployment's `terminationGracePeriodSeconds`
+/// used at startup to sanity-check the shutdown phase budgets. The iac repo
+/// (TODO(#692): track in divinevideo/divine-iac-coreconfig sibling PR) targets
+/// 75s but the issue spec accepts 60–90s; 120s covers the full range with
+/// headroom. If an operator bumps grace higher they can raise this via
+/// `SHUTDOWN_GRACE_PERIOD_CEILING_SECS`. This constant does NOT change the
+/// kubelet behavior — it is purely a misconfig guardrail.
+const DEFAULT_SHUTDOWN_GRACE_CEILING_SECS: u64 = 120;
```
NotThatKindOfDrLiz: This ceiling is only an env-configured assertion; it does not prove the deployed pod actually has a matching `terminationGracePeriodSeconds`. With the 120s default, startup can log that the budget fits even if the sibling IaC has not landed or if an alternate deploy path uses a shorter grace. Please make the IaC dependency explicit and verifiable before merge, or default/ship the ceiling and phase budgets to the actual deployed grace for every active deploy path.

NotThatKindOfDrLiz (changes requested):
Requesting changes on the deployment-safety boundary. The Rust shutdown helpers look solid, but this needs one of two things before merge: either remove/disable/update the still-present Cloud Run deploy path so it cannot deploy these Kubernetes-sized defaults, or prove/link the GKE/IaC rollout that sets terminationGracePeriodSeconds >= 75s as a merge prerequisite. Also please rebase onto current origin/main and rerun the normal gates; this PR head is old relative to main.

keycast/src/main.rs
dcadenas: Addressed in commit 6e8107e: cloudbuild.yaml now sets Cloud Run-safe SHUTDOWN_* values for the managed Cloud Run deploy path (ceiling 10s, pre-drain 0, HTTP 3s, signer 3s, teardown margin 4s). That keeps the full modeled shutdown budget within Cloud Run’s SIGTERM window instead of using the GKE-sized defaults.

keycast/src/main.rs
dcadenas: Addressed across commits a6c5106 and 6e8107e: the default grace ceiling is now 75s for the intended GKE deployment, and Cloud Run explicitly overrides the ceiling and phase budgets in cloudbuild.yaml. The startup clamp/validation now sizes the runtime phase plan to the configured platform ceiling rather than merely logging that an oversized plan was configured.

dcadenas:
@NotThatKindOfDrLiz
> Requesting changes on the deployment-safety boundary. The Rust shutdown helpers look solid, but this needs one of two things before merge: either remove/disable/update the still-present Cloud Run deploy path so it cannot deploy these Kubernetes-sized defaults, or prove/link the GKE/IaC rollout that sets terminationGracePeriodSeconds >= 75s as a merge prerequisite. Also please rebase onto current origin/main and rerun the normal gates; this PR head is old relative to main.

The deployment-safety boundary is now covered: Cloud Run has explicit safe shutdown env vars, the GKE defaults are sized to 75s, and the branch was rebased onto current origin/main with no conflicts. Fresh checks passed: cargo fmt --all -- --check, cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated, and bun run test after starting/migrating the local test DB.

NotThatKindOfDrLiz:
Takeover cleanup completed. I updated the PR title to `fix(keycast): harden autoscaling shutdown for viral spikes` and reconciled the PR body with the current code: 15/40/10/75/10 defaults, Cloud Run's explicit 0/3/3/10/4 budget, the phase-3 cluster deregister carve-out, queue-closed metric, and the current issue linkage to #275.

Verification from the takeover pass:

- `cargo test -p keycast --bin keycast` - 65 passed
- `cargo test -p keycast_signer work_queue` - 7 passed
- `cargo test -p cluster-hashring await_unless_cancelled` - 2 passed
- `git diff --check origin/main...pr-196` - clean
- semantic PR rerun after the title edit - success

Final diff pass is clean from my side. The remaining risk is operational, not a code blocker: future deploy paths need to set `SHUTDOWN_GRACE_PERIOD_CEILING_SECS` to the real post-SIGTERM grace if they do not use the Cloud Run overrides in `cloudbuild.yaml`.

NotThatKindOfDrLiz (approved):
Approved after takeover. I verified the Cloud Run budget fix, updated stale PR metadata, reran the semantic title check successfully, and completed the targeted local test pass documented in the PR comment. No blocking findings remain.

PR #204, opened 2026-05-04
fix: harden dangling OAuth policy handling and error classification

### LINKED ISSUE #141: chore(auth): decide policy enforcement for missing or dangling oauth_authorizations.policy_id
## Background

`oauth_authorizations.policy_id` has no foreign-key constraint to `policies.id`. As a result, an OAuth authorization can reference a non-existent `policy_id` (typo, stale, deleted policy). When the signer loads such an authorization, `oauth_auth.permissions(...)` returns an empty `Vec<Permission>`, and `validate_permissions_for_sign` / `validate_permissions_for_encrypt` / `validate_permissions_for_decrypt` fall through the `permissions.is_empty() → Ok(())` backward-compat branch.

Net effect: a typo'd, stale, or deleted `policy_id` silently disables policy enforcement.

PR #131's `test_18_no_permissions_allows_encrypt_and_decrypt` pins the analogous behavior for "valid policy_id with zero linked permissions" (intended backward-compat). PR #132 (`test_20_oauth_invalid_policy_id_allows_signing`) pins the dangling-FK case as a regression test for current behavior — but current behavior may not be desired behavior.

## Decision needed

Should an invalid / dangling `policy_id` fail closed (deny) or remain permissive (allow)? The current behavior is permissive; the question is whether that's intentional or a security gap.

## Possible fixes if we want fail-closed

1. Add a `FOREIGN KEY (policy_id) REFERENCES policies(id) ON DELETE RESTRICT` constraint via migration. Prevents the dangling state at the DB level.
2. Distinguish "dangling policy_id" from "valid policy with zero permissions" in the load path; deny on the former, allow on the latter (preserve backward-compat).
3. Both.

## Out of scope

- Whitelist/team-membership and other higher-level enforcement.
- Changing empty-permissions backward-compat for valid policies (PR #131's `test_18` explicitly pins that).

## References

- PR #131 — encrypt/decrypt enforcement + `test_18` no-permissions fallback
- PR #132 — regression test for invalid `policy_id` behavior
- `signer/src/signer_daemon.rs` — `validate_permissions_for_*` methods

sergey3bv:
## Summary

- fail-closed handling for missing policy_id;
- migration cleanup to make FK addition safe;
- regression test coverage update.

## Motivation

- A follow-up for #131 and #132

## Related Issue

- Closes #141

## Testing

## Visuals

- [x] No visual change

### DIFF
--- a/database/migrations/20260504124000_add_oauth_policy_fk.sql
+++ b/database/migrations/20260504124000_add_oauth_policy_fk.sql
@@ -1,0 +1,26 @@
+-- Fail closed for dangling oauth_authorizations.policy_id references.
+-- 1) Revoke active OAuth authorizations pointing at missing policies.
+-- 2) Clear dangling policy_id for all affected rows so FK can be added safely.
+-- 3) Add FK to prevent future dangling references.
+--
+-- Affected rows end up with policy_id NULL and revoked_at set. Application code
+-- must treat NULL policy_id as full access only when revoked_at IS NULL; otherwise
+-- a revoked row could be mistaken for an active unrestricted authorization.
+
+UPDATE oauth_authorizations oa
+SET revoked_at = CASE
+        WHEN oa.revoked_at IS NULL THEN NOW()
+        ELSE oa.revoked_at
+    END,
+    policy_id = NULL,
+    updated_at = NOW()
+WHERE oa.policy_id IS NOT NULL
+  AND NOT EXISTS (
+      SELECT 1
+      FROM policies p
+      WHERE p.id = oa.policy_id
+  );
+
+ALTER TABLE ONLY public.oauth_authorizations
+    ADD CONSTRAINT oauth_authorizations_policy_id_fkey
+    FOREIGN KEY (policy_id) REFERENCES public.policies(id) ON DELETE RESTRICT;
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -2888,6 +2888,35 @@ pub async fn get_authorization_for_origin(
     }
 }
 
+async fn load_policy_permissions_or_deny(
+    pool: &PgPool,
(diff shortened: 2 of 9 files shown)


database/migrations/20260504124000_add_oauth_policy_fk.sql
```diff
@@ -0,0 +1,22 @@
...
+-- 1) Revoke active OAuth authorizations pointing at missing policies.
+-- 2) Clear dangling policy_id for all affected rows so FK can be added safely.
+-- 3) Add FK to prevent future dangling references.
+
+UPDATE oauth_authorizations oa
+SET revoked_at = CASE
+        WHEN oa.revoked_at IS NULL THEN NOW()
+        ELSE oa.revoked_at
+    END,
+    policy_id = NULL,
```
dcadenas, line 15: Keep revoked OAuth rows out before any `NULL` policy branch can allow access.
This migration sets `revoked_at` and clears dangling `policy_id` to `NULL`, but the later origin lookup and signer reload can still evaluate that row as unrestricted access.
Add `revoked_at IS NULL` to the active OAuth authorization lookup and make signer permission validation reject any refetched OAuth authorization with `revoked_at` set.
Please cover both the origin-based HTTP validator and signer validation for a revoked `policy_id = NULL` row, while preserving active `NULL`-policy access and valid empty-policy access.

dcadenas (changes requested):
The migration still lets a revoked OAuth row reach permission checks as unrestricted access. Fix the active-authorization checks before merge so clearing `policy_id` to `NULL` never becomes the full-access path for revoked rows.

sergey3bv:
Hey, @dcadenas, I updated the PR according to your comment, could you please take a look

dcadenas:

PR #209, opened 2026-05-04
fix: reject malformed emails during registration

### LINKED ISSUE #202: fix: reject malformed emails during registration
## Summary
Keycast should reject malformed email addresses during registration before storing pending signup state or trying to send verification email.

## Environment
- Service: Keycast API
- Endpoints: `/api/headless/register`, `/api/auth/register`

## Steps to Reproduce
1. Call a registration endpoint with an email like `person@gmail..com`.
2. Use an otherwise valid password and client request body.
3. Observe that registration can proceed into the email verification flow.

## Actual Result
The server lowercases the email, stores pending registration state, attempts to send a verification email, and can return success even when the email address is malformed and cannot receive mail.

## Expected Result
The server rejects malformed email addresses with a stable 400-style error code and message before creating pending registration state or attempting email delivery.

## Regression?
Unknown.

## Evidence
Observed via Flutter signup report: a user typed `gmail..com` instead of `gmail.com`, did not receive the verification email, and the app had already advanced to the check-email flow.

Relevant code inspected:
- `api/src/api/http/headless.rs`: `headless_register` lowercases `req.email` but does not validate format.
- `api/src/api/http/auth.rs`: `register` lowercases `req.email` but does not validate format.

## Acceptance Criteria
- [ ] `/api/headless/register` rejects malformed email addresses before storing pending registration state.
- [ ] `/api/auth/register` rejects malformed email addresses before creating user or verification state.
- [ ] Invalid domains like `gmail..com` are rejected.
- [ ] Valid common emails still pass.
- [ ] Responses include a stable error code/message that clients can map.
- [ ] Tests cover valid and invalid email formats for both registration paths.

dcadenas:
## Summary

Users could get stuck after entering an email address that could not receive mail.
This change rejects malformed registration emails before signup state is stored or verification email is sent.
The same check now protects standard registration, headless registration, and OAuth popup registration.

## Motivation

The registration paths only lowercased email input before doing side-effect work.
The shared validator now checks the raw trimmed input as ASCII before lowercasing.
It rejects bad dot placement, invalid local characters, invalid domain labels, edge hyphens, length violations, and Unicode case-folding bypasses.

## Related Issue

- Closes #202

## Testing

I tested the validator, the affected registration paths, and the full Rust workspace.
Manual browser verification was not run because this has no visual change.
Fetching latest main failed because SSH auth is unavailable here; rebase against cached `origin/main` was up to date.

## Visuals

- [x] No visual change

## Risks

The validator is intentionally conservative.
It supports common deliverable ASCII addresses, but it does not support quoted local parts or internationalized domains.
That trade-off is safer for registration because users must receive a verification email to finish signup.

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -106,14 +106,17 @@ fn normalize_nip05_username(raw_username: &str) -> Result<String, AuthError> {
 }
 
 pub(crate) fn normalize_registration_email(email: &str) -> Result<String, &'static str> {
-    let normalized = email.trim().to_lowercase();
+    let trimmed = email.trim();
 
-    if normalized.is_empty()
-        || normalized.len() > 254
-        || normalized.bytes().any(|byte| byte <= b' ' || byte == 0x7f)
+    if trimmed.is_empty()
+        || trimmed.len() > 254
+        || !trimmed.is_ascii()
+        || trimmed.bytes().any(|byte| byte <= b' ' || byte == 0x7f)
     {
         return Err(INVALID_EMAIL_CODE);
     }
+
+    let normalized = trimmed.to_ascii_lowercase();
 
     let Some((local, domain)) = normalized.split_once('@') else {
         return Err(INVALID_EMAIL_CODE);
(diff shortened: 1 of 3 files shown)


NotThatKindOfDrLiz (approved):
Reviewed the actual PR commit scope. This is a narrow security hardening change that closes the Unicode email normalization bypass and adds direct regression coverage without introducing new tech debt in the touched code.

PR #218, opened 2026-05-06
docs(keycast): document health probe ordering and shutdown semantics

### LINKED ISSUE #136: docs(health): document health probe invariants and shutdown semantics
PR #93 changes health probe behavior, but some important invariants and ordering assumptions are only implicit in the code.

Requested follow-up:
- add a short comment near `/healthz/startup` explaining that it is only safe because the HTTP listener binds after `Database::new().await` and migrations complete
- add a short comment in the shutdown/readiness path explaining that DB checks are intentionally skipped during shutdown and the outer `Shutting down` response wins
- add a short comment noting that Cloud Run does not use readiness probes for traffic routing, so `/healthz/ready` is mainly useful for smoke tests and Kubernetes-style consumers

Context:
- These were raised as non-blocking inline review notes on PR #93.

This is non-blocking for PR #93 but worth documenting so the current behavior does not depend on silent assumptions.

sergey3bv:
## Summary

- Documenting the reason `healthz/startup` is safe without a DB check.

## Motivation

- A follow-up for #93 

## Related Issue

- Closes #136

## Testing

## Visuals

- [x] UI/web change with screenshots/video attached
- [x] No visual change

### DIFF
--- a/keycast/src/main.rs
+++ b/keycast/src/main.rs
@@ -58,6 +58,9 @@ struct ReadinessState {
     shutting_down: Arc<AtomicBool>,
 }
 
+/// Maps probe state to HTTP status. Shutdown is checked before DB readiness so
+/// during graceful shutdown we skip the DB query in [`readyz`] yet still return
+/// 503 with body `"Shutting down"` (not `"Database unavailable"`).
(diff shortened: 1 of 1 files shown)


dcadenas:

Issue #231, opened 2026-05-14
Add audit trail for admin-issued preload UCAN signing operations


rabble:
## Context

PR [feat/preload-sign-after-claim] removes the `is_unclaimed` gate at `api/src/api/http/nostr_rpc.rs:359` so admin-issued preload UCANs can sign for users after they claim their accounts. This unblocks publishing additional legacy Vine archives discovered after handover.

That change expands what an admin token can do, but **does not add corresponding observability**. We currently have no durable record of which admin signed what as which user. The mint side logs (`tracing::info!("User token generated for pubkey: ... by admin: ...")`) but those are ephemeral — they roll out of Cloud Logging retention and are not queryable months later.

If a user reports a surprise event in their feed, we have no way to trace it back to which admin issued the signing UCAN that produced it.

## What we need

Per-sign audit row written from the signing path, ideally without slowing the hot path measurably.

### Schema (extend existing `admin_audit_events` table)

```sql
-- existing table per recent commits 135dc74, 95caa55
INSERT INTO admin_audit_events
  (action, admin_pubkey, target_pubkey, tenant_id, metadata, created_at)
VALUES
  ('preload_sign', '<from UCAN issued_by_admin fact>', '<user pubkey>', $tenant,
   '{"event_id": "...", "kind": 1, "ucan_jti": "..."}'::jsonb, NOW());
```

`metadata` JSON should include at minimum:
- `event_id` — the signed event id (so we can correlate with what landed on relays)
- `kind` — Nostr event kind
- `ucan_jti` or fingerprint — to group all signs from a single issued UCAN
- `created_at` of the signed event (vs. row insert time, in case of backdating)

### Insertion site

Best place: inside `HttpRpcHandler::sign_event` (or whatever the equivalent is for `encrypt`/`decrypt` — those should be audited too) when the handler was created from a server-signed preload UCAN. The handler already knows whether it was loaded via `load_preloaded_user_handler` vs `load_handler_on_demand` (OAuth path); we should plumb a tag through so we only audit preload-mode operations, not OAuth operations (those have their own per-authorization records).

The `issued_by_admin` field is already in the UCAN facts (set at `admin.rs:413` in `generate_preload_ucan`), so we can read it from the UCAN at handler-load time and stash it on the handler struct.

### Performance

- The cache hit path (`get_handler` fast path, `nostr_rpc.rs:439`) cannot afford a synchronous DB write per sign. Either:
[...]

PR #232, opened 2026-05-14
feat(rpc): allow admin-issued preload UCANs to sign for claimed users

rabble:
## Summary

- Drop the `is_unclaimed` gate in `load_preloaded_user_handler` (`api/src/api/http/nostr_rpc.rs`) so admin-issued preload UCANs continue to work after a user claims their account.
- Existence check preserved — non-existent users still rejected with `InvalidToken`.
- Updated docstrings on `load_preloaded_user_handler` and the MODE 2 branch in `get_handler` to reflect the new behavior.

## Why

We've discovered more classic Vine archives belonging to accounts that users have already claimed, and users want them published under their accounts. The mint side (`/api/admin/preload-user`) was already idempotent on `vine_id` and happily returned a fresh signing UCAN regardless of claim status; the signing endpoint at `nostr_rpc.rs` was the sole remaining block.

## Behavior change

Before: a server-signed preload UCAN could only sign for users where `email IS NULL`. Once a user claimed, their preload UCANs became dead.

After: the same preload UCAN works for both unclaimed and claimed users, until the UCAN itself expires (30 days) or gets evicted from cache.

The mint gate is unchanged: only `is_full_admin` callers can mint these UCANs, so the externally-reachable attack surface is the same.

## What this expands

A leaked admin token can now sign as any user (claimed or not) for as long as a 30-day preload UCAN lives. Before, claim was a user-side seal that closed that door. Acceptable for our small ops team and the Vine republish workflow, but the threat model genuinely changed.

## Follow-ups (not in this PR)

- #231 — add per-sign audit trail to `admin_audit_events` so we can answer "who signed what as whom" months after the fact. Schema + insertion-site notes are in the issue.
- Consider shortening preload UCAN TTL from 30d → 24h to cut the leaked-token window.
- `get_user_token` (pubkey-based admin endpoint at `admin.rs:344-393`) still has its own `is_unclaimed` block. Inconsistent with the new vine_id behavior, but the import script doesn't use that path so it's not blocking. Worth aligning later.

## Test plan

- [x] Build passes (`cargo build -p keycast_api`)
- [ ] Existing `admin_preload_test`, `admin_token_test`, `admin_batch_claim_test` still pass (they test mint-side and `is_unclaimed` repository behavior, both unchanged)
- [ ] Manual: in staging, mint a preload UCAN for a claimed test account, hit `/api/nostr` with `sign_event`, verify it returns a signed event instead of `InvalidToken`
- [ ] Manual: confirm preload UCAN for nonexistent pubkey still returns `InvalidToken`

### DIFF
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -391,8 +391,9 @@ fn is_server_signed(ucan: &ucan::Ucan) -> bool {
 }
 
 /// Load handler for preloaded user (server-signed UCAN, no bunker_pubkey)
-/// These are users imported from Vine that haven't claimed their accounts yet.
-/// They have no policy restrictions - full access until account is claimed.
+/// Originally for unclaimed Vine imports; gate dropped to allow admin-issued
+/// preload UCANs to keep signing for accounts after claim — needed to publish
+/// additional legacy Vine archives discovered after handover.
 async fn load_preloaded_user_handler(
     auth_state: &AuthState,
     pool: &sqlx::PgPool,
@@ -402,17 +403,16 @@ async fn load_preloaded_user_handler(
 ) -> Result<Arc<HttpRpcHandler>, RpcError> {
     let key_manager = auth_state.state.key_manager.as_ref();
 
-    // Verify user exists and is unclaimed (email IS NULL)
+    // Verify user exists (encrypted key lookup below will also catch missing rows).
     let user_repo = UserRepository::new(pool.clone());
-    let is_unclaimed = user_repo
+    if user_repo
         .is_unclaimed(user_pubkey_hex, tenant_id)
         .await
-        .map_err(|e| RpcError::Internal(format!("Database error: {}", e)))?;
-
-    if is_unclaimed != Some(true) {
-        // User either doesn't exist or has already claimed their account
+        .map_err(|e| RpcError::Internal(format!("Database error: {}", e)))?
+        .is_none()
+    {
         tracing::warn!(
-            "Preloaded user RPC denied: user {} not unclaimed",
+            "Preloaded user RPC denied: user {} not found",
             &user_pubkey_hex[..8]
         );
         return Err(RpcError::Auth(AuthError::InvalidToken));
@@ -563,7 +563,8 @@ async fn get_handler(
         h
     } else if is_server_signed(&ucan) {
         // MODE 2: Preloaded user - server-signed UCAN without bunker_pubkey
-        // These are Vine-imported users who haven't claimed their accounts yet
+        // Used for Vine import; works for both unclaimed and claimed accounts so
+        // admins can publish additional legacy archives after handover.
         let h = load_preloaded_user_handler(
             auth_state,
             pool,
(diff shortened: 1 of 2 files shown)


irab (approved):
Looks fine - definitely would be good to change that expiry from 30D to however long the import script normally takes to run. I assume 1-4hrs?

api/src/api/http/nostr_rpc.rs
```diff
@@ -563,7 +563,8 @@ async fn get_handler(
         h
     } else if is_server_signed(&ucan) {
```
dcadenas, line 591: Restrict this branch to real preload UCANs before loading personal keys.
Right now any server-signed UCAN without `bunker_pubkey` reaches `load_preloaded_user_handler`. That means claim-session tokens, or the admin UCAN returned by `GET /api/admin/token`, can become full-access RPC signing tokens for whoever the UCAN's audience is, provided `personal_keys` exists for them.
Require preload-specific facts, such as `redirect_origin == "preload"` and `issued_by_admin`, or add a dedicated token-purpose fact before this path loads the handler.
Add tests that accept an admin preload UCAN for a claimed user and reject claim/admin server-signed no-bunker tokens.

dcadenas (changes requested):
The new preload signing path needs tighter auth checks before this can merge.
It must only accept real preload UCANs, and cached preload handlers must stop at the UCAN expiry.

The PR description bounds the leaked-token window at 30 days. With the cache-expiry issue below, a warm cache pushes that to effectively unbounded, so the stated tradeoff doesn't hold yet.

---
**Additional findings not shown inline**

- **Carry UCAN expiry into cached preload handlers**
  https://github.com/divinevideo/keycast/blob/dd4d4e486c8e42467a73be6a2c06485d0516e46a/api/src/api/http/nostr_rpc.rs#L453
  Carry the UCAN expiry into cached preload handlers.
This path caches preload handlers with `expires_at: None`, so after the first valid use, later cache hits return the handler without revalidating the UCAN.
The cache uses `time_to_idle(3600)`, which resets on every access, so a token hit at least once per hour never evicts. A warm cache entry can keep signing past the 30-day UCAN lifetime until process restart, effectively unbounded for continuously-used tokens.
Store the UCAN expiration on the handler or cache entry, and add a test that a warm cache rejects the token after expiry.

rabble:
@dcadenas Both findings addressed in df843b9:

**1. Real preload UCANs only** — `load_preloaded_user_handler` now takes the UCAN's `redirect_origin` as a parameter and rejects anything that isn't `"preload"`. Server-signed admin-session UCANs and claim UCANs can no longer fall through MODE 2 to sign as users.

**2. UCAN expiry carried into cached handler** — `get_handler` now extracts `ucan.expires_at()` and threads it into `HttpRpcHandler::new` instead of passing `None`. The cache hit path's `handler.is_valid()` check will now return false once the UCAN lifetime ends, so warm cache entries can't keep signing past `exp`.

**Tests added** (`api/tests/nostr_rpc_integration_test.rs`):

- `test_warm_cache_preload_handler_rejected_after_ucan_expiry` — exactly what you asked for. Mints a preload UCAN with `lifetime: 2s`, exercises a successful sign to populate the cache, sleeps 3s past expiry, then asserts the second call (cache hit) returns `InvalidToken` and the cache entry is evicted.
- `test_server_signed_non_preload_redirect_origin_rejected` — mints a server-signed UCAN with `redirect_origin: "admin"` and verifies the signing path rejects it before touching the user's key.

Both new tests pass. The 6 pre-existing tests in this file still pass, as do the adjacent `admin_preload_test`, `admin_token_test`, and `admin_batch_claim_test` suites.

dcadenas (approved):
🔑 claimed-user preload signing lands cleanly

PR #235, opened 2026-05-18
feat: account-level suspension for under-16 age review

mbradley:
## Summary

- Adds `status` (active/suspended/banned), `suspended_reason`, and `suspended_at` columns to the `users` table
- Auth flow includes `account_status` in UCAN token facts and account status response when non-active
- Service-token admin endpoints (`GET/PUT /api/admin/users/:pubkey/status`) for relay-manager to suspend/unsuspend accounts
- Signer daemon refuses to sign events or encrypt/decrypt for non-active users (DB query per request, no cache)
- Suspended and banned users can still log in to see restriction/ban notices

Relates to divinevideo/support-trust-safety#142

## Context

Track C of the under-16 system coordination doc. The relay-manager side (case table, state machine, admin queue) and the mobile client (PR #3531 with `restrictedMinorReview`) are already built. This gives Keycast the ability to mark accounts as suspended and enforce that through the signer.

Design spec: `docs/superpowers/specs/2026-05-18-account-suspension-design.md`

## Test plan

- [ ] Run `bun run db:reset` and verify new columns exist
- [ ] `cargo test --lib` -- all 184 unit tests pass
- [ ] Integration tests for admin endpoints (7 tests in `user_status_admin_test.rs`)
- [ ] Manual: suspend a user via admin PUT, verify `get_account_status` returns status
- [ ] Manual: verify suspended user can still log in but signer refuses `sign_event`
- [ ] Manual: unsuspend user, verify signing works again
- [ ] Verify `KEYCAST_SERVICE_TOKEN` env var is set in staging before deploy

### DIFF
--- a/api/src/api/http/oauth.rs
+++ b/api/src/api/http/oauth.rs
@@ -531,7 +531,7 @@ pub async fn auth_status(
 
         // Return authenticated with pubkey, optionally with email info if user exists in DB
         // NIP-07 admins may not have a user record, but their UCAN session is still valid
-        if let Some((email, email_verified)) = user_info {
+        if let Some((email, email_verified, _status, _reason)) = user_info {
             Ok(Json(AuthStatusResponse {
                 authenticated: true,
                 pubkey: Some(user_pubkey),
@@ -625,7 +625,7 @@ pub async fn authorize_get(
                     tracing::warn!("UCAN cookie has pubkey {} but user doesn't exist in tenant {}, clearing stale cookie", pubkey, tenant_id);
                     (None, true, None) // User was deleted, clear the cookie
                 }
-                Some((email, _verified)) => (user_pubkey, false, email),
+                Some((email, _verified, _status, _reason)) => (user_pubkey, false, email),
             }
         }
     } else {
@@ -2453,6 +2453,7 @@ async fn handle_refresh_token_grant(
         &auth_state.state.server_keys,
         false, // Refresh tokens are not first-party
         None,
+        None,
     )
     .await
     .map_err(|e| OAuthError::InvalidRequest(format!("UCAN generation failed: {:?}", e)))?;
@@ -2878,6 +2879,7 @@ async fn create_oauth_authorization_and_token(
         &auth_state.state.server_keys,
         is_headless, // first_party fact for headless flow tokens
         None,
+        None,
     )
     .await
     .map_err(|e| OAuthError::InvalidRequest(format!("UCAN generation failed: {:?}", e)))?;
@@ -3070,7 +3072,7 @@ pub async fn oauth_login(
 
     // Validate credentials
     let user_repo = UserRepository::new(pool.clone());
-    let (public_key, password_hash, email_verified) =
+    let (public_key, password_hash, email_verified, _user_status) =
         match user_repo.find_with_password(&req.email, tenant_id).await? {
             Some(user) => user,
             None => {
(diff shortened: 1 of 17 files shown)


signer/src/signer_daemon.rs
```diff
@@ -120,6 +120,25 @@ impl Nip46Handler {
         }
     }
 
+    /// Check that the user's account is active (not suspended or banned).
+    /// Queries the DB on each request for immediate enforcement.
+    async fn check_user_active(&self) -> SignerResult<()> {
```
dcadenas, line 126: Do not run the account-status check against `self.user_keys.public_key()` for regular team authorizations.

That key can be the stored signing key, not the user account pubkey, so `check_user_active()` can reject valid non-OAuth signing, encryption, and decryption requests with `User not found`.

Gate this check to OAuth or user-backed handlers, or pass the real user pubkey separately instead of using the signing key as the account identity.

dcadenas (changes requested):
The suspension work needs to cover every path that can still sign or issue user tokens before this can merge. The current implementation mixes user accounts and signing keys in one signer path, leaves the direct HTTP signing fallback unchecked, and lets several token issuance paths omit the new account-status fact.

---
**Additional findings not shown inline**

GitHub only allows PR review comments on lines present in the current diff hunk, so these accepted findings are included in the review body instead.

- **Check account status before the HTTP signing slow path**
  https://github.com/divinevideo/keycast/blob/e8dfe6c5c68c2c9bb4e1fd3560b355010dfd7cb8/api/src/api/http/auth.rs#L3210
  Check account status before the direct `unsigned_event.sign(&keys)` fallback.

The cached signer path has the new gate, but this slow path can still sign for a suspended or banned user with a valid UCAN when the handler cache is unavailable or misses.

Add the same `(tenant_id, user_pubkey)` status check before decrypting or signing with the personal key.
- **Centralize account status propagation for user token issuance**
  https://github.com/divinevideo/keycast/blob/e8dfe6c5c68c2c9bb4e1fd3560b355010dfd7cb8/api/src/api/http/oauth.rs#L2872
  Make user-token issuance require the current `UserStatus` instead of letting each call site pass `None`.

OAuth code exchange, refresh-token issuance, and popup or headless login paths can still issue UCANs for suspended or banned users without the `account_status` fact this PR added.

Fetch the status inside the user-token helper or require callers to pass it explicitly, then cover the OAuth, refresh, popup, and headless paths with tests that decode the UCAN facts.

api/src/api/http/oauth.rs
```diff
@@ -2453,6 +2453,7 @@ async fn handle_refresh_token_grant(
         &auth_state.state.server_keys,
         false, // Refresh tokens are not first-party
         None,
+        None,
```
dcadenas: Make user UCAN issuance require account status instead of letting callers pass `None`.
This refresh path passes `None`, and the OAuth authorization and popup login paths do the same, so suspended or banned users can still receive UCANs without an `account_status` fact.
Move the status lookup into the user-token helper or make the status argument non-optional for all user-token callers, while keeping admin and other non-user token paths separate.
Add coverage for refresh, OAuth code/headless issuance, and popup login with a suspended user.

dcadenas (changes requested):
The suspension feature still leaves signing and UCAN issuance paths open for suspended or banned users, so those gates need to be enforced before merge. The inline comments point to the two remaining bypasses that should be fixed in the authoritative paths.

---

- **Gate HTTP signing on account status before either signing path** ([api/src/api/http/auth.rs:3216](https://github.com/divinevideo/keycast/blob/2f5f24b716e952db26cba78d21cc3d4f89f0a3db/api/src/api/http/auth.rs#L3216))
  Validate the user's account status before either signing path can run.
  Right now `sign_event` can fall through to the DB/decrypt slow path and call `unsigned_event.sign(&keys)` without checking `users.status`.
  A suspended or banned user with a still-valid token can still get an event signed when the cached signer handler is missing or misses.
  Load the user status after extracting `tenant_id` and `user_pubkey`, deny non-active users before the cached-handler branch and the slow path, and add a test that covers a suspended user on the slow path.

dcadenas (approved):
🔒 account status gates are covered

PR #208, opened 2026-05-04
chore: Narrow generic 503 mappings for expected auth errors

dcadenas:
## Summary

Users could see a temporary outage when they actually hit a normal auth conflict or incomplete account state.
This change returns clear client errors for duplicate email, username conflicts, bad usernames, missing account keys, and incomplete registration state.
Retryable dependency failures still return service unavailable, but they now use targeted messages instead of falling through a generic internal error.

## Motivation

The root problem was broad error mapping in the auth layer.
Raw email verification SQL bypassed repository duplicate classification, and several expected auth states used `AuthError::Internal`.
The change keeps unknown database and unexpected crypto or signing failures generic, while giving known user and dependency outcomes their own mappings.

## Related Issue

- Closes #201

## Testing

I tested the affected regression paths and the full Rust workspace.
The full test run used the local Postgres test database.
`git fetch origin main` failed because GitHub SSH auth is unavailable here, so I rebased onto the existing local `origin/main`.

- [x] `DATABASE_URL=postgres://postgres:password@localhost:15432/keycast_test cargo test --workspace --verbose`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- [x] `cargo fmt --all -- --check`
- [x] Manual verification completed

## Visuals

There is no UI or web visual change.

- [ ] UI/web change with screenshots/video attached
- [x] No visual change
- [x] Visuals and text avoid sensitive external brand or partner names unless explicitly approved

## Risks

The main risk is over-classifying unexpected failures.
Database uniqueness handling only special-cases the known email constraint, and username conflicts now return a fixed bounded message instead of reflecting dependency text.
Unknown internal failures still return the generic 503 response.

PR #236, opened 2026-05-18
fix(auth): add EMAIL_ALREADY_EXISTS code to duplicate-email 409

### LINKED ISSUE #198: Handle duplicate email during OAuth email verification
## Summary
OAuth/headless email verification can return a generic 503 when the pending registration email already exists in the tenant.

## Evidence
Cloud Run prod logs showed repeated failures around 2026-05-02T12:12-12:15Z:

```text
POST /api/auth/verify-email 503
Database error: duplicate key value violates unique constraint "idx_users_email_tenant"
```

The failure repeated on retry because the pending OAuth verification record was not resolved after the insert failed.

## Suspected Cause
`verify_email` checks whether the user exists by pubkey before inserting into `users`, but the unique constraint that fails is email+tenant:

```rust
let user_already_exists = user_repo.exists(&oauth_data.user_pubkey, tenant_id).await?;
```

If another user already has `pending_email`, the insert into `users` fails with `idx_users_email_tenant`.

## Acceptance Criteria
- Duplicate email during OAuth/headless email verification is handled as a conflict or idempotent verified state, not as a 503.
- The pending verification state does not get stuck in a retry loop.
- Add coverage for retrying email verification when the email already exists in the tenant.

dcadenas:
## Summary

- Adds a stable machine-readable `code` to duplicate-email 409 responses so clients can classify the conflict on a token instead of bare HTTP status or the human-readable message. Body becomes `{"error":"This email is already registered. Please log in instead.","code":"EMAIL_ALREADY_EXISTS"}`, following the existing `INVALID_EMAIL` precedent.
- Covers both duplicate-email paths in auth error rendering: the email-verification database constraint path and the direct `AuthError::EmailAlreadyExists` path used by duplicate registration.
- Factors coded auth error JSON through a small helper to avoid drift between duplicate-email and invalid-email responses.
- Updates OpenAPI to document the optional stable `code` field and the register `409` duplicate-email response.

## Motivation

The mobile duplicate-email recovery had no machine-readable signal to key off; it relied on bare 409 status. A stable `EMAIL_ALREADY_EXISTS` token lets clients classify the failure robustly without string-matching the human message.

The 503-to-409 behavior already shipped in #208; this completes #198 by giving the conflict a classifiable signal while keeping the existing status and message stable.

## Related Issue

Closes #198

## Testing

- [x] `cargo test -p keycast_api test_duplicate_email_response_has_stable_code`
- [x] `cargo test -p keycast_api test_register_rejects_malformed_email_with_stable_error`
- [x] `test_verify_email_duplicate_email_returns_conflict` was added for the integration path and passed in CI before the follow-up amendment
- [ ] Full `cargo test --workspace --verbose` not run locally after the amendment; CI runs the full suite

## Visuals

- [x] No visual change

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -57,6 +57,9 @@ components:
         error:
           type: string
           description: Error message
+        code:
+          type: string
+          description: Stable application error code, when available
       required:
         - error
 
@@ -342,7 +345,13 @@ paths:
               schema:
                 $ref: '#/components/schemas/RegisterResponse'
         '400':
-          description: Invalid request (email already exists, weak password, invalid nsec)
+          description: Invalid request (weak password, invalid nsec)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '409':
+          description: Email already exists
           content:
             application/json:
               schema:
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -1553,7 +1553,7 @@ pub async fn verify_email(
             let now = Utc::now();
             let mut tx = pool.begin().await?;
 
-            sqlx::query(
+            if let Err(err) = sqlx::query(
                 "INSERT INTO users (pubkey, tenant_id, email, password_hash, email_verified, email_verification_token, created_at, updated_at)
                  VALUES ($1, $2, $3, $4, $5, $6, $7, $8)",
             )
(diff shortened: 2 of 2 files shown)


NotThatKindOfDrLiz:
Follow-up pushed in 80eb085 after final review.

What changed:
- Factored the coded auth JSON response into `coded_error_response(...)` so duplicate-email and invalid-email responses do not drift.
- Added direct unit coverage for `AuthError::EmailAlreadyExists`, which is the duplicate registration path.
- Updated OpenAPI to document the optional `code` field and the register `409` duplicate-email response.
- Updated the PR body to reflect the amended scope and keep `Closes #198`.

Local verification:
- `cargo test -p keycast_api test_duplicate_email_response_has_stable_code`
- `cargo test -p keycast_api test_register_rejects_malformed_email_with_stable_error`

GitHub checks are green. I am approving separately; Daniel, please confirm you are good with the follow-up commit and merge when ready.

NotThatKindOfDrLiz (approved):
Approved. The implementation is additive, keeps the existing 409 status and human-readable message stable, and now has direct coverage for the duplicate-email error arm plus OpenAPI documentation for the response contract.

@dcadenas please confirm you agree with the follow-up commit (`80eb085`) and merge when you are ready.

PR #237, opened 2026-05-19
fix(cluster-hashring): heartbeat/full-sync never reconnect after Redis endpoint change

### LINKED ISSUE #238: fix(redis): reconnect command connections after socket loss
Redis command connections can get stuck after Redis closes an old socket.

When this happens, later cluster heartbeat, full sync, polling, or admin Redis commands may keep using the broken connection until the process restarts.

The fix should make command paths reconnect after socket loss while keeping the existing Pub/Sub reconnect behavior intact.

dcadenas:
## Summary

Redis command connections could get stuck after Redis closed an old socket.
This change makes cluster membership and API Redis helper commands reconnect, so later heartbeats, full syncs, polling writes, and admin Redis calls can recover without restarting the process.

## Motivation

The root cause was that command paths reused long-lived multiplexed Redis connections.
When Redis closed one of those sockets, later commands kept using the broken connection.

- Switched registry and prefixed API Redis commands to `redis::aio::ConnectionManager`.
- Removed the unused multiplexed factory path so new command callers use the reconnecting path.
- Left Pub/Sub on its existing reconnect and resubscribe loop.
- Kept IAM token refresh by recreating the connection manager when needed.

## Related Issue

- Closes #238

## Testing

I tested the broken-socket case directly and ran the Rust workspace checks.
The regression tests kill the active Redis client connection, confirm the first command can see the closed socket, then confirm the next command reconnects and succeeds.

- [x] Manual verification completed: `TEST_REDIS_URL=redis://localhost:16379 cargo test -p cluster-hashring registry::tests::test_registry_recovers_after_connection_killed -- --ignored --exact` and `TEST_REDIS_URL=redis://localhost:16379 cargo test -p keycast_api redis::tests::test_prefixed_redis_recovers_after_connection_killed -- --ignored --exact`

## Visuals

- [x] No visual change

## Risks

The command that first discovers a killed socket can still fail once.
That matches the existing retry/backoff behavior, and the following command should use the reconnected manager.

- `git fetch origin main` could not run here because SSH auth was unavailable.
- The branch rebased cleanly onto the local `origin/main` ref.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -58,7 +58,7 @@ crossbeam-channel = "0.5"
 num_cpus = "1.16"
 secrecy = { workspace = true }
 zeroize = { workspace = true }
-redis = { version = "0.27", features = ["tokio-comp", "aio"] }
+redis = { version = "0.27", features = ["tokio-comp", "connection-manager", "aio"] }
 cluster-hashring = { path = "../cluster-hashring" }
 
 [dev-dependencies]
--- a/api/src/redis.rs
+++ b/api/src/redis.rs
@@ -2,7 +2,7 @@
 // ABOUTME: Enables multi-app GCP Memorystore deployments with isolated namespaces
 
 use cluster_hashring::ValkeyConnectionFactory;
-use redis::aio::MultiplexedConnection;
+use redis::aio::ConnectionManager;
 use redis::{AsyncCommands, RedisResult};
 use std::borrow::Cow;
 use std::sync::Arc;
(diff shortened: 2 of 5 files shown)


NotThatKindOfDrLiz:
Pushed follow-up cleanup in commit bdbc033.

What changed:
- clarified RedisRegistry::connection() to describe the command-connection/PUBLISH usage
- documented the ConnectionManager IAM caveat: socket reconnect reuses the client credentials, so token rotation still relies on rebuilding the manager through ValkeyConnectionFactory

Validation I ran on the updated branch:
- cargo fmt --all -- --check
- cargo test -p cluster-hashring --lib
- cargo test -p keycast_api redis::tests --lib

cargo test --workspace --verbose still hits the existing local Postgres setup failure in atproto_http_test (PoolTimedOut in api/tests/common/mod.rs), so I’m using the GitHub test check as the merge gate for the full suite on this push.

NotThatKindOfDrLiz (approved):
Final pass looks good. The follow-up commit only tightens the Redis reconnect/IAM documentation and fixes the stale registry connection comment. GitHub checks are green on bdbc033, and I do not see any remaining blockers.

PR #243, opened 2026-05-24
ci(actions): use standard GitHub Actions runners

### LINKED ISSUE #271: Reduce GitHub Actions larger runner cost
## Summary

Reduce GitHub Actions runner cost by replacing `ubuntu-24.04-16core` larger runner labels with a lower-cost schedulable Ubuntu runner where appropriate.

## Motivation

The org has used most of its monthly Actions budget. Moving off the 16-core larger runner should materially reduce spend while keeping CI on a runner label that can actually schedule in this repository.

## Acceptance Criteria

- Workflow files no longer reference `ubuntu-24.04-16core`.
- Affected workflow files use a lower-cost schedulable Ubuntu runner label.
- The PR remains CI-safe and does not change application code.

rabble:
## Summary
- replace `ubuntu-24.04-16core` GitHub larger runner labels with standard `ubuntu-24.04` hosted runners
- reduce Actions spend while avoiding the unschedulable `ubuntu-24.04-4core` custom runner label

## Motivation
The Divine org has used 90% of its May 2026 Actions budget. The standard Ubuntu runner removes the 16-core larger-runner cost while keeping CI on a schedulable GitHub-hosted label.

## Linked issue
Closes #271

## Manual validation
- Verified changed workflow files no longer contain `ubuntu-24.04-16core`.
- Verified changed workflow files no longer contain `ubuntu-24.04-4core`.
- Verified changed workflow files now use `ubuntu-24.04`.

### DIFF
--- a/.github/workflows/build-test-push.yaml
+++ b/.github/workflows/build-test-push.yaml
@@ -30,7 +30,7 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04-4core
+    runs-on: ubuntu-24.04
     services:
       postgres:
         image: postgres:16
@@ -97,7 +97,7 @@ jobs:
   docker:
     name: Build & Push
     needs: test
-    runs-on: ubuntu-24.04-4core
+    runs-on: ubuntu-24.04
     if: github.event_name != 'pull_request'
     permissions:
       id-token: write
@@ -203,7 +203,7 @@ jobs:
   deploy:
     name: Trigger Deploy
     needs: docker
-    runs-on: ubuntu-24.04-4core
+    runs-on: ubuntu-24.04
     if: github.event_name != 'pull_request'
     environment: staging
 
--- a/.github/workflows/build-test-push.yaml
+++ b/.github/workflows/build-test-push.yaml
@@ -30,7 +30,7 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
     services:
       postgres:
         image: postgres:16
@@ -97,7 +97,7 @@ jobs:
   docker:
     name: Build & Push
     needs: test
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
     if: github.event_name != 'pull_request'
     permissions:
       id-token: write
@@ -203,7 +203,7 @@ jobs:
   deploy:
     name: Trigger Deploy
     needs: docker
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
     if: github.event_name != 'pull_request'
     environment: staging
 
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -16,7 +16,7 @@ env:
 
 jobs:
   e2e:
-    runs-on: ubuntu-24.04-4core
+    runs-on: ubuntu-24.04
     timeout-minutes: 45
 
     services:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -16,7 +16,7 @@ env:
 
 jobs:
   e2e:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
     timeout-minutes: 45
 
     services:


.github/workflows/build-test-push.yaml
```diff
@@ -30,7 +30,7 @@ env:
 
 jobs:
   test:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
```
realmeylisdev: Blocker: this label doesn't resolve to a runner. On this branch the `test` job waited 24h and was auto-cancelled with no runner ever assigned (`runner_name` empty, `runner_id` 0, 0 steps) — GitHub's queue timeout for a `runs-on` that matches no available runner. `ubuntu-24.04-4core` is used nowhere else in the org (the other 13 repos use `-16core`, which schedules fine in the same window). Please confirm a 4-core larger runner is provisioned in the org runner group and get a green run here before merge.

.github/workflows/build-test-push.yaml
```diff
@@ -97,7 +97,7 @@ jobs:
   docker:
     name: Build & Push
     needs: test
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
```
realmeylisdev: `docker` runs on push to main and `v*` tags. With the label unprovisioned this job hangs ~24h then cancels on every main build — and there's no required status check to block the merge that introduces it. Same prerequisite as `test`: confirm the 4-core runner exists first.

.github/workflows/build-test-push.yaml
```diff
@@ -211,7 +211,7 @@ jobs:
   deploy:
     name: Trigger Deploy
     needs: docker
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
```
realmeylisdev: Same as `docker` — `deploy` runs on push to main / tags and would hang on the unprovisioned label, stalling the Cloud Run deploy dispatch. No `timeout-minutes` here, so it waits the full 24h queue window.

.github/workflows/e2e.yaml
```diff
@@ -16,7 +16,7 @@ env:
 
 jobs:
   e2e:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04-4core
```
realmeylisdev: Once the runner exists: this job has `timeout-minutes: 45` and does a full `cargo build` + Playwright browser install. On 4 cores with a cold `rust-cache` that can get close to the cap — worth bumping to ~60. (Memory is fine: 4-core larger runner = 16 GB, well above the Vite build's 2 GB heap cap.)

realmeylisdev (changes requested):
**Requesting changes.** The YAML swap is clean — no `16core` left, all four jobs now on `ubuntu-24.04-4core`, and this touches CI only (no app / NIP / Docker-artifact / divine-mobile impact). The problem is the new runner doesn't actually work yet:

- This PR's `test` job sat *waiting for a runner* for 24h and was auto-cancelled — no runner was ever assigned (`runner_name` empty, 0 steps run). That's GitHub's queue timeout for a `runs-on` label that matches no available runner. It's the only Build/Test run on this branch and it has never passed.
- `ubuntu-24.04-4core` is used nowhere else in the org — the 13 other repos on larger runners all use `-16core`, and 16-core jobs schedule fine in the same window. So this is the label, not an outage. These `-Ncore` labels are custom org-provisioned larger runners, not GitHub defaults, so a 4-core one has to be created before it resolves.
- `main` has no required status check, so the red ❌ won't stop a merge — this could land and then hang `docker` + `deploy` on every push to main / `v*` tag for ~24h before cancelling.

**To unblock:** provision (or confirm) the `ubuntu-24.04-4core` larger runner in the org runner group, then re-run and get this PR green before merge. If a 4-core runner can't be added, fall back to the default `ubuntu-24.04` / `ubuntu-latest`, or an 8-core label that already exists.

_Minor:_ the body says both "Linked issue: None" and "Closes #271" — pick one. Memory is fine on 4-core (16 GB vs the Dockerfile's 2 GB Vite heap cap); the only perf watch-item is e2e's 45-min timeout on a cold cache.

.github/workflows/build-test-push.yaml
dcadenas: Change this label to one provisioned for `divinevideo/keycast` before merging. The same `ubuntu-24.04-4core` value is used by `test`, `docker`, and `deploy` at lines 33, 100, and 214, and by `e2e` at `.github/workflows/e2e.yaml:19`; none can start when the label has no matching runner. Confirm with a CI run that the runner ID is non-zero and steps execute; otherwise revert to `ubuntu-24.04-16core` or use a confirmed smaller label.

dcadenas (changes requested):
The new runner label is not schedulable for this repository, so every CI job can wait 24 hours and be cancelled without running. Update all four `runs-on` entries to a provisioned label and verify that a real run reaches a runner before merging.

.github/workflows/build-test-push.yaml
NotThatKindOfDrLiz: This label does not currently resolve to a runner for this repo. The PR test job using `ubuntu-24.04-4core` waited about 24h, was cancelled, had `runner_id: 0`, empty `runner_name`, and ran zero steps. Please provision/confirm the 4-core runner and rerun to green, or use a schedulable runner label.

.github/workflows/build-test-push.yaml
NotThatKindOfDrLiz: Same blocker here: if this lands with an unschedulable label, the Docker build job will queue until cancellation on main/tag runs instead of producing an image.

.github/workflows/build-test-push.yaml
```diff
@@ -211,7 +211,7 @@ jobs:
   deploy:
```
NotThatKindOfDrLiz, line 203: Same blocker for deploy. Because `main` is not branch-protected, this could merge with red checks and then leave deploy-trigger runs waiting for an unavailable runner.

.github/workflows/e2e.yaml
NotThatKindOfDrLiz: Please update this to a runner label that is actually available, or provision `ubuntu-24.04-4core` and verify a real e2e run reaches a runner before merge.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The runner label change is mechanically clean, but `ubuntu-24.04-4core` is not schedulable for this repository right now. The PR run at https://github.com/divinevideo/keycast/actions/runs/26376196221/job/82441835362 waited about 24h, was cancelled, had `runner_id: 0`, an empty `runner_name`, and ran zero steps. Please either provision/confirm this larger runner and rerun the PR to green, or switch these jobs to a label that already resolves before merging. Also please clean up the PR body contradiction between `Linked issue: None` and `Closes #271`.

NotThatKindOfDrLiz:
Takeover update:

- Replaced the unschedulable `ubuntu-24.04-4core` labels with standard `ubuntu-24.04` in all four workflow jobs.
- Updated the PR body to remove the `Linked issue: None` / `Closes #271` contradiction.
- Updated #271 acceptance criteria to reflect the schedulable lower-cost runner fallback.

Verification:
- `git diff --check`
- GitHub Actions `test` passed on `ubuntu-24.04` with an assigned GitHub-hosted runner: https://github.com/divinevideo/keycast/actions/runs/29786495068/job/88499058207
- Semantic PR and CLA checks passed.
- PR-only deploy jobs skipped as expected.

NotThatKindOfDrLiz (approved):
Approved after takeover. The blocking runner-label issue is resolved by using the schedulable standard `ubuntu-24.04` runner, the PR/issue text now matches the implementation, and the new `test` check passed on an assigned runner.

.github/workflows/e2e.yaml
```diff
@@ -16,7 +16,7 @@ env:
 
 jobs:
   e2e:
-    runs-on: ubuntu-24.04-16core
+    runs-on: ubuntu-24.04
```
realmeylisdev, line 19: Same mechanical fix as the other three jobs, but this workflow only triggers on the nightly `schedule` or `workflow_dispatch`, not `pull_request` — so this label swap hasn't actually been exercised by any CI run on this branch. Worth a manual `workflow_dispatch` run right after merge to confirm it schedules and finishes inside the 45-minute timeout. Non-blocking.

Separate pre-existing issue, unrelated to this PR: the nightly `e2e` run has been failing on `main` every day for the last week+ on a Postgres `role "root" does not exist` error (e.g. run 29723570435). That's already broken on the current `ubuntu-24.04-16core` label, not something this change causes or is responsible for fixing.

realmeylisdev (approved):
Approving. The blocker from the earlier round of reviews (`ubuntu-24.04-4core` never resolving to a runner, jobs queuing 24h then getting auto-cancelled) is fixed. All four jobs now use the standard `ubuntu-24.04` label, and the diff is exactly that — 4 lines, nothing else bundled in (confirmed against current main; everything else in the branch history is unrelated rebase noise from main).

Verified directly, not just read:
- keycast is a public repo, so standard `ubuntu-24.04` runners get GitHub's 4-core/16GB spec for public repos — the same range the never-provisioned `-4core` label was aiming for, just on the always-schedulable default pool instead of a custom one. This run's `test` job shows `runner_group_name: "GitHub Actions"`, confirming it landed on that pool.
- `test` ran green on this exact commit in 3m23s with a real runner assigned (run 29786495068), so the fix isn't just theoretical.
- `docker` and `deploy` correctly no-op on the PR event (`if: github.event_name != 'pull_request'`); their first real exercise of the new label is on merge to main, which is inherent to how this workflow is already gated, not a gap this PR could close any other way.
- Checked the other divine-* repos (mobile, web, connect, router, relay-manager, rsky, osprey): standard `ubuntu-24.04` / `ubuntu-latest` is already the dominant convention org-wide. This brings keycast in line with the rest of the org rather than being a one-off.
- The PR body's earlier "Linked issue: None" vs "Closes #271" contradiction is gone; it cleanly closes #271 now, and #271's acceptance criteria were updated to match.

One non-blocking note left inline on `e2e.yaml`: that workflow doesn't trigger on `pull_request`, so its label swap never actually ran in CI on this branch — worth a manual confirmation after merge.

dcadenas (approved):
✅ smaller CI runners, ship it

NotThatKindOfDrLiz:
@rabble I re-reviewed this and agree the previous runner-label blocker is resolved on the current head. Could you please take one more look and merge if it looks ready to you?

PR #244, opened 2026-05-27
feat: approved minor onboarding via claim link

mbradley:
## Summary

- Adds `verified_minor` boolean + timestamp columns to users table
- New `POST /api/admin/create-minor-account` endpoint (service-token auth): generates keypair, creates user with `verified_minor=true`, claims username on divine-name-server (best-effort), generates 14-day claim token
- Extends `GET/PUT /api/admin/users/:pubkey/status` to return `verified_minor` and `verified_minor_at` fields
- Idempotent retry: re-calling with the same username returns 200 with existing/refreshed claim token
- Race condition safe: concurrent creates for the same username produce correct idempotent or conflict responses (not a confusing 400)

**API contract note:** `UserStatusResponse` (GET/PUT `/api/admin/users/:pubkey/status`) now includes `verified_minor: bool` and `verified_minor_at: Option<String>`. These are additive fields — existing JSON consumers that ignore unknown fields are unaffected.

Relates to divinevideo/support-trust-safety#148
Part of divinevideo/support-trust-safety#138

Companion PR in divine-relay-manager (divinevideo/divine-relay-manager#90) depends on this being deployed first.

## Test plan

- [ ] Migration runs cleanly on staging (`verified_minor` column added, existing users default to `false`)
- [ ] `POST /api/admin/create-minor-account` with valid username returns `{ pubkey, claim_url, expires_at }`
- [ ] Duplicate username returns 409
- [ ] Invalid username format (special chars, leading hyphen) returns 400
- [ ] `GET /api/admin/users/:pubkey/status` includes `verified_minor: true` for created account
- [ ] Claim link resolves to the claim page and allows email/password setup
- [ ] divine-name-server shows NIP-05 for the created username

## Integration tests (9 tests)

1. Happy path — 201 with claim URL
2. Idempotent retry — same username returns 200 with valid claim token
3. Conflict — username taken by non-minor user (409)
4. Conflict — username taken by claimed minor with email set (409)
5. Conflict — username taken by Vine preloaded user, unclaimed non-minor (409)
6. Invalid username rejected (empty, spaces, leading hyphen → 400)
7. Missing auth token (401)
8. Wrong auth token (401)
9. Retry with expired token creates new token (200, different claim URL)

## Local verification

- [x] Migration runs cleanly (`verified_minor` boolean + `verified_minor_at` timestamptz added to users table)
- [x] Existing users default to `verified_minor=false` (confirmed via `GET /status` on a standard registered user)
- [x] `POST /api/admin/create-minor-account` with valid username returns `{ pubkey, claim_url, expires_at }` with 14-day TTL
- [x] Duplicate username returns 409 with descriptive error
- [x] Invalid username format (leading hyphen) returns 400 with validation message
- [x] Missing/wrong auth token returns 401
- [x] `display_name` optional param works correctly
- [x] `GET /api/admin/users/:pubkey/status` returns `verified_minor: true` and `verified_minor_at` for minor accounts
- [x] `GET /api/admin/users/:pubkey/status` returns `verified_minor: false` (no `verified_minor_at`) for normal accounts
- [x] DB records confirm `verified_minor=true` and `verified_minor_at` set on created users
- [x] `cargo check` and `cargo fmt` pass
- [x] Race condition fix verified: `RepositoryError::Duplicate` caught and retried idempotently


api/src/api/http/admin.rs
```diff
@@ -1921,10 +1926,135 @@ pub async fn set_user_status_admin(
...
+    let keys = Keys::generate();
+    let pubkey_hex = keys.public_key().to_hex();
+
+    let secret_bytes = keys.secret_key().to_secret_bytes();
+    let encrypted_secret = key_manager
+        .encrypt(&secret_bytes)
+        .await
+        .map_err(|e| ApiError::Internal(format!("Failed to encrypt secret: {}", e)))?;
+
+    user_repo
```
dcadenas: Make this path idempotent for pending approved-minor accounts.
Right now the endpoint creates the user and key before it creates the claim token, but a retry for the same username always hits the duplicate-username conflict path.
If token creation fails, or if the service caller loses the response, the caller cannot recover the claim link through this API.
On duplicate username, return or create a valid claim token only when the existing account is an unclaimed approved-minor account.
Keep normal users, claimed accounts, and unrelated duplicates as conflicts.

dcadenas (changes requested):
Make the minor account creation endpoint safe to retry before this merges. The current flow can reserve the username but leave the caller without a usable claim link.

dcadenas (approved):
✅ smooth claim-link onboarding path

PR #247, opened 2026-05-30
feat: RSA / ActivityPub HTTP-Signature signing (key custody + signing oracle)

### LINKED ISSUE #273: Add ActivityPub RSA key custody and signing oracle
## Summary

Add per-user RSA-2048 ActivityPub key custody and an RSA-SHA256 HTTP-Signature signing oracle to Keycast for the ActivityPub gateway.

## Motivation

The gateway needs Keycast to generate, store, expose, and use ActivityPub actor RSA keys without letting private keys leave Keycast. This must live alongside the existing Nostr signing path without changing Nostr behavior.

## Acceptance Criteria

- Keycast can idempotently generate an RSA-2048 keypair per `(tenant_id, user_pubkey)`.
- Private keys are encrypted at rest using the existing `KeyManager`.
- Public keys are exposed as SPKI PEM for ActivityPub actor documents.
- A service-token or UCAN-authenticated signing endpoint signs caller-provided HTTP Signature strings with RSA-SHA256.
- Tenant isolation and suspended-account checks are preserved.
- OpenAPI and tests cover the new endpoints and signing behavior.

rabble:
## Summary

Adds per-user **RSA-2048 key custody** and an **RSA-SHA256 HTTP-Signature signing oracle** to Keycast for ActivityPub HTTP Signatures, alongside the existing Nostr/secp256k1/Schnorr path.

Per AP actor, Keycast now:
1. Generates or returns an RSA-2048 keypair (`POST /api/ap/keys`, idempotent).
2. Stores the private key encrypted-at-rest through the existing `KeyManager`.
3. Exposes the public SPKI PEM for ActivityPub actor documents (`GET /api/ap/keys/{actor_or_pubkey}`).
4. Signs caller-supplied draft-cavage signing strings with RSASSA-PKCS1-v1_5 + SHA-256 (`POST /api/ap/sign`).

Takeover update: AP RSA key material is now wired into account lifecycle, current gateway actor-shaped requests are supported, and AP UCAN signing requires a live OAuth authorization.

Design spec and implementation plan: `docs/superpowers/specs/2026-05-30-ap-rsa-http-signature-signing-design.md`, `docs/superpowers/plans/2026-05-30-ap-rsa-http-signature-signing.md`.

## Changes

- Adds the AP RSA crypto core, tenant-scoped `ap_actor_keys` repository, HTTP routes, and OpenAPI docs.
- Adds a new hardening migration that removes existing orphan AP key rows, drops the redundant AP key lookup index, and adds cascade FK custody from AP keys to users.
- Deletes AP key material during account deletion and preserves the published AP PEM across Nostr key rotation.
- Makes the AP active-account gate fail closed for missing users and applies it consistently to create, get, and sign.
- Accepts either hex Nostr `pubkey` or ActivityPub `actor` URL, including the current gateway `{ actor, signing_string }` request shape.
- Lazily mints AP keys on public-key lookup/sign so the gateway does not need a separate preflight create call.
- Requires UCAN callers to have a live OAuth authorization bunker and rejects revoked/expired authorizations.
- Checks stored AP `key_type` before decrypting/signing.
- Carries account restrictions (`status`, `suspended_reason`, `suspended_at`, `verified_minor`, `verified_minor_at`) onto the replacement identity in `change_key_transaction`, so a suspended or banned actor cannot regain Nostr or ActivityPub signing by rotating its Nostr key.

## Test Plan

- [x] `cargo test -p keycast_core ap_signing`
- [x] `DATABASE_URL=postgres://divine:divine_dev@localhost/keycast_test_pr247_core REDIS_URL=redis://localhost:16379 TEST_REDIS_URL=redis://localhost:16379 cargo test -p keycast_core --features integration-tests --lib`
- [x] `DATABASE_URL=postgres://postgres:password@localhost/keycast_test_pr247_fk cargo test -p keycast_api --test ap_signing_integration_test --features integration-tests`
- [x] `DATABASE_URL=postgres://postgres:password@localhost/keycast_test_pr247 cargo test --workspace --verbose`
- [x] Rotation-restriction fix (rebased on `origin/main` `85d4736`, DB `keycast_test_pr247_rot`): `cargo test -p keycast_api --features integration-tests --lib --tests`, `cargo test -p keycast_core --features integration-tests --lib`, `cargo test --workspace`, plus fmt/clippy. New coverage: `key_rotation_preserves_suspension_for_ap_actor` and `test_change_key_transaction_preserves_account_restrictions` (both verified to fail on the pre-fix code).

Note: the default local `keycast_test` database had a stale migration version, so the DB-backed runs above used fresh throwaway local databases.

## Visuals

No visual changes.

## Notes / Follow-ups

- Service-token auth remains a tenant-wide gateway credential; AP actor/pubkey requests now must resolve to an existing user in the request tenant.
- `rsa` 0.9 carries RUSTSEC-2023-0071 for decryption timing, but this feature only performs signing.

## Related Issue

Closes #273

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -49,6 +49,10 @@ components:
       scheme: bearer
       bearerFormat: JWT
       description: UCAN token from /auth/register or /auth/login
+    ServiceToken:
+      type: http
+      scheme: bearer
+      description: KEYCAST_SERVICE_TOKEN for trusted service-to-service callers (e.g. the AP gateway)
 
   schemas:
     Error:
@@ -62,6 +66,75 @@ components:
           description: Stable application error code, when available
       required:
         - error
+
+    ApCreateKeyRequest:
+      type: object
+      properties:
+        pubkey:
+          type: string
+          description: Hex Nostr pubkey of the actor. Required for service-token callers; for UCAN callers it defaults to the token's pubkey and, if present, must match it.
+          example: "e1ff3bfdd4e40315959b08b4fcc8245eaa514637e1d4ec2ae166b743341be1af"
+    ApKeyResponse:
+      type: object
+      properties:
+        pubkey:
+          type: string
+        public_key_pem:
+          type: string
+          description: SPKI PEM public key for the actor's publicKey.publicKeyPem.
+          example: "-----BEGIN PUBLIC KEY-----\nMIIBIjAN...\n-----END PUBLIC KEY-----\n"
+        key_type:
+          type: string
+          example: "rsa-2048"
+        created:
+          type: boolean
+          description: true if a new key was minted; false if an existing key was returned (idempotent).
+      required:
+        - pubkey
+        - public_key_pem
+        - key_type
+        - created
+    ApPublicKeyResponse:
+      type: object
+      properties:
+        pubkey:
+          type: string
+        public_key_pem:
+          type: string
+        key_type:
+          type: string
+          example: "rsa-2048"
+      required:
+        - pubkey
+        - public_key_pem
+        - key_type
+    ApSignRequest:
+      type: object
+      properties:
+        pubkey:
+          type: string
+          description: Hex Nostr pubkey of the actor (required for service-token callers).
+        signing_string:
+          type: string
+          description: The exact draft-cavage HTTP-Signature signing string. Signed as UTF-8 bytes.
+          example: "(request-target): post /inbox\nhost: mastodon.social\ndate: Mon, 01 Jan 2026 00:00:00 GMT\ndigest: SHA-256=X..."
+      required:
+        - signing_string
+    ApSignResponse:
+      type: object
+      properties:
+        pubkey:
+          type: string
+        signature:
+          type: string
+          description: base64-encoded RSASSA-PKCS1-v1_5 SHA-256 signature for the Signature header.
+        algorithm:
+          type: string
+          example: "rsa-sha256"
+      required:
+        - pubkey
+        - signature
+        - algorithm
 
     RegisterRequest:
       type: object
@@ -456,6 +529,118 @@ paths:
                 $ref: '#/components/schemas/Error'
         '500':
           description: Signing failed (KMS error, invalid key, etc.)
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+
+  /ap/keys:
+    post:
+      summary: Create or return an actor's ActivityPub RSA key
+      description: |
+        Mints (or returns, idempotently) an RSA-2048 keypair for an actor and returns its
+        SPKI public PEM for embedding in `publicKey.publicKeyPem`. The private key is stored
+        encrypted-at-rest and is never returned. Calling again returns the existing key
+        with `created: false` — the key is never regenerated (federated keys are cached by
+        remote servers). Auth: ServiceToken (with `pubkey`) or UCAN (BearerAuth).
+      tags:
+        - ActivityPub
+      security:
+        - ServiceToken: []
+        - BearerAuth: []
+      requestBody:
+        required: false
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ApCreateKeyRequest'
+      responses:
+        '200':
+          description: Key created or already existed
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ApKeyResponse'
+        '400':
+          description: Invalid or missing pubkey
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '401':
+          description: Unauthorized
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '403':
+          description: UCAN pubkey mismatch
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  /ap/keys/{pubkey}:
+    get:
+      summary: Get an actor's ActivityPub public key PEM
+      tags:
+        - ActivityPub
+      security:
+        - ServiceToken: []
+        - BearerAuth: []
+      parameters:
+        - in: path
+          name: pubkey
+          required: true
+          schema:
+            type: string
+          description: Hex Nostr pubkey of the actor
+      responses:
+        '200':
+          description: Public key PEM
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ApPublicKeyResponse'
+        '404':
+          description: No AP key for this actor
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+  /ap/sign:
+    post:
+      summary: Sign an HTTP-Signature signing string (RSA-SHA256)
+      description: |
+        Signs the exact caller-supplied signing string with RSASSA-PKCS1-v1_5 + SHA-256 and
+        returns the base64 signature. The caller builds the draft-cavage signing string and
+        assembles the final `Signature:` header. The private key never leaves Keycast.
+        Suspended/banned accounts are rejected.
+      tags:
+        - ActivityPub
+      security:
+        - ServiceToken: []
+        - BearerAuth: []
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/ApSignRequest'
+      responses:
+        '200':
+          description: Signature
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ApSignResponse'
+        '403':
+          description: Account restricted or pubkey mismatch
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+        '404':
+          description: No AP key for this actor
           content:
             application/json:
               schema:
@@ -1061,3 +1246,5 @@ tags:
     description: User profile and session management
   - name: Discovery
     description: NIP-05 identifier discovery
+  - name: ActivityPub
+    description: RSA key custody and HTTP-Signature signing for the ActivityPub gateway
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -72,8 +72,12 @@ components:
       properties:
         pubkey:
           type: string
-          description: Hex Nostr pubkey of the actor. Required for service-token callers; for UCAN callers it defaults to the token's pubkey and, if present, must match it.
+          description: Hex Nostr pubkey of the actor. For service-token callers, either pubkey or actor is required; for UCAN callers it defaults to the token's pubkey and, if present, must match it.
           example: "e1ff3bfdd4e40315959b08b4fcc8245eaa514637e1d4ec2ae166b743341be1af"
+        actor:
+          type: string
+          description: ActivityPub actor URL. Alternative to pubkey for the gateway; resolves /ap/users/{username} against the tenant selected by the Keycast request Host.
+          example: "https://divine.video/ap/users/alice"
     ApKeyResponse:
       type: object
       properties:
@@ -113,7 +117,10 @@ components:
       properties:
         pubkey:
           type: string
-          description: Hex Nostr pubkey of the actor (required for service-token callers).
+          description: Hex Nostr pubkey of the actor. For service-token callers, either pubkey or actor is required.
+        actor:
+          type: string
+          description: ActivityPub actor URL. Alternative to pubkey for the gateway; resolves /ap/users/{username} against the tenant selected by the Keycast request Host.
         signing_string:
           type: string
           description: The exact draft-cavage HTTP-Signature signing string. Signed as UTF-8 bytes.
@@ -542,14 +549,15 @@ paths:
         SPKI public PEM for embedding in `publicKey.publicKeyPem`. The private key is stored
         encrypted-at-rest and is never returned. Calling again returns the existing key
         with `created: false` — the key is never regenerated (federated keys are cached by
-        remote servers). Auth: ServiceToken (with `pubkey`) or UCAN (BearerAuth).
+        remote servers). Auth: ServiceToken (with `pubkey` or `actor`) or an OAuth UCAN
+        for the same actor.
       tags:
         - ActivityPub
       security:
         - ServiceToken: []
         - BearerAuth: []
       requestBody:
-        required: false
+        required: true
         content:
           application/json:
             schema:
@@ -579,9 +587,12 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Error'
-  /ap/keys/{pubkey}:
+  /ap/keys/{actor_or_pubkey}:
     get:
       summary: Get an actor's ActivityPub public key PEM
+      description: |
+        Returns the actor's public PEM, lazily minting the RSA key if needed. The path
+        accepts either a hex Nostr pubkey or a URL-encoded ActivityPub actor URL.
       tags:
         - ActivityPub
       security:
@@ -589,11 +600,11 @@ paths:
         - BearerAuth: []
       parameters:
         - in: path
-          name: pubkey
+          name: actor_or_pubkey
           required: true
           schema:
             type: string
-          description: Hex Nostr pubkey of the actor
+          description: Hex Nostr pubkey or URL-encoded ActivityPub actor URL
       responses:
         '200':
           description: Public key PEM
@@ -614,7 +625,9 @@ paths:
         Signs the exact caller-supplied signing string with RSASSA-PKCS1-v1_5 + SHA-256 and
         returns the base64 signature. The caller builds the draft-cavage signing string and
         assembles the final `Signature:` header. The private key never leaves Keycast.
-        Suspended/banned accounts are rejected.
+        Suspended/banned accounts are rejected. Service-token callers identify the actor
+        with `pubkey` or `actor`; UCAN callers need a live OAuth authorization for the
+        same actor.
       tags:
         - ActivityPub
       security:
--- a/api/src/api/http/ap.rs
+++ b/api/src/api/http/ap.rs
@@ -9,11 +9,13 @@ use axum::{
     Json,
 };
 use keycast_core::ap_signing::{self, AP_KEY_TYPE};
-use keycast_core::repositories::{ApActorKeysRepository, RepositoryError, UserRepository};
+use keycast_core::repositories::{
+    ApActorKeysRepository, OAuthAuthorizationRepository, RepositoryError, UserRepository,
+};
 use nostr_sdk::PublicKey;
 use serde::{Deserialize, Serialize};
 
-use crate::api::http::auth::extract_user_from_token;
+use crate::api::http::auth::extract_user_and_origin_from_token;
 use crate::api::http::routes::AuthState;
 use crate::api::tenant::TenantExtractor;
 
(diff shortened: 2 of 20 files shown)


database/migrations/20260530000000_add_ap_actor_keys.sql
```diff
@@ -0,0 +1,19 @@
...
+CREATE TABLE public.ap_actor_keys (
+    id                     bigserial PRIMARY KEY,
+    user_pubkey            character(64) NOT NULL,
+    tenant_id              bigint NOT NULL DEFAULT 1,
+    encrypted_private_key  bytea NOT NULL,                 -- PKCS#8 DER, AES-256-GCM via KeyManager
+    public_key_pem         text  NOT NULL,                 -- SPKI PEM, public, safe at rest
+    key_type               text  NOT NULL DEFAULT 'rsa-2048',
+    created_at             timestamp with time zone NOT NULL DEFAULT now(),
+    updated_at             timestamp with time zone NOT NULL DEFAULT now(),
+    CONSTRAINT ap_actor_keys_tenant_user_unique UNIQUE (tenant_id, user_pubkey)
```
NotThatKindOfDrLiz, line 16: Blocking: this table stores encrypted private key material but is not tied to `users(pubkey)` with `ON DELETE CASCADE`, unlike `personal_keys` and `oauth_authorizations`. `UserRepository::delete_account` relies on those cascades, so AP RSA keys would survive account deletion. Please add the user FK cascade here, and preferably the tenant FK as well for consistency with the rest of the schema.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+}
+
+/// Reject signing for suspended/banned accounts (mirrors auth::sign_event).
+async fn ensure_active(
+    pool: &sqlx::PgPool,
+    tenant_id: i64,
+    user_pubkey: &str,
+) -> Result<(), ApError> {
+    let user_repo = UserRepository::new(pool.clone());
+    if let Some((status, _, _)) = user_repo
```
NotThatKindOfDrLiz: Blocking: `get_user_status()` returning `None` currently falls through as active. Combined with an orphaned `ap_actor_keys` row, `/ap/sign` can keep minting valid HTTP signatures after the user row is deleted. Please reject a missing user row before signing, and add a regression test for deleted/missing users.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+
+/// POST /api/ap/keys — create-or-return the actor's RSA key. Idempotent: never regenerates.
+pub async fn create_key(
+    tenant: TenantExtractor,
+    State(auth_state): State<AuthState>,
+    headers: HeaderMap,
+    Json(req): Json<CreateKeyRequest>,
+) -> Result<Json<KeyResponse>, ApError> {
+    let tenant_id = tenant.0.id;
+    let user_pubkey = resolve_principal(&headers, tenant_id, req.pubkey.as_deref()).await?;
```
NotThatKindOfDrLiz: This path should also confirm the actor exists before generating durable encrypted key material. On the service-token path, `resolve_principal` only validates that the supplied pubkey is syntactically valid, so a trusted caller bug could create AP keys for non-existent users. Please make the actor existence/active check explicit for key creation as well as signing.

database/migrations/20260530000000_add_ap_actor_keys.sql
```diff
@@ -0,0 +1,19 @@
...
+);
+
+CREATE INDEX idx_ap_actor_keys_tenant_user ON public.ap_actor_keys (tenant_id, user_pubkey);
```
NotThatKindOfDrLiz, line 19: This index duplicates the implicit btree index created by the unique constraint on `(tenant_id, user_pubkey)`. Please drop the explicit index to avoid extra storage and write amplification.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+        };
+        (status, Json(serde_json::json!({ "error": message }))).into_response()
+    }
+}
+
+// ---- auth: resolve the acting user_pubkey from service token OR UCAN ----
+
+/// Constant-time check of the bearer against KEYCAST_SERVICE_TOKEN.
+/// Mirrors `admin::authorize_service_token` (duplicated to keep that path untouched).
+fn is_service_token(headers: &HeaderMap) -> bool {
```
NotThatKindOfDrLiz, line 56: Please avoid duplicating this security-sensitive service-token comparison. `admin.rs` already has the same blake3 + constant-time comparison logic; extract a shared helper so future hardening cannot drift between admin and AP routes. The AP route can still wrap the helper with its bool/fallback-to-UCAN behavior.

api/openapi.yaml
```diff
@@ -452,6 +525,118 @@ paths:
...
+        encrypted-at-rest and is never returned. Calling again returns the existing key
+        with `created: false` — the key is never regenerated (federated keys are cached by
+        remote servers). Auth: ServiceToken (with `pubkey`) or UCAN (BearerAuth).
+      tags:
+        - ActivityPub
+      security:
+        - ServiceToken: []
+        - BearerAuth: []
+      requestBody:
+        required: false
```
NotThatKindOfDrLiz: Spec/implementation drift: the handler uses `Json<CreateKeyRequest>`, so callers must send an `application/json` body (`{}` for UCAN callers). Please mark this request body required or change the extractor to actually accept an absent body.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The crypto path looks careful, but the new AP RSA key custody needs to be wired into the same account-lifecycle and schema hygiene patterns as the existing key material before this is mergeable.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+) -> Result<String, ApError> {
+    if is_service_token(headers) {
+        let pk = body_pubkey.ok_or_else(|| {
+            ApError::BadRequest("pubkey is required for service-token calls".into())
+        })?;
+        return validate_hex_pubkey(pk);
+    }
+
+    // UCAN path
+    let token_pubkey = extract_user_from_token(headers, tenant_id)
```
dcadenas: Decide and enforce the authorization model for UCAN-based AP signing before allowing the custodial key to sign. The UCAN branch in `api/src/api/http/ap.rs:94-107` checks only the token pubkey, while the Nostr signing path requires a live per-origin authorization and honors revocation. Require that same live authorization for the token origin, or document next to `resolve_principal` why pubkey possession is intentionally sufficient for this capability.

api/openapi.yaml
dcadenas: Make the OpenAPI contract agree with the handler. `api/openapi.yaml:543` marks the create-key body optional and advertises a bodyless UCAN path, but `create_key` uses the mandatory `Json<CreateKeyRequest>` extractor at `api/src/api/http/ap.rs:170-177`, so a conforming bodyless request is rejected before the handler runs. Either mark the request body required or accept an absent body in the extractor, and add a test for the chosen contract.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct SignResponse {
+    pub pubkey: String,
+    pub signature: String,
+    pub algorithm: &'static str,
+}
+
+// ---- handlers ----
+
+/// POST /api/ap/keys — create-or-return the actor's RSA key. Idempotent: never regenerates.
+pub async fn create_key(
```
dcadenas: Call `ensure_active` in `create_key` before generating and storing a key. The signing path already checks status at `api/src/api/http/ap.rs:278`, but this create path lets a suspended or banned account publish a public key that other ActivityPub servers may cache. Add coverage that a suspended account is rejected while the documented gateway flow for an actor without a users row still succeeds.

dcadenas (dismissed):
Before this can merge, close the account-lifecycle and authorization gaps in the new ActivityPub signing flow and make its public contract match the handler. The create path also needs to respect suspended accounts so key provisioning and signing follow the same account controls.

---

- **AP RSA keys are not tied into the account-deletion lifecycle, so a deleted actor keeps a live signing oracle** ([core/src/repositories/user.rs:1543](https://github.com/divinevideo/keycast/blob/10b577c7c216cc7f74d6a409f0886f3ad1964bbf/core/src/repositories/user.rs#L1543))
  Delete the tenant-scoped `ap_actor_keys` row inside `delete_account` before deleting the user. The migration at `database/migrations/20260530000000_add_ap_actor_keys.sql:16` has no foreign key or cascade, so the current flow leaves the encrypted RSA key behind; after deletion, `ensure_active` in `api/src/api/http/ap.rs:111-127` allows the missing user and the signing endpoint can still produce a valid signature. Add a regression test that verifies the row is removed and a later sign request returns 404.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+pub struct PublicKeyResponse {
+    pub pubkey: String,
+    pub public_key_pem: String,
+    pub key_type: String,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct SignRequest {
+    /// Hex Nostr pubkey of the actor. Required for service-token callers.
+    pub pubkey: Option<String>,
```
realmeylisdev, line 231: **Blocking (cross-repo contract).** The only consumer, `divine-activity-pub`, POSTs `{ actor, signing_string }` and passes the **actor URL** (`https://<domain>/ap/users/<username>`) as the identifier — `src/keycast.mjs:55`, `src/index.mjs:361`, `src/delivery.mjs:66`. This struct reads `pubkey` (the unknown `actor` field is dropped) and validates it as 64-hex via `PublicKey::from_hex`. On the gateway's service-token path a missing `pubkey` → 400 “pubkey is required”; the actor-URL value → 400 “Invalid hex pubkey”. So `SIGNER_MODE=keycast` fails every sign/get before any crypto runs. The handler follows its own design doc (the gateway is supposed to resolve actor→hex pubkey), but no `divine-activity-pub` branch does that, and the gateway never calls `POST /api/ap/keys` to provision. Fix needs to land on both sides together — or this endpoint could accept the actor URL and lazily mint.

api/src/api/http/ap.rs
```diff
@@ -0,0 +1,305 @@
...
+    Json(req): Json<SignRequest>,
+) -> Result<Json<SignResponse>, ApError> {
+    let tenant_id = tenant.0.id;
+    let user_pubkey = resolve_principal(&headers, tenant_id, req.pubkey.as_deref()).await?;
+    let pool = &auth_state.state.db;
+
+    ensure_active(pool, tenant_id, &user_pubkey).await?;
+
+    let repo = ApActorKeysRepository::new(pool.clone());
+    let row = repo
```
realmeylisdev: `row.key_type` is read but never checked before the bytes are parsed as PKCS#8 RSA and signed as `rsa-sha256`. Fine today (only `create` writes the table, always `rsa-2048`), but the moment a second `key_type` exists this silently mis-signs while still returning `algorithm: "rsa-sha256"`. Assert `row.key_type == AP_KEY_TYPE` before decrypt/sign.

api/tests/ap_signing_integration_test.rs
```diff
@@ -0,0 +1,338 @@
...
+            "/ap/keys",
+            serde_json::json!({ "pubkey": pubkey }),
+            Some("not-the-service-token"),
+        ))
+        .await
+        .unwrap();
+    assert_eq!(resp.status(), StatusCode::UNAUTHORIZED);
+}
+
+// NOTE: Test D (UCAN-for-A requesting pubkey B -> 403) is intentionally NOT
```
realmeylisdev: Coverage gap on the controls that matter here: (a) suspended/banned `/ap/sign` → 403, (b) deleted-user sign (today returns 200 with a valid signature, not 404 — see the cascade issue), and (c) a request in the consumer's real wire shape (`{ actor, … }`), which would surface the field/identifier mismatch. Together with the UCAN-mismatch 403 you already note as untested, each of these would catch a real regression or a current defect CI is green on.

realmeylisdev (changes requested):
## Formal review — RSA / ActivityPub HTTP-Signature signing

Checked the crypto, the endpoints, the DB/lifecycle wiring, and the real consumer (`divine-activity-pub`). **Crypto is solid** — an independent `openssl dgst -sha256 -verify` accepts a Keycast-produced (PEM, signature) pair (SPKI PEM + PKCS#1 v1.5 + SHA-256 + standard base64 all correct). Requesting changes on the lifecycle wiring and the gateway contract.

### Blocking

**1. Doesn't sign end-to-end with the actual gateway.** `divine-activity-pub` (the only consumer) sends `{ actor: <actor-URL> }`; this endpoint reads `{ pubkey: <hex> }`. On the gateway's service-token path every `/api/ap/sign` and `GET /api/ap/keys/{actor}` returns 400 (missing `pubkey` / invalid hex). The gateway also never calls `POST /api/ap/keys`, so keys are never provisioned (404). The endpoint matches its own design doc (gateway resolves actor→hex pubkey first), but no gateway branch does that yet. Needs a coordinated `divine-activity-pub` change — or accept the actor URL + lazy-mint here. Fails closed; not active in prod (`SIGNER_MODE` ≠ `keycast`).

**2. Deletion and rotation leave the RSA key behind.** `ap_actor_keys` has no FK/cascade and appears nowhere in `user.rs`:
- Account deletion orphans the encrypted key; since `ensure_active` treats a missing user as active, `/ap/sign` keeps returning valid signatures (200, not 404) for the deleted actor. *(Confirms @NotThatKindOfDrLiz + @dcadenas.)*
- Key rotation ([`change_key_transaction`](https://github.com/divinevideo/keycast/blob/10b577c7c216cc7f74d6a409f0886f3ad1964bbf/core/src/repositories/user.rs#L996)) orphans the old key and gives the new pubkey no row, so the next sign **mints a new key** — silently rotating the actor's published `publicKeyPem`, which the design explicitly forbids (remote servers cache it). Handle deletion **and** rotation, plus the FK cascade.

**3. UCAN path ignores revocation.** `resolve_principal` authorizes on pubkey possession only: a revoked OAuth app's UCAN — rejected on the Nostr path — can still sign here, and bare session UCANs are accepted. Enforce the same live authorization or document why possession suffices. *(Confirms @dcadenas.)*

### Worth fixing
- `create_key` / `get_key` skip the `ensure_active` gate `sign` enforces (inconsistent).
- `key_type` is stored but never checked before signing — latent footgun *(inline)*.
- OpenAPI `POST /ap/keys` body marked optional; handler requires it *(existing comment)*.
- Redundant index on `(tenant_id, user_pubkey)` *(existing)*. The duplicated service-token helper is a DRY nit — the design doc explicitly allowed duplicating it, so not a spec violation.
- Missing tests: suspended-sign 403, deleted-user sign, UCAN-mismatch 403, and one using the consumer's real `{ actor, … }` shape (would have caught #1) *(inline)*.

### Note
The single global `KEYCAST_SERVICE_TOKEN` now grants cross-tenant, any-actor signing (full ActivityPub identity forgery) — a larger blast radius than the admin routes it reuses. Worth scoping per tenant.

NotThatKindOfDrLiz:
Takeover pushed in 6f29117.

What changed:
- tied AP RSA key custody into account lifecycle with explicit delete + cascade FK migration;
- preserved the published AP PEM across Nostr key rotation;
- made AP active-account checks fail closed and apply to create/get/sign;
- supported the gateway actor-shaped contract and lazy AP key creation on get/sign;
- required AP UCAN signing to have a live OAuth authorization and rejected revoked/expired/mismatched UCANs;
- added key_type validation, OpenAPI corrections, and regression coverage for the review blockers.

Verification:
- cargo fmt --all -- --check
- cargo test -p keycast_core ap_signing
- cargo test -p keycast_core --features integration-tests --lib
- cargo test -p keycast_api --test ap_signing_integration_test --features integration-tests
- cargo test --workspace --verbose
- GitHub Actions: test, semantic PR, and CLA are green on 6f29117

NotThatKindOfDrLiz (approved):
Re-reviewed after takeover push 6f29117. The previously blocking lifecycle, rotation, UCAN revocation, and gateway contract issues are resolved, with regression coverage and green CI.

core/src/repositories/user.rs
```diff
@@ -1421,6 +1421,20 @@ impl UserRepository {
         .execute(&mut *tx)
         .await?;
 
+        // Preserve the actor's published ActivityPub RSA public key across
+        // Nostr key rotation; remote servers cache publicKeyPem.
+        sqlx::query(
+            "UPDATE ap_actor_keys
```
realmeylisdev, line 1612: This moves the `ap_actor_keys` row to `new_pubkey`, but `username` isn't part of the transfer — the old row keeps it (only `email`/`password_hash` are cleared above, at line 1400), and the new row's `INSERT` above (line 1411) never sets one. `status` isn't touched either, so the old row stays active.

The real gateway (`divine-activity-pub/src/keycast.mjs`) never calls `/api/ap/keys` or `/api/ap/sign` with a raw pubkey — it always resolves actor URL → username → pubkey via `find_pubkey_by_username`. After rotation, that lookup still returns `old_pubkey`, whose AP key row was just moved away. `ensure_active(old_pubkey)` still passes, so `find_or_create_actor_key` lazily mints a brand-new RSA key under the old identity — a different `publicKeyPem` than the one cached by remote servers, and different from the one now sitting unreachable under `new_pubkey`. That's the exact "rotation silently changes the published key" defect this PR set out to fix, still live for the only request shape production actually sends.

`key_rotation_preserves_ap_public_pem_on_new_pubkey` doesn't catch this because it only asserts through direct `ApActorKeysRepository` lookups by pubkey, not through the actor/username HTTP path the gateway actually uses.

realmeylisdev (changes requested):
Re-reviewed after 6f29117. Two of the three blocking items check out — traced them against the real `divine-activity-pub` client source, not just the design doc:

- Gateway contract (actor vs pubkey shape): fixed. `actor_username`/`resolve_actor_pubkey` parse exactly the actor-URL shape `divine-activity-pub` sends, and the lazy-mint on GET/sign means the gateway never needing to call `POST /api/ap/keys` is no longer fatal.
- UCAN revocation: fixed. Bare session UCANs (no `bunker_pubkey`) are rejected, and `exists_active_for_bunker` now checks `revoked_at` and `expires_at`, tenant-scoped.

The third one doesn't actually hold up for the only request shape production sends. Left an inline comment on `change_key_transaction` — key rotation moves the `ap_actor_keys` row by pubkey, but the username the gateway actually resolves actors by never moves, so post-rotation actor lookups still land on the old pubkey and lazily mint a brand-new, never-published key there. That's the exact "rotation silently changes the published key" bug this PR set out to fix, still reachable.

Non-blocking, for awareness:
- `GET /api/ap/keys/{actor_or_pubkey}` now mints a key as a side effect of what was a pure-read GET. Low risk given the auth gate, but worth a line in the PR description.
- The `key_type` mismatch guard before signing returns 500 instead of a 4xx. Dead code today (only one key type exists), but will confuse callers if a second type is ever added.
- The already-accepted global `KEYCAST_SERVICE_TOKEN` risk is slightly wider now: a leaked token only needs a target's public AP username, not their internal hex pubkey, to sign as them. Already documented as an accepted tradeoff in the OpenAPI description, just flagging that this PR is what widened it.

core/src/repositories/user.rs
```diff
@@ -1421,6 +1421,20 @@ impl UserRepository {
         .execute(&mut *tx)
         .await?;
 
+        // Preserve the actor's published ActivityPub RSA public key across
+        // Nostr key rotation; remote servers cache publicKeyPem.
+        sqlx::query(
```
dcadenas, line 1611: Keep actor URL resolution and RSA key custody anchored to the same pubkey after rotation. In `change_key_transaction`, the `ap_actor_keys` update at `core/src/repositories/user.rs:1426-1436` moves the row to `new_pubkey`, but the username remains on the old row because the update at `user.rs:1399-1407` clears only email and password, while the new row at `user.rs:1410-1422` has no username. The gateway still resolves actor-shaped key and signing requests to the old pubkey, so `find_or_create_actor_key` at `api/src/api/http/ap.rs:288-302` mints a fresh RSA key instead of returning the preserved PEM. Remove the move or migrate the username atomically, and add a regression that exercises GET `/api/ap/keys/{actor}` and POST `/api/ap/sign` after rotation, asserting the original PEM is returned with `created == false`.

dcadenas (dismissed):
Address the post-rotation ActivityPub key custody path before merging. The current rotation logic can strand the preserved RSA key and cause actor-based key and signing requests to mint a different key, so the PR does not yet preserve federated signature identity across Nostr key changes.

NotThatKindOfDrLiz (approved):
The remaining rotation blocker is fixed. Actor URL resolution now follows the transferred username after key rotation, the AP key remains on the new pubkey, and the regression covers GET/sign through the actor path with the original PEM. Local verification and GitHub checks are green.

NotThatKindOfDrLiz:
Takeover follow-up pushed in 88a2c60.

What changed:
- moved the username during Nostr key rotation so actor URL resolution follows the preserved AP RSA key row;
- expanded the AP rotation regression to create/lookup/sign through the gateway actor URL shape after rotation and verify against the original public PEM;
- changed unsupported stored AP key types from a 500 to a 400 response.

Verification:
- cargo fmt --all -- --check
- cargo test -p keycast_core ap_signing
- DATABASE_URL=postgres://divine:divine_dev@localhost/keycast_test_pr247_takeover cargo test -p keycast_api --test ap_signing_integration_test --features integration-tests
- DATABASE_URL=postgres://divine:divine_dev@localhost/keycast_test_pr247_takeover cargo test --workspace --verbose
- GitHub checks are green on 88a2c60: semantic PR, CLA, and test.

@rabble please review the takeover changes and merge when you are ready.

core/src/repositories/user.rs
```diff
@@ -1406,18 +1413,34 @@ impl UserRepository {
         .execute(&mut *tx)
         .await?;
 
-        // Create new user identity with email/password
+        // Create new user identity with email/password/username. The AP gateway
+        // resolves actor URLs by username, so it must move with the AP key row.
         sqlx::query(
-            "INSERT INTO users (pubkey, tenant_id, email, password_hash, email_verified, created_at, updated_at)
-             VALUES ($1, $2, $3, $4, $5, $6, $7)",
+            "INSERT INTO users (pubkey, tenant_id, email, password_hash, email_verified, username, created_at, updated_at)
```
dcadenas: Carry `status`, `suspended_reason`, and `suspended_at` from the old user into this replacement insert. The username and ActivityPub key move to the new row here, but the omitted fields default to an active account, so a suspended or banned actor can pass `ensure_active` and sign again after rotation. Add a regression that keeps actor-shaped ActivityPub GET/sign requests forbidden after rotating a restricted account.

dcadenas (changes requested):
Preserve account restrictions when transferring the ActivityPub identity during key rotation. The replacement user currently defaults to active, so a suspended or banned actor can regain ActivityPub signing after rotation. Carry the lifecycle fields into the replacement insert and add regression coverage before merge.

realmeylisdev:
Fixed in `249de0f` — rotation now carries the account restrictions onto the replacement identity.

**What was wrong:** `change_key_transaction` inserted the replacement `users` row with only email/password/username, so `status` fell back to the column default `active`. A suspended or banned actor could rotate its Nostr key and land on an unrestricted row — `ensure_active` on `/api/ap/keys` and `/api/ap/sign` then passed and handed the ActivityPub signing oracle back. The post-rotation session UCAN was minted from the new row too, so it also came out `active`.

**Fix:** the pre-orphan read now pulls `status`, `suspended_reason`, `suspended_at`, `verified_minor` and `verified_minor_at` off the old row (typed `RotatedIdentityRow`, matching the named-`FromRow` convention in that file) and the insert re-applies them, so restrictions travel with the identity exactly like the username and the AP key already do. `verified_minor` is inert today because `refuse_key_egress_for_verified_minor` blocks minors from `change_key` entirely — it is carried so the designation cannot be dropped if that gate ever moves.

**Regression coverage:**
- `api/tests/ap_signing_integration_test.rs::key_rotation_preserves_suspension_for_ap_actor` — suspend an actor that already owns an AP key, rotate, assert the new pubkey is still `Suspended` (with reason and `suspended_at` carried) and that both `GET /ap/keys/{actor}` and `POST /ap/sign` return 403 through the actor-URL path. Verified it fails on the pre-fix code with `left: Active, right: Suspended`.
- `api/tests/change_key_test.rs::test_change_key_transaction_preserves_account_restrictions` — repository-level: a banned + verified-minor identity keeps all five columns on the replacement row.

**Verification** (rebased on `origin/main` `85d4736`, throwaway local DB):
- `cargo test -p keycast_api --features integration-tests --lib --tests` (includes both suites above)
- `cargo test -p keycast_core --features integration-tests --lib`
- `cargo test --workspace`

core/src/repositories/user.rs
realmeylisdev: Done in `249de0f`. The pre-orphan read now returns a typed `RotatedIdentityRow` (`username`, `status`, `suspended_reason`, `suspended_at`, `verified_minor`, `verified_minor_at`) and the replacement insert re-applies all of it, so restrictions move with the identity like the username and AP key do. `verified_minor` is carried too — inert today since `refuse_key_egress_for_verified_minor` blocks minors from `change_key`, but it means the designation can't be silently dropped if that gate moves.

Regression added as asked: `key_rotation_preserves_suspension_for_ap_actor` suspends an actor that already owns an AP key, rotates, then asserts the new pubkey is still `Suspended` and that actor-shaped `GET /ap/keys/{actor}` and `POST /ap/sign` both return 403. It fails on the pre-fix code (`left: Active, right: Suspended`). `test_change_key_transaction_preserves_account_restrictions` covers the repository level for a banned + verified-minor row.

dcadenas (approved):
🔐 RSA key custody and signing, ship it

realmeylisdev (approved):
Approving — every blocking item I raised is resolved at `249de0f`, re-verified against the current head rather than taken from the fix commit messages.

**My earlier blockers, now closed:**
- **Actor-URL contract with `divine-activity-pub`** — `CreateKeyRequest`/`SignRequest` accept `actor`, `resolve_actor_pubkey` maps the actor URL through `find_pubkey_by_username`, and `resolve_requested_pubkey` rejects a `pubkey`/`actor` pair that disagrees. `gateway_actor_shape_lazily_creates_key_and_signs` exercises the wire shape the gateway actually sends.
- **`key_type` unchecked before RSA parse** — gated at `api/src/api/http/ap.rs:398` before decrypt/sign.
- **Username stranded on the old row after rotation** — `change_key_transaction` now NULLs `username` on the orphaned row and carries it onto the replacement, so actor-URL resolution and the moved `ap_actor_keys` row stay anchored to the same pubkey.
- **Coverage gap** — suspended, deleted-user, revoked-UCAN, pubkey-mismatch, and post-rotation actor-path cases are all covered in `api/tests/ap_signing_integration_test.rs`.

**Others' blockers, verified closed:**
- FK `ON DELETE CASCADE` on `(user_pubkey, tenant_id)` plus the tenant FK, and the redundant `idx_ap_actor_keys_tenant_user` dropped — `database/migrations/20260720000000_harden_ap_actor_keys_lifecycle.sql`, with a pre-flight delete of already-orphaned rows so the constraint can be added.
- `ensure_active` returns `NotFound` when `get_user_status` yields `None`, so a missing user row no longer reads as active, and it runs on create as well as sign (`ap.rs:323,350,381`).
- UCAN callers now need a live OAuth authorization (`exists_active_for_bunker`), not just pubkey possession.
- OpenAPI create-key body marked `required: true`, matching the `Json<CreateKeyRequest>` extractor.

**One non-blocking item left open** (thread intentionally unresolved): the service-token comparison at `ap.rs:56` still duplicates `admin::authorize_service_token` (`admin.rs:2011`) rather than sharing a helper. It is the same blake3 + `ct_eq` logic, so it is correct today; the risk is drift if one side is hardened later. Fine to land as-is and extract in a follow-up.

PR #251, opened 2026-06-16
fix(deploy): persist teams UI flag

dcadenas:
## Summary

Teams UI disappeared on login.divine.video because the Cloud Run deploy config did not include the teams feature flag.
A manual Cloud Run env update restored it immediately, but the next Cloud Build deploy would remove it again.
This change adds the same flag to the durable Cloud Build deploy step so future Cloud Run revisions keep teams visible.

## What Changed

The deploy config now sets the runtime flag that the frontend reads before showing Teams links and dashboard sections.
This keeps the Cloud Run environment aligned with the manual production update that is already serving traffic.

- Added `SHOW_TEAMS_FUNCTIONALITY=true` to the `gcloud run deploy --set-env-vars` list in `cloudbuild.yaml`.
- Left secrets and unrelated deploy settings unchanged.

## Testing

I verified the manual production update first, then verified this config change locally and through the repo pre-push hook.
The first push attempt exposed that local Postgres was not running for database-backed tests; after starting the repo Postgres dependency and creating `keycast_test`, the normal hook passed.

- Live Cloud Run revision `keycast-00246-g5q` is serving 100% traffic with `SHOW_TEAMS_FUNCTIONALITY` present.
- Live `https://login.divine.video/` injects `"SHOW_TEAMS_FUNCTIONALITY":"true"` in `window.__ENV__`.
- `ruby -e 'require "yaml"; YAML.load_file("cloudbuild.yaml"); puts "cloudbuild.yaml parses"'`
- Pre-push hook passed: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`, `cargo test --workspace`, `cargo build --release --workspace`.

## Risks

Risk is low because this only adds one clear-text feature flag to the existing Cloud Run env var list.
The main risk is environment divergence: GKE keycast configs do not set this flag today.

- Asked `divine-iac-coreconfig` via Ouija before opening this PR.
- Reply: `SHOW_TEAMS_FUNCTIONALITY` was not found in GKE IaC, and no equivalent teams flag is currently set there.
- Prod keycast currently runs on Cloud Run, so this PR fixes prod Cloud Run only.

## Follow-ups

If teams UI should also be enabled for GKE keycast environments, add `SHOW_TEAMS_FUNCTIONALITY=true` in `divine-iac-coreconfig` separately.


mbradley (approved):
### Code review

No issues found. Verified the flag is wired end to end: `SHOW_TEAMS_FUNCTIONALITY` is read in [`keycast/src/main.rs`](https://github.com/divinevideo/keycast/blob/ed67aee9a9988084131c91a9aea53c6dbbf38c53/keycast/src/main.rs#L158-L161) and consumed by [`web/src/lib/utils/env.ts`](https://github.com/divinevideo/keycast/blob/ed67aee9a9988084131c91a9aea53c6dbbf38c53/web/src/lib/utils/env.ts#L47-L50), where `isTeamsEnabled()` matches `'true'` exactly. `--set-env-vars` replaces the full env set on each deploy, so baking the flag into `cloudbuild.yaml` is the correct way to persist it. Consistent with the env-var additions in #148 and #152.

PR #256, opened 2026-06-22
feat(rpc): nip17_unwrap_batch — server-side NIP-59 gift-wrap unwrap (server verb for #254)

realmeylisdev:
## Description

Adds a server-side gift-wrap unwrap verb, `nip17_unwrap_batch`, to `POST /api/nostr`.
It accepts an ordered list of kind:1059 gift wraps and returns ordered, index-aligned
per-item slots — `{rumor, sender}` on success or `{error}` on a per-item failure.

This is the keycast-side implementation of #254 (Approach B). NIP-17 DM history drain
currently costs **two sequential `nip44_decrypt` RPCs per message** (gift wrap → seal,
seal → rumor). By unwrapping both layers server-side, this collapses the drain to **one
call per page** and the per-request uncacheable user-status `SELECT` from **N → 1**. It
compounds the client-side parallelism already shipped in `divinevideo/divine-mobile#5426`.

### How it works

`HttpRpcHandler::nip17_unwrap`:
- delegates the unwrap to nostr-sdk's vetted `nip59::UnwrappedGift::from_gift_wrap`, which
  decrypts both layers, **verifies the seal's signature** (the sole cryptographic anchor of
  sender identity), and rejects a rumor whose author disagrees with the seal signer;
- additionally verifies the **outer** wrap's id + signature, for parity with the mobile
  client's defense-in-depth check (`from_gift_wrap` verifies only the seal);
- runs the crypto on `spawn_blocking` (same pattern as `sign_event`);
- enforces the cached decrypt policy on the **authenticated inner sender** — `from_gift_wrap`
  decrypts internally, so without this an `encrypt_to_self`-restricted authorization could read
  others' DMs.

The dispatch path bounds batch size (`MAX_UNWRAP_BATCH = 100`, well under axum's 2 MB default
body limit), bounds fan-out concurrency (`UNWRAP_BATCH_CONCURRENCY = 16`, pure-tokio — no new
dependency), preserves request order, returns per-item error slots so one malformed wrap never
fails the batch, gates suspended users via the existing per-request status check, and coalesces
the activity log to one write per request. The token's auth / DPoP enforcement happens once per
request, as for the single methods.

### Wire shape

```jsonc
// request
{ "method": "nip17_unwrap_batch",
  "params": [ <giftwrap_1059_event>, <giftwrap_1059_event>, ... ] }

// response (ordered, index-aligned with params)
{ "result": [
    { "rumor": { /* kind:14 unsigned event */ }, "sender": "<hex authenticated author>" },
    { "error": "sender_mismatch" }
] }
```

Per-item error codes: `invalid_event`, `not_gift_wrap`, `invalid_wrap_signature`,
`decrypt_failed`, `sender_mismatch`, `permission_denied`, `internal`.

## Behavioral note (reviewers please weigh in)

Sender-mismatch is handled **strictly**: when `rumor.pubkey != seal.pubkey`, the item returns a
`sender_mismatch` error slot. The current mobile client (`GiftWrapUtil.getRumorEvent`) is
**lenient** — it re-attributes the message to the authenticated seal pubkey and keeps it. For
remote-signer accounts on a new server, the rare mismatched-pubkey messages would therefore
become error slots (falling into the client's existing retry path) rather than being shown.
Strict is arguably more correct (NIP-59 requires the match), but this is a deliberate, reversible
choice — happy to add a metric and/or restore the lenient behavior if preferred.

## Verification

- New handler unit tests (run under default `cargo test`): round-trip, non-gift-wrap rejected,
  wrong-recipient rejected, expired-auth denied, and `encrypt_to_self` policy denies a DM from
  another sender.
- New integration tests (`--features integration-tests`, via `invoke_nostr_rpc` against a local
  Postgres): happy path + partial-failure ordering with authenticated sender, size-cap rejection,
  and suspended-user gate.
- Full `keycast_api` lib suite: 137 passed, 0 failed.

## Out of scope (follow-ups)

- **Client adoption** (`divine-mobile`): swap the remote-signer DM drain
  (`_parallelDecryptGiftWraps`, two RPCs/message) for chunked `nip17_unwrap_batch` calls with a
  method-not-found fallback. Tracked in a divine-mobile issue.
- **Benchmark**: extend the #255 harness to the unwrap verb to measure the end-to-end win.

## Related

Refs #254. Compounds `divinevideo/divine-mobile#5426`.

### DIFF
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -1050,4 +1050,62 @@ mod tests {
         let result = enforce_cached_dpop_binding(&handler, &headers, htu).await;
         assert!(result.is_ok());
     }
+
+    /// Build a kind:1059 gift wrap (from `sender` to `receiver`) carrying a kind:14
+    /// DM rumor with the given text, serialized to its JSON event value.
+    async fn gift_wrap_value(sender: &Keys, receiver: &PublicKey, text: &str) -> JsonValue {
+        let rumor =
+            nostr_sdk::EventBuilder::private_msg_rumor(*receiver, text).build(sender.public_key());
+        let wrap = nostr_sdk::EventBuilder::gift_wrap(
+            sender,
+            receiver,
+            rumor,
+            Vec::<nostr_sdk::Tag>::new(),
+        )
+        .await
+        .expect("gift wrap should build");
+        serde_json::to_value(&wrap).expect("serialize gift wrap")
+    }
+
+    /// `unwrap_gift_wrap_batch` chunks at `UNWRAP_BATCH_CONCURRENCY` (16). The
+    /// integration ordering test uses 4 items, so the multi-chunk path never runs.
+    /// This drives 40 items (> 2 chunks) through the function directly and asserts
+    /// every slot's content + authenticated sender stays index-aligned with the
+    /// request across chunk boundaries.
+    #[tokio::test]
+    async fn unwrap_gift_wrap_batch_preserves_order_across_chunks() {
+        let handler = Arc::new(create_test_handler_with_dpop(None));
+        let receiver = handler.public_key();
+
+        // 40 = 2 * UNWRAP_BATCH_CONCURRENCY + 8 -> spans three chunks (16, 16, 8).
+        const N: usize = 40;
+        assert!(N > UNWRAP_BATCH_CONCURRENCY, "must exceed one chunk");
+        let mut senders = Vec::with_capacity(N);
+        let mut items = Vec::with_capacity(N);
+        for i in 0..N {
+            let sender = Keys::generate();
+            items.push(gift_wrap_value(&sender, &receiver, &format!("ordered message {i}")).await);
+            senders.push(sender);
+        }
+
+        let results = unwrap_gift_wrap_batch(&handler, &items).await;
+        assert_eq!(results.len(), N, "one ordered slot per input");
+
+        for (i, slot) in results.iter().enumerate() {
+            assert!(
+                slot.get("error").is_none(),
+                "slot {i} should succeed, got {slot:?}"
+            );
+            assert_eq!(
+                slot["rumor"]["content"].as_str().unwrap(),
+                format!("ordered message {i}"),
+                "content out of order at slot {i}"
+            );
+            assert_eq!(
+                slot["sender"].as_str().unwrap(),
+                senders[i].public_key().to_hex(),
+                "authenticated sender out of order at slot {i}"
+            );
+        }
+    }
 }
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -1079,7 +1079,6 @@ mod tests {
 
         // 40 = 2 * UNWRAP_BATCH_CONCURRENCY + 8 -> spans three chunks (16, 16, 8).
         const N: usize = 40;
-        assert!(N > UNWRAP_BATCH_CONCURRENCY, "must exceed one chunk");
         let mut senders = Vec::with_capacity(N);
         let mut items = Vec::with_capacity(N);
         for i in 0..N {
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -777,7 +777,10 @@ fn unwrap_error_slot(code: &str) -> JsonValue {
 /// never fails the batch. Items run with bounded concurrency
 /// (`UNWRAP_BATCH_CONCURRENCY`); `chunks` + ordered await preserves request
 /// order (NIP-17 history is positional).
-async fn unwrap_gift_wrap_batch(handler: &Arc<HttpRpcHandler>, items: &[JsonValue]) -> Vec<JsonValue> {
+async fn unwrap_gift_wrap_batch(
+    handler: &Arc<HttpRpcHandler>,
+    items: &[JsonValue],
+) -> Vec<JsonValue> {
     let mut results: Vec<JsonValue> = Vec::with_capacity(items.len());
 
     for chunk in items.chunks(UNWRAP_BATCH_CONCURRENCY) {
@@ -804,7 +807,11 @@ async fn unwrap_gift_wrap_batch(handler: &Arc<HttpRpcHandler>, items: &[JsonValu
         }
         // Await in spawn order to keep the response index-aligned with the request.
         for handle in handles {
-            results.push(handle.await.unwrap_or_else(|_| unwrap_error_slot("internal")));
+            results.push(
+                handle
+                    .await
+                    .unwrap_or_else(|_| unwrap_error_slot("internal")),
+            );
         }
     }
 
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -30,10 +30,6 @@ use super::routes::AuthState;
 /// limit (gift wraps are ~1-2 KB each). Mirrors the `batch_lookup_users` cap.
 const MAX_UNWRAP_BATCH: usize = 100;
 
-/// Bounded fan-out width for unwrapping a batch. Caps in-flight blocking-pool
-/// tasks so one large batch can't starve the runtime.
-const UNWRAP_BATCH_CONCURRENCY: usize = 16;
-
 /// RPC request format (mirrors NIP-46)
 #[derive(Debug, Deserialize)]
 pub struct NostrRpcRequest {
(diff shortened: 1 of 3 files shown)


mbradley (commented):
Thanks for this. I ran a couple of review passes, and the security core holds up well: the `sender` returned is the authenticated seal signer rather than any attacker-controllable rumor field, the decrypt policy is enforced on that authenticated sender on every item before any rumor is returned (which closes the encrypt_to_self bypass), the outer-wrap verification adds correctly-scoped integrity defense-in-depth, suspended users are gated, and the block_on-inside-spawn_blocking pattern matches `sign_event` exactly. The flag on the strict sender-mismatch behavior was noted and helpful.

A few things I'd like addressed before this lands. Contributions for the items below are in the works and will be committed to this PR:

1. `sender_mismatch` has no test, and since it's the strict-vs-lenient behavior you flagged, it's worth one pinning it. It's straightforward to construct: nostr-sdk's own suite does it in about ten lines (`test_sender_mismatch` in nip59.rs) by building the rumor with `.build(impersonated_pubkey)` so the rumor author differs from the sealing key, then wrapping and asserting the error.

2. `invalid_wrap_signature`, the outer-wrap defense-in-depth check, is also untested. It's harder to construct, but it's the only thing behind a branch that's load-bearing by design, since `from_gift_wrap` doesn't verify the outer wrap. If it turns out too awkward to forge with EventBuilder, a short comment explaining why it isn't covered will stand in.

3. The ordering test uses four items, below `UNWRAP_BATCH_CONCURRENCY` (16), so the multi-chunk path never actually runs. Ordering looks correct by inspection, but a >16-item case asserting ordered content will lock it in.

4. Minor: the `nip59::Error::NotGiftWrap` arm in the error map is unreachable, since kind is gated before `spawn_blocking`. Harmless, but a one-line "defensive" comment saves the next reader a double-take.

This next one is a question rather than a request, so I'm not proposing a change for it: `spawn_log_activity` is called once per batch unconditionally and bumps `activity_count` by one regardless of how many items succeeded, including all-error batches, whereas the single-method paths only log after success. Depending on what `activity_count` and `last_activity` feed into (audit, idle revocation, stats), an all-error batch would keep an authorization looking active, and a 100-message page would under-count relative to per-message decrypts. I'm not sure whether that matters given how those fields are actually used, so I'm flagging it to confirm intent. Once we know which way it should go, I'm happy to fold in any adjustment.

I've also filed a follow-up issue (#257) for one thing broader than this PR: the fan-out is capped at 16 per request, but there's no global bound on concurrent blocking crypto, so a burst of large batches across requests could saturate tokio's default 512-thread blocking pool that's shared with all signing. It queues rather than deadlocks, so it isn't a blocker here, and it's tracked there.

Deferring to @dcadenas for final approval.

realmeylisdev:
Thanks for the thorough pass. All four addressed in c0fdcda:

1. Added `nip17_unwrap_rejects_sender_mismatch`, built like nostr-sdk's `test_sender_mismatch` — the rumor is `.build(impersonated_pubkey)` while the seal is signed by `sender`, so `from_gift_wrap` authenticates `sender` but the inner author disagrees → `sender_mismatch`.

2. `invalid_wrap_signature` turned out forgeable with `EventBuilder` after all, so no comment fallback needed. `Event` deserialization doesn't verify (it's documented "This method doesn't verify the signature!"), so taking a valid wrap, tampering its `content`, and round-tripping through JSON yields a kind:1059 event whose recomputed id no longer matches — the outer `gift_wrap.verify()` then fails before any decrypt runs. Added `nip17_unwrap_rejects_invalid_wrap_signature`, which also asserts the tampered event stays kind:1059 so it genuinely reaches the outer-verify branch.

3. Added `unwrap_gift_wrap_batch_preserves_order_across_chunks` — 40 items spanning three `UNWRAP_BATCH_CONCURRENCY` chunks (16/16/8), each with index-distinct content and a distinct sender, asserting per-slot content + authenticated sender stay index-aligned across chunk boundaries. Made it a unit test directly on `unwrap_gift_wrap_batch` so it runs in the default `cargo test` suite rather than only under `--features integration-tests`.

4. Added the defensive comment on the unreachable `nip59::Error::NotGiftWrap` arm.

On the `activity_count` / `last_activity` question: keeping the single per-batch log as-is. Those fields are display-only — they feed the sessions dashboard (per-session "last active" + count, and aggregated totals) and nothing revokes, expires, or audits off them anywhere in the tree (the only two writers are `nostr_rpc.rs` and `signer_daemon.rs`, both `SET last_activity = NOW(), activity_count = activity_count + 1`). So an all-error batch nudging `last_activity`, and a 100-message page counting as one, are both harmless. If we later tie revocation or audit to activity, we'll revisit (log-on-success, or increment by success count).

Re #257 (global bound on concurrent blocking crypto): agreed it's out of scope here; thanks for filing it.

mbradley:
c0fdcda covers the three review items: the sender_mismatch and invalid_wrap_signature guards, and the multi-chunk ordering test (a unit test driving unwrap_gift_wrap_batch directly, 40 items across three chunks). I pulled the branch and the full lib suite passes locally (141 passed), including the three new tests.

Two threads from the review remain open, neither blocking the test work:
- the activity-log question: whether logging once per batch and bumping activity_count by one, including on all-error batches, is the intended semantics, given it diverges from the per-success single-method paths.
- the broader blocking-pool bound, tracked in #257.

Deferring to @dcadenas for final approval.

api/src/api/http/nostr_rpc.rs
```diff
@@ -723,6 +765,59 @@ fn parse_decrypt_params(params: &[JsonValue]) -> Result<(PublicKey, String), Rpc
...
+    items: &[JsonValue],
+) -> Vec<JsonValue> {
+    let mut results: Vec<JsonValue> = Vec::with_capacity(items.len());
+
+    for chunk in items.chunks(UNWRAP_BATCH_CONCURRENCY) {
+        let mut handles = Vec::with_capacity(chunk.len());
+        for item in chunk {
+            let handler = handler.clone();
+            let item = item.clone();
+            handles.push(tokio::spawn(async move {
```
dcadenas: Replace the detached `tokio::spawn` loop here with a fan-out that dies with the request.
Right now, if the request future is dropped (client disconnects mid-batch), the spawned tasks keep decrypting and the coalesced activity log at the end never runs.
Switch to a `JoinSet` or a stream with `buffered(UNWRAP_BATCH_CONCURRENCY)` so dropping the request cancels the in-flight unwraps.

Also bound the work across requests, not just within one.
Per-request width is capped at 16, but N concurrent batches put up to N*16 `spawn_blocking` crypto units on the shared blocking pool that also serves `sign` and `decrypt`, so a burst of batches can add latency to unrelated signing.
Gate each item's unwrap on a single module-level `tokio::sync::Semaphore` so total in-flight unwrap crypto has a ceiling.

The current loop to update:

https://github.com/divinevideo/keycast/blob/2688e8020156d90a68372802ff730ef028b1258b/api/src/api/http/nostr_rpc.rs#L786-L816

Keep the response index-aligned with the request when you rewrite it.

dcadenas (dismissed):
The new batch unwrap fans each item out with a detached spawn, and that needs to change before this can merge.
Two problems follow from it.
If the client disconnects mid-batch the spawned tasks keep decrypting and the activity log at the end is skipped, and there is no cap on unwrap crypto across requests, so many concurrent batches can crowd out unrelated signing on the shared blocking pool.
Please make the fan-out cancel with the request and bound total unwrap work with a single shared limit.

NotThatKindOfDrLiz (changes requested):
I think this still needs one adjustment before push/merge.

The `JoinSet` change correctly ties the per-request fan-out tasks to the request future, but the global semaphore permit is currently held by the async task around `handler.nip17_unwrap(event).await`. `nip17_unwrap` then dispatches the real crypto via `tokio::task::spawn_blocking`.

That means if the request is dropped after the blocking closure has started, aborting the `JoinSet` drops the async task and releases the semaphore permit, but the blocking crypto keeps running. New requests can then acquire permits and enqueue more unwrap crypto, so the intended global cap is no longer a true cap under disconnect/cancel churn.

I think the permit needs to live for the lifetime of the blocking crypto itself, e.g. be acquired as an owned permit and moved into the `spawn_blocking` closure, or otherwise held until that closure exits. The current doc comment should also be softened unless/until that is true, since `JoinSet` abort cannot preempt an already-running `spawn_blocking` closure.

Final approval of this PR should come from `dcadenas`, since this is on the live blocker he raised.

realmeylisdev:
Pushed `70cc837` addressing both blockers.

**@dcadenas** — the detached `tokio::spawn` fan-out is replaced with a `JoinSet`, so dropping the request future aborts the in-flight unwraps (the coalesced activity log no longer gets skipped on disconnect). Cross-request work is now bounded by a single process-wide `Semaphore` (`UNWRAP_CRYPTO_PERMITS`, cap 64) acquired on every item, replacing the per-request chunk-of-16. The response stays index-aligned via reassembly by position.

**@NotThatKindOfDrLiz** — the permit now lives for the lifetime of the blocking crypto, not the async wrapper: `nip17_unwrap` acquires an *owned* permit and moves it into the `spawn_blocking` closure, so it releases only when the closure returns. Since `JoinSet::abort` can't preempt an already-running `spawn_blocking`, the permit (not the cancellable async task) is what enforces the cap under disconnect/cancel churn. The doc comments now state this explicitly instead of implying abort stops running crypto.

The ordering test now drives a full 100-item page (larger than the 64-permit pool), so items queue on the semaphore and complete out of order — exercising both the queueing and the index-reassembly paths. CI is green (build + full test suite); clippy and fmt clean.

Ready for another look — deferring final approval to @dcadenas.

NotThatKindOfDrLiz (approved):
Re-reviewed the current PR tip 70cc837. My prior blocker is fixed: the unwrap crypto permit is now acquired as an owned permit and moved into the spawn_blocking closure, so cancellation cannot release global capacity while blocking crypto continues running. The JoinSet also ties per-item async fan-out to the request future, and the tests now cover sender_mismatch, invalid_wrap_signature, and full-page ordering across the 64-permit pool.

I verified locally against 70cc837 in an isolated worktree:
- cargo test -p keycast_api nip17_unwrap --lib
- cargo test -p keycast_api unwrap_gift_wrap_batch_preserves_order --lib
- cargo clippy -p keycast_api --all-targets -- -D warnings -A deprecated
- cargo fmt --all -- --check

Approving to clear my stale CHANGES_REQUESTED. @dcadenas, this still needs your final re-review/approval since 70cc837 addresses the blocker you raised around request cancellation and the shared unwrap crypto bound.

dcadenas:
I re-requested the review myself, remember this is a signal that increases visibility, I'm taking a look now

dcadenas (approved):
🎁 batched gift-wrap unwrap stays bounded by the crypto semaphore, ship it.

realmeylisdev:
@NotThatKindOfDrLiz, @dcadenas thank your for your review and feedback, merging now.

PR #258, opened 2026-06-23
feat: self-serve email change with dual confirmation

### LINKED ISSUE #223: feat: self-serve email change with dual confirmation
## Summary

Users with email/password auth should be able to change their email address through a self-serve flow. The change requires password re-verification and confirmation from both the old and new email addresses before taking effect.

## Problem or Opportunity

There is no way for a user to update their email address today. If someone loses access to their email, typos their address during registration, or simply wants to use a different address, they have no path forward without manual intervention.

### Flow

1. **Initiate** -- authenticated user submits new email + current password via `POST /api/user/change-email`
   - Re-verify password (existing `get_credentials` + `spawn_blocking(bcrypt::verify)` pattern)
   - Validate new email with existing `normalize_registration_email()`
   - Check new email is not already registered (unique index)
   - Generate two tokens via `generate_secure_token()` -- one for old address, one for new
   - Store pending state on user row (see schema below)
   - Send confirmation link to new address, security notification + confirm/cancel link to old address
   - Return 200 regardless of outcome (anti-enumeration)

2. **Confirm** -- `POST /api/auth/confirm-email-change` (unauthenticated, token-based)
   - Look up user by token, record which side confirmed
   - When both old and new addresses have confirmed: atomically update `email`, set `email_verified = true`, clear pending state
   - Log to `auth_events`

### Why dual confirmation

- Old-address confirmation proves the real account holder initiated the change (not just someone with a stolen session)
- New-address confirmation proves the new address is reachable and owned
- Separate tokens per address prevent cross-confirmation (old address holder can't confirm for new, and vice versa)

### Schema migration

Six new nullable columns on `users` (consistent with existing `password_reset_token` / `email_verification_token` pattern -- no new tables):

```sql
ALTER TABLE users ADD COLUMN pending_email TEXT;
ALTER TABLE users ADD COLUMN pending_email_old_token TEXT;
ALTER TABLE users ADD COLUMN pending_email_new_token TEXT;
ALTER TABLE users ADD COLUMN pending_email_expires_at TIMESTAMPTZ;
ALTER TABLE users ADD COLUMN pending_email_old_confirmed_at TIMESTAMPTZ;
ALTER TABLE users ADD COLUMN pending_email_new_confirmed_at TIMESTAMPTZ;

CREATE INDEX idx_users_pending_email_old_token
  ON users (pending_email_old_token) WHERE pending_email_old_token IS NOT NULL;
CREATE INDEX idx_users_pending_email_new_token
  ON users (pending_email_new_token) WHERE pending_email_new_token IS NOT NULL;
```

### Files affected

- `core/src/repositories/user.rs` -- new methods for pending email CRUD
- `api/src/api/http/auth.rs` -- two new handlers (initiate, confirm)
- `api/src/email_service.rs` -- two new methods on `EmailSender` trait (old-address notification, new-address confirmation)
- `api/src/api/http/routes.rs` -- register new routes
- New migration in `database/migrations/`

### Security considerations

- Password re-verification before initiating (session hijack can't change email without password)
- 24h token expiry (matches registration verification)
- 5-minute resend cooldown (matches existing verification resend)
- Anti-enumeration on initiation endpoint
- Reuse existing email normalization to prevent case/dot bypass
- Unique index prevents setting an already-registered email
- SendGrid click/open tracking remains disabled (tokens stay off redirect servers)
- Both addresses notified, so the real owner knows if they didn't request it

## Alternatives Considered

- **Single confirmation (new address only):** Simpler, but a stolen session + password could silently hijack an account. Dual confirmation is the industry standard (Google, GitHub, etc.).
- **Redis-based token storage:** Avoids a migration but Redis is optional in Keycast and tokens would be lost on restart. DB columns are consistent with how all other token flows work.
- **Separate `pending_email_changes` table:** Cleaner normalization but breaks the established pattern of per-row token columns. Not worth the divergence.

## Acceptance Criteria

- [ ] Authenticated user can initiate email change with password re-verification
- [ ] Old address receives security notification with confirm/cancel link
- [ ] New address receives confirmation link
- [ ] Email change only completes after both addresses confirm
- [ ] Expired tokens are rejected
- [ ] Duplicate email (already registered) is rejected
- [ ] Initiating a new change cancels any pending change
- [ ] `auth_events` records both initiation and completion
- [ ] Anti-enumeration: initiation always returns success
- [ ] Tests cover happy path, expiry, duplicate email, wrong password, and partial confirmation

mbradley:
Closes #223

## Summary

Adds a self-serve email change flow for email/password users. Initiating requires password re-verification; the change only takes effect after **both** the old and new addresses confirm. The old-address email also carries a cancel link so the real account holder can immediately kill a change they didn't request.

This implements the design in #223. The dual-confirmation + per-row token-column approach matches the existing `password_reset_token` / `email_verification_token` flows rather than introducing a new table.

## Flow

1. **Initiate** (`POST /api/user/change-email`, authenticated): re-verify password, normalize + validate the new email, reject if already registered (anti-enumeration: always returns 200), generate two tokens, store pending state, email a confirm link to the new address and a confirm/cancel notice to the old address.
2. **Confirm** (`POST /api/auth/confirm-email-change`, token-based): records the confirming side; once both sides confirm, finalizes atomically.
3. **Cancel** (`POST /api/auth/cancel-email-change`, token-based): clears pending state; generic response so it leaks nothing.

## Decisions

- **Cancel endpoint included.** A token-bound cancel is defense-in-depth: if a live session is hijacked and used to move the account, the alert lands on the still-controlled old address and the owner can abort immediately. Standard practice pairs the old-address notice with a "this wasn't me" affordance. The old-address token serves both confirm and cancel; the action is distinguished by endpoint, not by a separate token.
- **Frontend included.** Beyond the backend the issue scoped, this adds a Change Email section in `settings/security` (password-gated) plus `confirm-email-change` and `cancel-email-change` pages, so the flow works end to end.
- **One column beyond the spec's six:** `pending_email_sent_at`, to back the resend cooldown cleanly (mirrors `email_verification_sent_at`).

## Security notes

- Password re-verification before initiating (a stolen session alone can't change the email).
- 64-char tokens, 24h expiry, distinct per address.
- Anti-enumeration on initiate (always 200; duplicate target returns success without writing or sending).
- Finalize is a single atomic UPDATE guarded on both confirmed-at columns. This both prevents a TOCTOU race when the two links are clicked concurrently and handles the case where the target email gets registered between initiate and confirm (unique violation -> clear pending, return conflict).
- Failed initiations (invalid email, wrong password) are recorded as `failure` auth_events, feeding the same brute-force monitoring as login / forgot-password.
- The resend cooldown is scoped to re-initiations of the *same* target (a change to a different address is a new request and supersedes immediately); the cooldown response is honest that no fresh email was sent.
- Both addresses are notified, so the owner learns of an unrequested change.
- The new email is HTML-escaped where it's interpolated into the notification email body (defense-in-depth; callers already normalize it).
- SendGrid click/open tracking stays disabled, so tokens don't pass through redirect servers.

## Code review

Reviewed by a fresh reviewer against the requirements; no Critical issues. All Important and Minor findings are resolved:

- Resend cooldown no longer short-circuits the "new change supersedes prior" path, and its message is honest about not resending.
- Failed initiations are now audited.
- Shared old-address confirm/cancel token is documented at the call site.
- Two theoretical Minor findings (no unique constraint on token columns; stale-token cross-advance) were verified to resolve correctly and need no change.

## Verification

- `cargo build` (workspace) clean; `cargo clippy` on `keycast_api` + `keycast_core`: 0 warnings.
- 10 integration tests pass (`api/tests/email_change_test.rs`): dual-confirm finalize (both orders), partial-no-finalize, wrong password (with failure-event assertion), duplicate-email anti-enumeration, expiry, cancel, invalid token, new-change-supersedes-prior, same-target-cooldown-no-resend.
- `svelte-check` clean on the three frontend files.

## Notes for review

- This is an auth path, so it warrants manual scrutiny. The design-review question I left on #223 was never answered; this follows the spec as written.
- Implementation plan: `docs/superpowers/plans/2026-06-23-self-serve-email-change.md`.


web/src/routes/confirm-email-change/+page.svelte
```diff
@@ -0,0 +1,171 @@
...
+	const api = new KeycastApi();
+
+	const token = $derived($page.url.searchParams.get('token'));
+
+	let status = $state<'idle' | 'loading' | 'done' | 'error'>('idle');
+	let message = $state('');
+
+	// Auto-submit once on mount when a token is present.
+	let submitted = $state(false);
+	$effect(() => {
```
dcadenas: Require a human click before this page submits the token.
Right now the `$effect` fires `confirm()` as soon as a token is present, so anything that loads the link consumes the confirmation.

A mail security scanner or link-preview service that runs page JavaScript will POST the confirmation on its own.
The old-address confirmation is the main proof that the real account holder approved the change, so firing it automatically takes the human out of the loop.
If the old mailbox's gateway pre-loads the link, an attacker who already controls the new address can finish the swap.

Replace the auto-submit with a button the user clicks, which calls `confirm()`.
Keep the existing loading, done, and error states for after the click.

---
The cancel page (`web/src/routes/cancel-email-change/+page.svelte:14-17`) has the identical auto-submit-on-mount pattern. Since the old-address email carries both the confirm and the cancel link, a gateway that pre-loads links can hit the cancel endpoint and silently kill a legitimate change — so on link-scanning mail providers the flow may never complete. Apply the same fix here too: require an explicit user action instead of firing on mount.

core/src/repositories/user.rs
```diff
@@ -557,6 +587,216 @@ impl UserRepository {
...
+        &self,
+        pubkey: &str,
+        tenant_id: i64,
+        side: PendingEmailSide,
+    ) -> Result<(), RepositoryError> {
+        // Column name is a fixed internal mapping, never user input.
+        let col = match side {
+            PendingEmailSide::Old => "pending_email_old_confirmed_at",
+            PendingEmailSide::New => "pending_email_new_confirmed_at",
+        };
```
dcadenas, line 718: Bind these updates to the token you just looked up, not just pubkey and tenant.
`confirm_email_change` resolves a token to a (pubkey, side) with `find_by_pending_email_token`, then `mark_pending_email_confirmed` updates by pubkey and tenant only.
`cancel_email_change` does the same through `clear_pending_email_change`.

Because `set_pending_email_change` overwrites any prior pending change in place and rotates both tokens, a second `change_email` that lands between the read and the write makes the stale write hit the new pending change.
A stale confirm marks a side approved that never approved it, and a stale cancel wipes a fresh, legitimate change.

`finalize_email_change_if_ready` already shows the right pattern, a single guarded UPDATE:

https://github.com/divinevideo/keycast/blob/2846d1f6abc166b794b3ccbebce6d451e83fd164/core/src/repositories/user.rs#L738-L753

Add the per-side token column to the WHERE clause of `mark_pending_email_confirmed` (Old side: `AND pending_email_old_token = $token`, New side: `AND pending_email_new_token = $token`) and pass the token through from `confirm_email_change`.
Gate the cancel-path clear on the token matching one of the two token columns.
Keep `finalize_email_change_if_ready`'s EmailTaken cleanup at user.rs:765 clearing unconditionally, so add an optional token param or a separate token-gated method rather than changing that path.
Check `rows_affected` so a superseded change surfaces instead of silently succeeding.

dcadenas (changes requested):
This flow has two gaps that weaken the guarantee that both the old and new mailbox must approve, and both need fixing before merge.
The confirm and cancel pages both submit their token the moment the link loads, so a mail scanner or link preview can approve the old side — or silently cancel a legitimate change — with no human action.
The confirm and cancel database updates key only on pubkey and tenant, so a second change that lands mid-request lets a stale token confirm or wipe a fresh change.
Details are in the line comments.

web/src/routes/confirm-email-change/+page.svelte
mbradley: Fixed in e283def. Both the confirm and cancel pages no longer submit on mount. They start in an idle state and only POST the token from an explicit button click, so a link-scanning gateway or preview that runs page JS can't consume the confirmation or silently cancel a pending change. The loading, done, and error states are unchanged after the click.

core/src/repositories/user.rs
mbradley: Fixed in 6860efa. `mark_pending_email_confirmed` now binds the per-side token in its WHERE clause and the cancel path uses a new token-gated `clear_pending_email_change_by_token` (matching one of the two token columns). Both return whether a row was affected; a superseded confirm now returns 409 instead of silently succeeding (b976a54), and cancel only writes its audit event on an actual clear. `finalize_email_change_if_ready`'s EmailTaken cleanup still clears unconditionally as you asked, and feca09f documents and tests the invariant that keeps finalize safe keyed on pubkey+tenant. Repository tests cover both stale and current tokens.

dcadenas (approved):
🔒 dual-confirm email change with the rotation races sealed off, ship it.

PR #259, opened 2026-06-23
chore: Improve expired-session UX on protected pages

### LINKED ISSUE #252: Improve expired-session UX on protected pages
A support admin can currently see a raw authentication error if the page stays open after their session expires.
Make protected pages treat that as an expired login instead of a broken admin action.
The user should get a clear login prompt or redirect, and internal API auth text should not appear in the UI.

dcadenas:
## Summary

A support admin saw the raw API error `Missing authentication - expected UCAN Bearer token or keycast_session cookie` after a user lookup. The real cause was an expired login, not a broken lookup. Protected pages now treat that 401 as an expired session: the user gets a clear message and is sent to login with a return path, instead of seeing an internal error string.

## Motivation

The session cookie (`keycast_session`) lasts 24h, but the client-set `keycastUserPubkey` cookie lasts 14 days. The route gate trusted either cookie, so after the session expired the page still loaded and its cookie-authenticated API calls failed with raw 401 text that the page rendered verbatim.

## What Changed

The gate now trusts only `keycast_session`, matches by path prefix (so `/teams/123` and `/admin/registered-clients` are covered, and the dead `/keys` entry is gone), and redirects to `/login?redirect=<path>`. A shared `redirectToLoginOnAuthError` helper turns protected-page 401s into a session-expired toast plus login redirect; it is wired into every protected page's initial load (incl. the admin/support-admin/registered-clients onMount checks) and every action/mutation (support-admin, admin, registered-clients, and all teams CRUD incl. deleteTeam), captures the user's actual location for re-login, and dedupes a burst of simultaneous 401s to one toast and one navigation. It fires only on 401, so 403 (forbidden) and other errors keep their messages. The login return path and the nip07 `signin()` path run targets through a same-origin guard (`safeRedirectPath`) that resolves via `new URL`, rejects anything changing origin, and also rejects results whose normalized pathname starts with `//` — covering protocol-relative (`//host`), backslash (`/\host`), control-character (`/\t/host`), and normalization (`/%2e%2e//host`) vectors.

## Related Issue

- Closes #252

## Testing

I added Bun unit tests for the redirect guard, and ran the type-check, build, and a live SSR gate test.

- `cd web && bun test src/lib/utils/redirect.test.ts` — 9 pass (protocol-relative, backslash, control-char, and normalization vectors).
- `cd web && bun run check` — zero errors in changed files (8 pre-existing errors remain in unrelated files: QRCode.svelte, teams/[id]/keys/[pubkey]/+page.svelte).
- `cd web && bun run build` — succeeds.
- `vite dev` + curl: no cookie or only `keycastUserPubkey` -> 303 to login; nested `/teams/123` and `/admin/registered-clients` -> 303 (previously 200); `/keys` -> 404; valid `keycast_session` -> 200.
- Not run: full browser repro of the production client path (needs Postgres + Redis + the Rust binary). Cargo checks are not applicable — this PR changes no Rust.

## Risks

The deployed web app is a static SPA, so `hooks.server.ts` does not run per-request in production — it runs under `bun run dev`/SSR and at prerender time. In production the runtime protection comes from the client-side initial-load checks plus the per-action 401 handling, both of which redirect to login and never trust `keycastUserPubkey`. The gate fix still guards the dev/SSR path and any future server adapter.

## Visuals

- [x] No visual change (text-only: a session-expired toast and a redirect)
- [x] Visuals and text avoid sensitive external brand or partner names


mbradley (approved):
Reviewed this end to end and it holds up well. I verified the parts most likely to hide a silent bug:

- `ApiError` carries `.status` and the API throws it for every non-OK response, so the `instanceof ApiError && status === 401` check actually fires.
- `setCurrentUser(null)` clears both the in-memory state and the `keycastUserPubkey` cookie.
- Every catch that wraps an API call routes through `redirectToLoginOnAuthError`. The catches that don't are all clipboard writes, `nip19.npubEncode`, and input validation, so the coverage is complete.

The `safeRedirectPath` guard and its test vectors (protocol-relative, backslash, control-char, `%2e%2e//` normalization) are a nice touch.

Rather than leave comments for round-tripping, I pushed four small follow-up commits:

- drop a duplicate failure toast on create authorization/policy (both catch handlers fired two toasts for one failure)
- correct a stale comment on the claim-token load catch (it silently ignores every non-401, not just 403/404)
- add a 10s watchdog reset so the `redirecting` dedup flag can't latch forever if a navigation promise never settles and silently swallow later 401s; the `.finally` still does the normal fast reset
- document that the role-based post-login destination is intentional, so a future reader doesn't "fix" it

Two notes, no action needed:

- The nip07 sign-in (role-based console) vs password login (`/`) landing divergence is intentional, now documented in code.
- The `hooks.server.ts` gate is dev/SSR-only since production is a static SPA, as you call out. The real production protection is the client-side handling, which is comprehensive here, so the gate is fine as belt-and-suspenders.

Approving.

PR #266, opened 2026-06-30
feat(minor-safety): expose verified_minor on GET /user/account

### LINKED ISSUE #263: feat(minor-safety): expose verified_minor flag on GET /user/account
Keystone of divinevideo/support-trust-safety#173.

The durable approved-minor flag **already exists**: Keycast stores `users.verified_minor` (+ `verified_minor_at`), set at minor-account creation (`create_minor_account`), and a combined `status + verified_minor` query already backs admin status endpoints (`core/src/repositories/user.rs`). The gap is purely **exposure** — no client-facing status surface returns it:
- `GET /user/account` (`get_account_status` → `AccountStatusResponse`) returns email/status/suspended_reason and **omits `verified_minor`**. It also only sets `account_status` when *not* active, so an approved minor (active + `verified_minor`) reads as a plain active account.
- relay-manager `moderation-status` returns plain `active` too.

## Decision
Surface it via **Keycast `GET /user/account`** (chosen over routing through relay-manager `moderation-status`, which is gated on relay-manager#108). Clients read the flag directly from Keycast.

## What
- Add `verified_minor` (bool) and `verified_minor_at` (timestamp) to `AccountStatusResponse`.
- Populate from the existing combined `status + verified_minor` query already in `user.rs` (swap the handler's current `get_account_status` call for the combined one).
- Return `verified_minor` **independently of `account_status`** — an approved minor is active, so today's "None when active" logic would otherwise hide it.
- Document the response shape for the client-consume issue (#173 item 3).

## Band note
`verified_minor` is boolean; the 13-15 band lives in relay-manager `age_review_cases`. Under-13 is never approved, so `verified_minor=true` ≈ approved 13-15. The boolean is sufficient for client protections unless an explicit band is needed downstream.

## Acceptance
`GET /user/account` for an approved-minor account returns `verified_minor: true` (+ `verified_minor_at`), regardless of active status, with the shape documented for clients to consume.

Part of the protected-minor safeguards epic: divinevideo/support-trust-safety#173.

mbradley:
## What

Exposes the durable approved-minor (13-15) flag on Keycast's account-status endpoint so clients can detect the **protected-minor** state directly from Keycast. Keystone of the protected-minor safeguards epic (raised by Liz; tracked privately in the T&S repo, divinevideo/support-trust-safety#173).

Closes #263.

## Changes

- `AccountStatusResponse` (`GET /user/account`) gains two fields:
  - `verified_minor: bool` — always present
  - `verified_minor_at` (timestamp) — present only when set, omitted otherwise
- New repo method `get_account_status_with_minor` returns the existing account-status fields **plus** the minor flag in a single query. Kept separate from `get_account_status` so the shared oauth login/session paths are untouched.
- The flag is returned **independently of `account_status`**: an approved minor is `active`, so gating it behind the existing "None when active" logic would hide it.

## Response shape

`GET /user/account` for an approved-minor account:

```json
{
  "email": "user@example.com",
  "email_verified": true,
  "public_key": "<hex>",
  "verified_minor": true,
  "verified_minor_at": "2026-06-30T12:00:00Z"
}
```

A non-minor account returns `"verified_minor": false` and omits `verified_minor_at`. `account_status` / `suspended_reason` continue to appear only for non-active accounts. Field naming matches the admin `UserStatusResponse` for consistency.

## Tests

- Repo-level integration tests (`api/tests/account_status_minor_test.rs`): an approved minor surfaces the flag **while `active`**; a regular user reports `false` with no timestamp.
- Run locally against Postgres 16 (matching CI): **both pass**, and the full `keycast_core` (113) and `keycast_api` (376 across 38 binaries) integration suites are green with this change — no regressions from the added response fields or repo method.
- A handler-level HTTP test is deferred: it requires minting a UCAN user token. The data-exposure core is covered at the repo layer, and the handler change is straightforward wiring.

## Notes

- Additive, read-only change to an existing response. No migration — the `verified_minor` / `verified_minor_at` columns already exist.
- Unblocks the client-side protected-minor work (consume the flag, lock adult content, restrict DMs).

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -374,6 +374,13 @@ pub struct AccountStatusResponse {
     pub account_status: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suspended_reason: Option<String>,
+    /// True for an approved minor (13-15) account (Keycast `users.verified_minor`).
+    /// Returned independently of `account_status`: an approved minor is `active`, so gating
+    /// this behind the non-active check (like `account_status`) would hide the protected-minor
+    /// state from clients. Always present so a `false` reliably signals "not a protected minor".
+    pub verified_minor: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub verified_minor_at: Option<chrono::DateTime<chrono::Utc>>,
 }
 
 #[derive(Debug, Deserialize, Serialize)]
@@ -2345,11 +2352,18 @@ pub async fn get_account_status(
 
     let user_repo = UserRepository::new(pool.clone());
     let user = user_repo
-        .get_account_status(&user_pubkey, tenant_id)
+        .get_account_status_with_minor(&user_pubkey, tenant_id)
         .await?;
 
     match user {
-        Some((email, email_verified, status, suspended_reason)) => {
+        Some((
+            email,
+            email_verified,
+            status,
+            suspended_reason,
+            verified_minor,
+            verified_minor_at,
+        )) => {
             let account_status = if status.is_active() {
                 None
             } else {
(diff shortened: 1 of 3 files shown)


dcadenas (approved):
🔒 verified_minor now rides along on the account response, clean and well covered by tests.

PR #268, opened 2026-07-01
feat(auth): reliable email verification for in-app browsers — idempotent verify-on-load + in-app PIN fallback

### LINKED ISSUE #262: feat(auth): reliable email verification for in-app browsers — idempotent verify-on-load + in-app PIN fallback
## Problem

Android users who tap the email-confirmation link in a sandboxed in-app browser (Gmail Custom Tab / WebView) get stuck: the app doesn't open, "confirm email" does nothing, and they never finish registration. Failing for multiple users in the wild.

**Root cause** (confirmed via prod logs + mobile Crashlytics + code):

- The verify page completes via client JS (`onMount` → `POST /api/auth/verify-email`). In a sandboxed in-app browser that POST doesn't reliably fire/arrive, so **no request reaches keycast** — invisible in server logs *and* mobile Crashlytics.
- The headless device-code pipeline is healthy at scale (2-day prod: ~235 headless verifies ≈ 245 codes consumed; zero `headless_poll_delivery_failed`; `REDIS_KEY_PREFIX` unset → no key divergence). **Not** a Redis bug — verification just depends on JS running in a webview that may not run it.
- Android App Links don't escalate out of an in-app browser, so the deep link can't open the app either.

Email verification is load-bearing only for the 3 login gates + `export_key`, and as the trust anchor for **custodial-key recovery** (password reset + key export). Headless users are created already-verified — verification *is* the registration-completion step. So we keep verification but make completion robust against the in-app-browser dead-end. (Ban-evasion is **not** enforced in code; not a reason to keep it.)

## The two-clock model (must get right)

- **Verify window = 24h** — how long to prove email control. Applies to the link token, the pending `oauth_codes` row, **and** the new PIN. PIN brute-force is bounded by the attempt-cap, not time, so a 24h PIN is safe.
- **Exchange window = 10 min** — how long the app has to swap the minted OAuth authorization code for tokens (`auth.rs:1632` + Redis delivery `auth.rs:1653`, already matched). **Do not lengthen** — a long-lived auth code is a replay hole.
- **Bridge = idempotency / re-arming**, not a longer code: each verify (link GET or PIN submit) within 24h re-mints a *fresh* 10-min code.

## Part A — Idempotent verify-on-load (highest ROI, fixes all webviews)

- Verify on the link's **GET** server-side, removing the client-JS dependency (any webview, not just Android).
- **Idempotent for headless: do NOT delete the pending `oauth_codes` row on verify** (`auth.rs:1694`); re-arm (re-mint a fresh 10-min code) on each verify within 24h. Without this, a mail-scanner/link-preview prefetch verifies + materializes the user + writes the 10-min code + deletes the row before the user opens the app → permanently stuck, **strictly worse than today** (the GET-prefetch pattern rejected in #258). Idempotent re-verify makes prefetch harmless.

## Part B — In-app PIN fallback (robust for the worst contexts)

For webviews that never load/run the page, a different device, or a mangled link.

- At `headless_register` (`headless.rs` ~172, where the verification token is generated): also generate a **6-digit PIN**, store it **hashed** on the pending `oauth_codes` row. Email **both** the link and the PIN (same proof, two transports).
- New **`POST /api/headless/verify-pin { device_code, pin }`** in `headless.rs`:
  - **Look up the pending row by `device_code`** (unique per install; `idx_oauth_codes_device_code` exists → add `find_by_device_code` to `OAuthCodeRepository`). **Not** a global `{ pin }` lookup. `device_code` (the 64-char secret the app holds) is the real gate; the 6-digit PIN is **defense-in-depth**. A global pin-only lookup would be brute-forceable across all pending registrations — do not do this.
  - On PIN match, run the **same finalize path** as the link.
  - **Return the OAuth authorization code synchronously** in the response body → the app exchanges immediately, **Redis-independent** (also dodges the BASIC-tier Redis-restart mass-strand risk). Keep the `oauth_poll:{device_code}` push too for the link path.

## Shared refactor

Extract the finalize sequence (`auth.rs:1521-1695`: find pending → create user `email_verified=true` + personal_keys → mint 10-min code → for headless `setex oauth_poll:{device_code}`) into **`finalize_pending_registration(...)`**, called by both the token path and the PIN path. Guarantees identical behavior incl. duplicate-email handling. **Stop deleting the pending row** here (see Part A re-arm).

## Brute-force protection (net-new — keycast has no lockout primitive today)

- `METRICS.inc_login_failure()` (`auth.rs:3514`) is a Prometheus counter, **not** enforcement. Build the cap fresh and **shared-state** (Cloud Run min-instances=3 → an in-memory counter won't work).
- Add `pin_attempts INT NOT NULL DEFAULT 0` to `oauth_codes`; atomic `UPDATE ... SET pin_attempts = pin_attempts + 1 ... RETURNING`; reject at **5**. DB column over Redis (durable across a Redis restart; per-registration; dies with the row).
- On cap: invalidate the PIN, require **resend** (reuse the existing 5-min `email_verification_sent_at` cooldown) to mint a new PIN + reset the counter. ≈ 60 guesses/hr vs 1M space = negligible.
- Constant-time compare; store the PIN hashed (codebase already hashes verification tokens).
- **Uniform anti-enumeration errors**: never reveal whether `device_code` or `pin` was wrong, never reveal lockout state (same posture as resend at `auth.rs:1992`). Log failures via `auth_observability::record_auth_event_and_log` (event `email_pin_verify`, outcome `failure`) — same brute-force feed as #258.

## Interactions / reuse

- **Duplicate-email edge (#248 / #198 / #236):** the PIN path funnels through the shared finalize, so it inherits whatever #248 lands — reuse, don't re-implement. Key the PIN lookup by `device_code` (not email) so it adds no new duplicate-email ambiguity.

## Out of scope (separate follow-ups)

- Making verification **non-blocking** (let users in before verifying) — bigger product change, breaks the deferred model, weakens the recovery guarantee we're keeping.
- First-party `/api/auth/register` (browser-based, link works there) — don't widen the surface.
- ~9.5% `invalid_grant` on `/oauth/token`, and a Redis-restart alert (BASIC tier wipes `oauth_poll` keys) — flagged separately; file if wanted.

## Acceptance criteria

- [ ] Tapping the link verifies on GET (no client JS), idempotently, without deleting the pending row; re-clicking within 24h re-mints a fresh 10-min code.
- [ ] A prefetch of the verify URL does not strand the user.
- [ ] `POST /api/headless/verify-pin { device_code, pin }` finalizes via the shared path and returns the OAuth code synchronously.
- [ ] PIN valid for the 24h verify window; 5 failed attempts locks it until resend; uniform errors; hashed at rest; constant-time compare.
- [ ] Link and PIN paths produce identical results incl. duplicate-email handling.
- [ ] Tests cover: idempotent re-verify, prefetch-safety, PIN happy path, PIN attempt-cap, anti-enumeration, duplicate-email conflict.

## Mobile dependency

Blocks divinevideo/divine-mobile#5578 (in-app PIN entry UI + client wiring). Mobile needs the `verify-pin` endpoint, the synchronous OAuth-code return, and the corresponding `keycast_flutter` client method.

dcadenas:
## Summary

Some email-opening environments never finish the client-side verification flow, so registration stalls after the user taps the link.
This change verifies on the server when the link is opened (no page JavaScript required), adds a 6-digit in-app PIN fallback, and hardens pending-registration rows for production concurrency and replay.
It also reconciles the duplicate-pending-registration dedupe work with the revised PIN contract and normalizes a few terminal OAuth error codes.

Closes #262

## Motivation

Email verification is the registration-completion step for app sign-up.
Some mail clients and in-app browsers open the link in a context that does not complete the SPA flow, and App Links do not always escalate out of that context either.
Production does not corroborate a single Gmail-Custom-Tab root cause, so the safer claim is environmental: server-side GET removes dependence on page JavaScript and app-link handoff.

Measured load that makes the concurrency work load-bearing rather than defensive:

- Of 587 repeat registrations, **77 arrived within ten seconds** of the previous attempt (dedupe justification).
- Samsung Internet fetched each verification link about **fifteen times**, measured across thirty days of production request logs (`consumed_at` and idempotent re-arm are load-bearing).
- Migration backfill (measured 2026-08-11): the `DELETE` removes **587 superseded duplicate rows** of **2,976** unconsumed pending registrations, keeping the newest per tenant and normalized email so **2,389** survive and no email loses its most recent registration. Re-measure immediately before deploy; that table is live.

## What Changed

Verification is shared through `finalize_pending_registration` so the link and PIN paths behave the same.

- `GET /api/auth/verify-email` verifies on the server and returns HTML (headless), a redirect (third-party OAuth), or a session cookie (first-party). POST and the SPA page remain for in-flight emails.
- `POST /api/headless/verify-pin` and `POST /api/headless/resend-pin` for the in-app PIN transport.
- Duplicate pending registrations for the same tenant + normalized email are superseded instead of minting a second live row; superseded links get a clear page.
- PIN contract: `401 PIN_INVALID`, `423 PIN_LOCKED`, `410 PIN_EXPIRED`, `403 PIN_UNAVAILABLE`, `409 EMAIL_ALREADY_EXISTS`; separate `pin_resend_at`; lifetime `pin_failed_total`; atomic resend claim.
- Terminal OAuth errors: register duplicates return `409 EMAIL_ALREADY_EXISTS`; spent or unknown authorization codes return RFC 6749 `invalid_grant` instead of password-style copy.
- Concurrency/replay: idempotent exchange-code re-arm, `consumed_at` terminal marker, atomic PIN attempt reservation, uniform bcrypt timing on verify-pin 401 paths, `ON CONFLICT` user materialization.

## Testing

Independent validation at `e10464e`:

- **42/42** server-side assertions against a fresh image built from that SHA (link states and dedupe, JS-less GET materializing the user, full PIN contract including correct-PIN-accepted, supersession preserving `pin_failed_total` while resetting `pin_attempts` and `pin_resend_at`, first-resend-not-swallowed, four concurrent finalizations minting exactly one exchange code and one user, expired distinct from superseded, both `invalid_grant` paths).
- Android app leg on a headless emulator: PIN entry completed sign-in to the feed.
- Dedupe migration proven against seeded production-shaped data.
- Pre-push: `cargo fmt`, clippy (`-D warnings -A deprecated`), `cargo test --workspace`, release build.

Not included: a formal independent code review frozen at `e10464e`. The last independent review was frozen at `193d844`. Green checks do not replace that.

JS-less curl against the verification GET is a reproduction of the failure mode (client JS never runs), not proof that any one mail client causes it.

## Review / approval note

The existing APPROVED review was given against `fb87d0d`.
This force-push replaces that lineage with `e10464e`, so that approval does not refer to the code that will merge.
Maintainer is merging on their own local testing; no re-review requested from this update.

## Risks

- Migration backfill deletes superseded duplicate pending rows; re-measure counts immediately before deploy.
- Pending rows live the full registration window and can mint short-lived exchange codes; `consumed_at` stops re-mint after token redemption.
- PIN guessing is bounded by per-PIN and lifetime caps; resend clears only the per-PIN counter.

## Visuals

- [x] No visual change in `web/`
- [x] New surface is a minimal server-side HTML status page for webviews that open the GET link

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -1990,11 +1990,16 @@ pub async fn finalize_pending_registration(
     // terminal guard above already refuses re-mint once a code has been redeemed.
     // `freshly_minted` records whether THIS call created the code (vs reused an existing live one).
     // It gates the delivery-failure cleanup below: a reused code is owned by a prior finalize.
-    let (new_code, freshly_minted) = if let Some(existing) = oauth_code_repo
-        .find_live_exchange_code(tenant_id, &oauth_data.user_pubkey, &oauth_data.client_id)
-        .await?
+    let (new_code, code_expires_at, freshly_minted) = if let Some((existing, expires_at)) =
+        oauth_code_repo
+            .find_live_exchange_code_with_expiry(
+                tenant_id,
+                &oauth_data.user_pubkey,
+                &oauth_data.client_id,
+            )
+            .await?
     {
-        (existing, false)
+        (existing, expires_at, false)
     } else {
         let minted: String = rand::thread_rng()
             .sample_iter(&rand::distributions::Alphanumeric)
@@ -2018,12 +2023,17 @@ pub async fn finalize_pending_registration(
             is_headless: oauth_data.is_headless, // Inherit from original registration
         };
         oauth_code_repo.store(store_params).await?;
-        (minted, true)
+        (minted, code_expires_at, true)
+    };
+    let redis_ttl_seconds = if freshly_minted {
+        600
+    } else {
+        (code_expires_at - Utc::now()).num_seconds().max(1) as u64
     };
 
     if let Some((device_code, redis)) = strict_delivery {
         let key = format!("oauth_poll:{}", device_code);
-        if let Err(e) = redis.setex(&key, 600, &new_code).await {
+        if let Err(e) = redis.setex(&key, redis_ttl_seconds, &new_code).await {
             tracing::error!(
                 event = "headless_poll_delivery_failed",
                 tenant_id = tenant_id,
@@ -2057,7 +2067,7 @@ pub async fn finalize_pending_registration(
         // is also returned synchronously in the response body (so a Redis miss is not fatal).
         if let Some(redis) = redis {
             let key = format!("oauth_poll:{}", device_code);
-            if let Err(e) = redis.setex(&key, 600, &new_code).await {
+            if let Err(e) = redis.setex(&key, redis_ttl_seconds, &new_code).await {
                 tracing::warn!("Failed to store OAuth code in Redis for polling: {}", e);
                 // Continue - redirect flow / synchronous return still works.
             } else {
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -6198,6 +6198,106 @@ mod tests {
         );
     }
 
+    /// Acceptance: the link path and the PIN path produce identical materialization, because both
+    /// funnel through finalize_pending_registration (keycast#262). The only difference is the
+    /// headless Redis-delivery strictness, which does not change the user/keys/code outcome.
+    #[cfg(feature = "integration-tests")]
+    #[tokio::test]
+    async fn test_link_and_pin_paths_finalize_identically() {
+        let pool = create_test_db().await;
+        let auth_state = create_test_auth_state_with_redis(pool.clone()).await;
+        let redis = auth_state.state.redis.clone();
+        let repo = keycast_core::repositories::OAuthCodeRepository::new(pool.clone());
+
+        let email_a = format!("identical-link-{}@example.com", Uuid::new_v4());
+        let email_b = format!("identical-pin-{}@example.com", Uuid::new_v4());
+        let (pubkey_a, token_a, _dc_a) =
+            insert_headless_pending_registration(&pool, &email_a).await;
+        let (pubkey_b, token_b, _dc_b) =
+            insert_headless_pending_registration(&pool, &email_b).await;
+
+        let data_a = repo
+            .find_by_verification_token(&token_a, 1)
+            .await
+            .unwrap()
+            .unwrap();
+        let data_b = repo
+            .find_by_verification_token(&token_b, 1)
+            .await
+            .unwrap()
+            .unwrap();
+
+        // A finalizes via the link path's delivery mode; B via the PIN path's.
+        let ra = super::finalize_pending_registration(
+            &pool,
+            redis.as_ref(),
+            1,
+            &data_a,
+            super::HeadlessDelivery::RedisRequired,
+        )
+        .await
+        .expect("link-path finalize succeeds");
+        let rb = super::finalize_pending_registration(
+            &pool,
+            redis.as_ref(),
+            1,
+            &data_b,
+            super::HeadlessDelivery::RedisBestEffort,
+        )
+        .await
+        .expect("pin-path finalize succeeds");
+
+        assert_eq!(ra.is_headless, rb.is_headless);
+        assert!(!ra.new_code.is_empty() && !rb.new_code.is_empty());
+
+        // Both paths materialize an identically-shaped result: verified user + 1 personal key +
+        // a fresh headless 10-minute exchange code, with the pending row preserved.
+        for (pubkey, code) in [(&pubkey_a, &ra.new_code), (&pubkey_b, &rb.new_code)] {
+            let (verified,): (bool,) = sqlx::query_as(
+                "SELECT email_verified FROM users WHERE pubkey = $1 AND tenant_id = 1",
+            )
+            .bind(pubkey)
+            .fetch_one(&pool)
+            .await
+            .unwrap();
+            assert!(verified, "user must be materialized as verified");
+
+            let (key_count,): (i64,) = sqlx::query_as(
+                "SELECT COUNT(*) FROM personal_keys WHERE user_pubkey = $1 AND tenant_id = 1",
+            )
+            .bind(pubkey)
+            .fetch_one(&pool)
+            .await
+            .unwrap();
+            assert_eq!(key_count, 1, "personal_keys must be created exactly once");
+
+            let minted = repo
+                .find_valid(1, code)
+                .await
+                .unwrap()
+                .expect("a fresh exchange code must be minted");
+            assert!(minted.is_headless);
+        }
+
+        assert!(
+            repo.find_by_verification_token(&token_a, 1)
+                .await
+                .unwrap()
+                .is_some(),
+            "link-path pending row must be preserved"
+        );
+        assert!(
+            repo.find_by_verification_token(&token_b, 1)
+                .await
+                .unwrap()
+                .is_some(),
+            "pin-path pending row must be preserved"
+        );
+
+        cleanup_verify_email_test_data(&pool, &pubkey_a, &token_a).await;
+        cleanup_verify_email_test_data(&pool, &pubkey_b, &token_b).await;
+    }
+
     #[cfg(feature = "integration-tests")]
     async fn insert_pending_oauth_registration(
         pool: &PgPool,
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -1811,42 +1811,35 @@ pub async fn finalize_pending_registration(
     // terminal guard above already refuses re-mint once a code has been redeemed.
     // `freshly_minted` records whether THIS call created the code (vs reused an existing live one).
     // It gates the delivery-failure cleanup below: a reused code is owned by a prior finalize.
-    let (new_code, code_expires_at, freshly_minted) = if let Some((existing, expires_at)) =
-        oauth_code_repo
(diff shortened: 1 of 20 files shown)


api/src/api/http/oauth.rs
```diff
@@ -2710,8 +2713,13 @@ async fn handle_authorization_code_grant(
         pending_email_val.clone()
     } else {
         // Normal token exchange (existing user, not registration)
-        // Delete the authorization code (one-time use)
-        oauth_code_repo.delete(tenant_id, code).await?;
+        // One-time exchange and pending-row completion must be one authoritative DB transition.
+        if !oauth_code_repo
```
NotThatKindOfDrLiz, line 3118: Blocking: this consumes the verified-registration exchange code before BYOK key material is validated/stored. For BYOK registrations, `finalize_pending_registration` creates a verified user without `personal_keys`, then this branch deletes the exchange code and marks the pending row consumed. If `create_oauth_authorization_and_token` later rejects a missing/malformed/mismatched nsec from `code_verifier`, the user is stranded: the code is gone and re-clicking the email link cannot re-arm because the pending row is terminal. Please validate/create the missing BYOK `personal_keys` before consuming the code, or make code redemption plus key creation one atomic transaction, and add a regression test that a failed BYOK exchange preserves retryability.

api/src/api/http/auth.rs
```diff
@@ -1955,52 +2094,285 @@ pub async fn verify_email(
...
+            StatusCode::OK,
+            "Email verified!",
+            "Your email is verified. You can return to the app to continue.",
+        ),
+        Ok(VerifyOutcome::AlreadyVerified) => verify_html_page(
+            StatusCode::OK,
+            "Already verified",
+            "Your email is already verified. You can return to the app or log in.",
+        ),
+        Ok(VerifyOutcome::Processing) => verify_html_page(
```
NotThatKindOfDrLiz: Low severity, but this is exactly the JS-less/no-address-bar webview case this PR is trying to handle. The POST path tells clients to retry after 1s; the GET page should use `verify_html_page_with_refresh(..., Some(1))` so users who click while bcrypt is still finishing are not left with only a manual refresh instruction.

api/src/api/http/auth.rs
```diff
@@ -1955,52 +2094,285 @@ pub async fn verify_email(
...
+        | Err(AuthError::EmailNotVerified)
+        | Err(AuthError::UserNotFound)
+        | Err(AuthError::MissingToken)
+        | Err(AuthError::InvalidToken)
+        | Err(AuthError::TokenExpired)
+        | Err(AuthError::DuplicateKey)
+        | Err(AuthError::InvalidEmail)
+        | Err(AuthError::BadRequest(_))
+        | Err(AuthError::Forbidden(_))
+        | Err(AuthError::RegistrationExpired)
```
NotThatKindOfDrLiz: Messaging issue: `RegistrationExpired` means the async bcrypt worker likely died and the account cannot complete because `password_hash` is still NULL. The POST mapping correctly returns 410 / “Please register again,” but this GET catch-all tells the user the link may already be verified and they can log in. Please give this variant its own “register again” page so GET and POST agree.

NotThatKindOfDrLiz (changes requested):
Requesting changes for one blocking BYOK retryability regression. The inline comments cover the required fix plus two low-severity GET verification UX cleanups. The main blocker is that a verified BYOK flow can consume the exchange code and terminally mark the pending row before the BYOK nsec is validated/stored, stranding retryability on token-exchange failure.

api/src/api/http/oauth.rs
dcadenas: Addressed in 7a81724: exchange codes remain one-shot, but the matching pending registration is marked consumed only after authorization and token issuance succeeds. A deterministic downstream issuance-failure regression verifies that the pending link/PIN remains active and can mint a replacement code.

api/src/api/http/auth.rs
dcadenas: Superseded by 4af2727: GET now handles only OAuth/headless pending rows and redirects every first-party token to the interactive page, whose POST owns processing and retry behavior. The GET handler can no longer produce the first-party Processing outcome described here.

dcadenas:
@NotThatKindOfDrLiz
> Requesting changes for one blocking BYOK retryability regression. The inline comments cover the required fix plus two low-severity GET verification UX cleanups. The main blocker is that a verified BYOK flow can consume the exchange code and terminally mark the pending row before the BYOK nsec is validated/stored, stranding retryability on token-exchange failure.

The blocking retryability issue is addressed by 34cc7e7, which delays terminal pending-row consumption until issuance succeeds and adds a downstream-failure re-arm regression. Commits bfa67ae and bda7b30 also bound verify-PIN bcrypt work and keep first-party session issuance on interactive POST.

api/src/api/http/auth.rs
dcadenas: Superseded by 4af2727: first-party tokens are no longer finalized by GET, so the RegistrationExpired path is reached only by the interactive POST, which already returns the register-again response. GET redirects those tokens to that interactive flow instead of rendering the misleading catch-all page.

NotThatKindOfDrLiz (approved):
Re-reviewed the updated auth/email verification changes. The previous blocking BYOK retryability issue and GET verification nits are addressed, and I did not find a remaining blocker. Approving; not merging.

PR #278, opened 2026-07-01
feat(web): rename dashboard settings link + page heading to "Account & Security"

mbradley:
Draft — pending code review before marking ready.

## What

Renames the user-facing dashboard settings entry point (and the destination page heading/title) from **"Security & Bluesky"** to **"Account & Security"**. Label-only; the `/settings/security` route path is unchanged so existing links and bookmarks keep working.

## Why

The gear-icon link on the dashboard read "Security & Bluesky", which enumerates two of the things inside `/settings/security` rather than naming the container. That page actually holds email verification, change password, export private key, change private key, and the Bluesky account card (claim handle, provision atproto identity, cross-posting).

"Account & Security" is the widely-recognized convention for "where I manage my password, email, identity, and keys" (Apple "Sign-In & Security", Microsoft, and Coinbase/Kraken-style "Account & Security" in crypto/custody products — the closest analog for a key-custody app). It leads with the container and keeps "Security" visible, which also stays consistent with the `/security` path and the pages sensitive key operations.

Historical note: the link originally read "Security Settings"; it was changed to "Security & Bluesky" when the atproto card was added (rabble, `6457683`, 2026-03-28) to advertise the new feature. This restores a container-style label.

## Changes

- `web/src/routes/+page.svelte` — gear link label (`Gear` icon + href unchanged).
- `web/src/routes/settings/security/+page.svelte` — `<h1>` and `<svelte:head><title>`.
- `web/src/routes/+page.svelte` — two inline "Your Keys Explained" links that pointed users to "Security Settings" for nsec export, renamed for wayfinding consistency (there is no longer any "Security Settings" label to look for).

Left as-is: the `🔐 Unlock Security Settings` sub-card heading, which names the security-gated key operations specifically and stays coherent since the label keeps "Security".

## Verification

- `svelte-check` clean on the two edited files (no new type errors).
- Visual render not verified locally (the label is behind logged-in identity state and needs the full auth stack); to be confirmed on staging after merge.

### DIFF
--- a/web/src/routes/+page.svelte
+++ b/web/src/routes/+page.svelte
@@ -417,7 +417,7 @@ $effect(() => {
 							{/if}
 							<a href="/settings/security" class="identity-link">
 								<Gear size={16} />
-								<span>Security &amp; Bluesky</span>
+								<span>Account &amp; Security</span>
 							</a>
 						</div>
 					{/if}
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
✅ clearer account settings wording

Issue #265, opened 2026-06-30
feat(minor-safety): lifecycle - age-up (->16) / revocation transitions on verified_minor


mbradley:
Part of Divine's protected-minor (13-15) safeguards epic (tracked privately: divinevideo/support-trust-safety#173). Equal sibling of the relay-manager T&S-visibility ticket: divinevideo/divine-relay-manager#141.

Keep the protected-minor state honest over time: when an approved minor ages up to 16, or approval is revoked, `verified_minor` and the dependent protections must lift/apply automatically.

## Scope
- Define + implement the age-up (->16) transition that clears `verified_minor` + protections. Open design question: `verified_minor` is set at minor-account creation with no expiry, and the 13-15 band is captured at review time (no stored DOB) - decide the mechanism (time-based from `verified_minor_at` + band, scheduled re-review, or manual).
- Revocation: if approval is reversed / re-review denied, clear `verified_minor` and re-apply restrictions.

Clients react to the flag flipping via the protected-minor state (support-trust-safety#174). Relates to keycast#263 (flag exposure).

## Acceptance
An account aging to 16 or having approval revoked has `verified_minor` cleared automatically, so protections lift across clients.

mbradley:
Note for the client protections (#175/#176): the clients persist last-known protected-minor status on-device and keep protections sticky, lifting them only when keycast positively reports the account is no longer a minor. So this age-up/revocation transition is the authoritative signal that clears client-side protection. Make sure a cleared `verified_minor` is observable to clients via `GET /user/account` so the protections actually lift when it fires.

mbradley:
closing the age-up half as won't-do.

the product decision landed last week (Liz's call): we're not capturing a birthdate or per-year age. nothing stored can compute a 16th birthday, so the automatic →16 transition this issue scoped can't be built, and we're not adding an age signal to enable it.

the revocation half of this issue already shipped and stands: #280 (clear primitive + revocation endpoint) and #295 (durable admin_audit_events). clearing `verified_minor` is a human action on the admin side, and clients lift protections when the flag clears, as designed.

if a protected minor ages out, they go through support and a moderator clears the flag on that existing path. no scheduled expiry, no computed age-up.

PR #280, opened 2026-07-02
feat(minor-safety): verified_minor lifecycle - clear primitive + revocation endpoint (#265)

mbradley:
Implements the keycast backend of #265 (part of the protected-minor epic, divinevideo/support-trust-safety#173).

## What this does

Adds the missing path that clears `verified_minor`. Until now the flag was a one-way door: set at minor-account creation, never cleared, so protected-minor protections were permanent-once-granted with no remediation short of a manual DB edit.

- `UserRepository::clear_verified_minor(pubkey, tenant_id)` sets `verified_minor = false`, `verified_minor_at = null`. Idempotent (no-op on already-cleared or never-minor accounts, 404 on unknown user), and never touches account status.
- `DELETE /api/admin/users/:pubkey/verified-minor` (service-token auth) is the endpoint relay-manager calls on revocation. Optional `actor` (moderator hex pubkey, validated) and `reason` query params.
- Writes a durable `admin_audit_events` row when an `actor` is supplied (best-effort, log-only fallback when absent), with `reason` sanitized (bounded to 500 chars, control characters stripped) before it is logged or stored.

## Design

- **Clear-only, not clear-and-suspend.** The primitive serves both lifecycle transitions, which want opposite status outcomes: age-up leaves status active, revocation is reason-dependent (under-13 ban, mistaken approval unsuspend). relay-manager composes `set_user_status(...)` + clear (relay-manager#147). Clearing the flag already lifts the protected-minor protections; the account-status decision belongs to the caller.
- **No migration** (the `verified_minor` columns already exist).
- **Claim-link closure (review follow-ups).** Clearing a still-unclaimed minor also invalidates outstanding claim links, and claim-token consumption is now atomic with validity: the consume re-checks used/invalidated/expired under the row lock in one transaction with the user claim, so a link invalidated mid-flight can never complete a claim, and a failed claim never burns the token. New `UserRepository::claim_account_consuming_token`; `claim_post` treats it as the authoritative gate and re-classifies dead tokens to their state-specific error page.

## Scope and deferrals

- **Automatic age-up (to 16) is deferred to support-trust-safety#179.** No birthdate or exact age is captured anywhere in the age-review flow, by design (parent email plus a consent video with a verbal "13 to 15"; relay-manager stores only a coarse `age_13_15` band; keycast stores only the flag plus the approval timestamp). keycast cannot compute a 16th birthday from that, so automatic age-up needs a policy decision, not a backend change. This PR delivers the mechanism that makes the lift automatic-on-clear.
- **relay-manager revoke wiring** (the reason-dependent "re-apply restrictions" status change, plus passing actor/reason): relay-manager#147.
- **Durable audit for suspend/ban** (the same bar-raise applied to `set_user_status_admin`): keycast#279.

## Tests

- 4 repository tests (clear, idempotent, not-found, status-untouched) and 8 endpoint tests (clear and response, missing/wrong auth, 404, audit-row written, log-only fallback, malformed-actor 400, reason sanitization). The audit and sanitization assertions were mutation-verified.
- 6 claim-consume race regression tests (`claim_consume_race_test.rs`): invalidated-between-classify-and-consume refusal with no user mutation, expired/reused refusal, consume rollback on an unclaimable user, and a concurrent invalidate-vs-claim exactly-one-wins invariant.
- Full `keycast_core` and `keycast_api` integration suites pass locally.

Spec: `docs/superpowers/specs/2026-07-01-verified-minor-lifecycle-design.md`.

### DIFF
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -1963,8 +1963,10 @@ fn is_unsafe_format_char(c: char) -> bool {
         '\u{200B}'..='\u{200F}'   // zero-width space .. right-to-left mark
         | '\u{2028}'..='\u{2029}' // line / paragraph separator
         | '\u{202A}'..='\u{202E}' // bidi embeddings / overrides
+        | '\u{2060}'..='\u{2064}' // word joiner + invisible math operators
         | '\u{2066}'..='\u{2069}' // bidi isolates
         | '\u{FEFF}'              // zero-width no-break space / BOM
+        | '\u{FFF9}'..='\u{FFFB}' // interlinear annotation anchors
     )
 }
 
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -1954,11 +1954,27 @@ pub struct ClearVerifiedMinorParams {
     pub reason: Option<String>,
 }
 
-/// Bound and strip control characters from a caller-supplied reason before it
-/// is logged or persisted (prevents log injection and unbounded audit rows).
-/// Returns None when empty after cleaning.
+/// True for Unicode bidi/zero-width format characters that `char::is_control`
+/// (category Cc) misses. Stripping these prevents log/console display spoofing
+/// via right-to-left overrides and invisible characters.
+fn is_bidi_or_zero_width(c: char) -> bool {
+    matches!(c,
+        '\u{200B}'..='\u{200F}'   // zero-width space .. right-to-left mark
+        | '\u{202A}'..='\u{202E}' // bidi embeddings / overrides
+        | '\u{2066}'..='\u{2069}' // bidi isolates
+        | '\u{FEFF}'              // zero-width no-break space / BOM
+    )
+}
+
+/// Bound and strip control + bidi/zero-width characters from a caller-supplied
+/// reason before it is logged or persisted (prevents log injection, display
+/// spoofing, and unbounded audit rows). Returns None when empty after cleaning.
 fn sanitize_reason(raw: Option<String>) -> Option<String> {
-    let cleaned: String = raw?.chars().filter(|c| !c.is_control()).take(500).collect();
+    let cleaned: String = raw?
+        .chars()
+        .filter(|c| !c.is_control() && !is_bidi_or_zero_width(*c))
+        .take(500)
+        .collect();
     let trimmed = cleaned.trim();
     if trimmed.is_empty() {
         None
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -1958,11 +1958,7 @@ pub struct ClearVerifiedMinorParams {
 /// is logged or persisted (prevents log injection and unbounded audit rows).
 /// Returns None when empty after cleaning.
 fn sanitize_reason(raw: Option<String>) -> Option<String> {
-    let cleaned: String = raw?
-        .chars()
-        .filter(|c| !c.is_control())
-        .take(500)
-        .collect();
+    let cleaned: String = raw?.chars().filter(|c| !c.is_control()).take(500).collect();
     let trimmed = cleaned.trim();
     if trimmed.is_empty() {
         None
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -1954,12 +1954,14 @@ pub struct ClearVerifiedMinorParams {
     pub reason: Option<String>,
 }
 
-/// True for Unicode bidi/zero-width format characters that `char::is_control`
-/// (category Cc) misses. Stripping these prevents log/console display spoofing
-/// via right-to-left overrides and invisible characters.
-fn is_bidi_or_zero_width(c: char) -> bool {
+/// True for Unicode bidi / zero-width / line-separator characters that
+/// `char::is_control` (category Cc) misses. Stripping these prevents log/console
+/// display spoofing via right-to-left overrides, invisible characters, and
+/// newline-equivalent separators (U+2028/U+2029).
+fn is_unsafe_format_char(c: char) -> bool {
     matches!(c,
         '\u{200B}'..='\u{200F}'   // zero-width space .. right-to-left mark
+        | '\u{2028}'..='\u{2029}' // line / paragraph separator
         | '\u{202A}'..='\u{202E}' // bidi embeddings / overrides
         | '\u{2066}'..='\u{2069}' // bidi isolates
         | '\u{FEFF}'              // zero-width no-break space / BOM
@@ -1972,7 +1974,7 @@ fn is_bidi_or_zero_width(c: char) -> bool {
 fn sanitize_reason(raw: Option<String>) -> Option<String> {
     let cleaned: String = raw?
         .chars()
-        .filter(|c| !c.is_control() && !is_bidi_or_zero_width(*c))
+        .filter(|c| !c.is_control() && !is_unsafe_format_char(*c))
         .take(500)
         .collect();
     let trimmed = cleaned.trim();
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -1944,6 +1944,106 @@ pub async fn set_user_status_admin(
     }))
 }
 
+// --- Clear verified_minor (service-token auth: age-up / revocation) ---
+
+/// Query params for the clear-verified_minor endpoint. Both optional; `actor`
+/// (the moderator's hex pubkey) drives the durable audit row.
+#[derive(Debug, Deserialize)]
+pub struct ClearVerifiedMinorParams {
+    pub actor: Option<String>,
+    pub reason: Option<String>,
+}
+
+/// Bound and strip control characters from a caller-supplied reason before it
+/// is logged or persisted (prevents log injection and unbounded audit rows).
+/// Returns None when empty after cleaning.
+fn sanitize_reason(raw: Option<String>) -> Option<String> {
+    let cleaned: String = raw?
+        .chars()
+        .filter(|c| !c.is_control())
+        .take(500)
+        .collect();
+    let trimmed = cleaned.trim();
+    if trimmed.is_empty() {
+        None
+    } else {
+        Some(trimmed.to_string())
+    }
+}
+
+pub async fn clear_verified_minor_admin(
+    tenant: crate::api::tenant::TenantExtractor,
+    State(auth_state): State<AuthState>,
+    headers: HeaderMap,
+    Path(pubkey): Path<String>,
+    Query(params): Query<ClearVerifiedMinorParams>,
+) -> ApiResult<Json<UserStatusResponse>> {
+    authorize_service_token(&headers)?;
+    let tenant_id = tenant.0.id;
+
+    // Validate the optional actor as a 64-char hex pubkey so a T&S action never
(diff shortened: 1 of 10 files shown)


NotThatKindOfDrLiz:
I approve of this direction and do not see a blocking issue in #280, but I am not marking the PR as formally approved here.

@dcadenas your approval is required for this PR to merge. I requested your review.

@mbradley is out of office, so if you find changes are needed, please make the changes and push them so we keep momentum. The exception is anything that requires a decision only mbradley can make; in that case, call out the decision point explicitly.

api/src/api/http/admin.rs
```diff
@@ -1944,6 +1944,120 @@ pub async fn set_user_status_admin(
...
+            return Err(ApiError::bad_request(
+                "actor must be a 64-character hex pubkey",
+            ));
+        }
+    }
+
+    let reason_clean = sanitize_reason(params.reason);
+
+    let user_repo = UserRepository::new(auth_state.state.db.clone());
+    user_repo.clear_verified_minor(&pubkey, tenant_id).await?;
```
dcadenas: Clearing `verified_minor` here does not cancel an outstanding claim link, so an account revoked before it is claimed can still be claimed and ends up as a normal account with no minor protections.

The primitive only touches the two minor columns plus `updated_at`:

https://github.com/divinevideo/keycast/blob/bc115a8fc7cb826d4acf39b1d374184db2ef37bb/core/src/repositories/user.rs#L1412-L1431

Neither it nor this handler invalidates the `account_claim_tokens` row, so a claim link issued before the revoke stays live.

Decide who owns closing that door.
Either document that the clear is deliberately claim-token-agnostic and rely on the relay-manager side to always pair it with a status change (relay-manager#147), or invalidate outstanding claim tokens for the pubkey inside `clear_verified_minor` when the account is still unclaimed.
`invalidate_valid_for_user` in `core/src/repositories/claim_token.rs` already does that.

One thing not to do: do not add a bare `WHERE verified_minor = TRUE` to the create-minor retry path expecting it to block a reissue.
That path is already blocked, because `is_unclaimed` embeds `verified_minor = TRUE` (`user.rs:1471`) and the endpoint returns a 409 once the flag is cleared.
The only open path left is redemption of an already-issued token.

api/src/api/http/admin.rs
```diff
@@ -1944,6 +1944,120 @@ pub async fn set_user_status_admin(
...
+        pubkey = %pubkey,
+        actor = ?params.actor,
+        reason = ?reason_clean,
+        "Admin cleared verified_minor"
+    );
+
+    // Durable audit row (best-effort) when an actor is supplied. The table's
+    // actor_pubkey is NOT NULL, so no actor -> log-only. A failed insert logs
+    // and the clear still succeeds. (keycast#279 raises the same bar for status changes.)
+    if let Some(actor) = params.actor.as_deref() {
```
dcadenas: A retry or a clear on an already-cleared account is not a true no-op.
The `UPDATE` is unconditional, so it bumps `updated_at` and reports a row changed even when `verified_minor` was already false:

https://github.com/divinevideo/keycast/blob/bc115a8fc7cb826d4acf39b1d374184db2ef37bb/core/src/repositories/user.rs#L1412-L1431

This handler then writes an `admin_audit_events` row on every call that supplies an actor, with no check on the prior state.
So a relay-manager retry after a lost response, or a duplicate revocation, writes a second clear event asserting a state change that never happened.
The shipped idempotency test only checks that the second call does not error, not that it has no side effects.

Audit only real transitions.
A no-op clear should not bump `updated_at` and should not write a second audit row.

Watch the fix.
Do not just add `WHERE verified_minor = TRUE` to the `UPDATE`.
That collapses missing-user and already-cleared into the same `rows_affected == 0` and breaks the `NotFound` contract.
Distinguish three cases in one statement: user missing (return `NotFound`), user present but already cleared (no-op, report no transition), and a real clear.
A CTE that selects the row for existence and conditionally updates works.
Then write the audit row here only when a real transition happened.

dcadenas (changes requested):
The clear-verified_minor primitive and endpoint are clean, but two lifecycle gaps need to close before this can merge.
Revoking an unclaimed minor account does not stop an outstanding claim link, so the account can still be claimed afterward and ends up with no minor protections.
And a retried or no-op clear still writes a fresh audit row and bumps the timestamp, so the safety trail shows state changes that never happened.
Both are called out inline with the concrete fix and the trap to avoid.

NotThatKindOfDrLiz:
Both blocking findings are addressed in `c0af314` (mbradley is OOO; pushing to keep momentum per @NotThatKindOfDrLiz). Rebased on `origin/main`; core (119) + api integration suites green locally.

**1. Outstanding claim link survives revocation** — closed in keycast, not deferred.

Worth flagging why I didn't take the "document it and rely on the relay-manager status change" option: I traced the redemption path, and claim redemption (`claim.rs::claim_post`) gates **only** on token validity — it never checks `status` or `verified_minor` (status is fetched afterward only to stamp the UCAN fact). So a relay-manager suspend/ban alone would **not** stop redemption; invalidating the token is the only thing that closes the door. The endpoint now calls `invalidate_valid_for_user` after the clear — run unconditionally (so a retry after a mid-request crash still closes the door), idempotent, a no-op for an already-claimed age-up account. Unlike the best-effort audit row this is treated as a safety invariant: a failure fails the request so the idempotent caller retries rather than leaving the link open.

**2. No-op clear writes a phantom audit row + bumps `updated_at`** — fixed as described, without the `WHERE verified_minor = TRUE` trap.

`clear_verified_minor` now distinguishes the three cases in one statement via a CTE (`existing` probes for the row; `changed` conditionally updates only a set flag and reports via `RETURNING`): missing user → `NotFound`; already-cleared → no-op, no `updated_at` bump, returns `false`; real clear → returns `true`. The handler writes the audit row only on a real transition.

**New/changed tests**
- repo: no-op-does-not-bump-`updated_at`; transition-return assertions on existing tests
- endpoint: retry-writes-no-second-audit-row; invalidates-outstanding-claim-token (asserts no valid token remains + `invalidated_by == actor`)

Spec (`docs/superpowers/specs/2026-07-01-verified-minor-lifecycle-design.md`) updated to match, including a new "Claim-link closure" section documenting the redemption-ignores-status reasoning.

@dcadenas re-requesting your review.

api/src/api/http/admin.rs
mbradley: Addressed in `c0af314` (@NotThatKindOfDrLiz). Confirming my read of why it closes the finding: `claim_post` gates only on token validity, never on `status` or `verified_minor`, so a relay-manager status change alone would not stop redemption — invalidating the token is the only thing that closes the door. `c0af314` calls `invalidate_valid_for_user` after the clear, run unconditionally (a retry after a mid-request crash still closes it) and treated as a safety invariant (failure fails the request so the idempotent caller retries), with a test asserting no valid token remains. On top of that I added `3d78fbc`: tests that the actor-less path still invalidates and that a foreign-tenant token is not swept.

api/src/api/http/admin.rs
mbradley: Addressed in `c0af314` (@NotThatKindOfDrLiz), using the CTE rather than the `WHERE verified_minor = TRUE`-only trap you flagged. Confirming my read of it: `clear_verified_minor` distinguishes all three cases in one statement — `existing` probes the row (missing → `NotFound`), `changed` updates only a set flag and reports via `RETURNING`, so already-cleared is a true no-op with no `updated_at` bump and returns `false`, and the handler writes the audit row only on a real transition. Tests cover no-op-does-not-bump-`updated_at` and retry-writes-no-second-audit-row. I re-reviewed this path closely and it holds.

mbradley:
Thanks @NotThatKindOfDrLiz! Both blocking findings are addressed in `c0af314`. I ran a fresh picky review pass on top and added `3d78fbc`, which closes two safety-invariant test gaps (actor-less invalidation, cross-tenant token isolation) and widens the reason format-char strip. `fmt`/`clippy` clean, all 14 endpoint + 6 repo tests green. @dcadenas re-requesting your review when you have a moment.

dcadenas (approved):
🧾 flag clears, the audit row stays. Ship it.

api/src/api/http/admin.rs
```diff
@@ -1944,6 +1944,169 @@ pub async fn set_user_status_admin(
...
+    // (age-up) account. Unlike the best-effort audit row this is a safety
+    // invariant: a failure fails the request so the idempotent caller retries
+    // rather than silently leaving the link open.
+    let claim_token_repo = ClaimTokenRepository::new(auth_state.state.db.clone());
+    let invalidated_by = params
+        .actor
+        .as_deref()
+        .unwrap_or("system:verified_minor_clear");
+    let invalidated = match claim_token_repo
+        .invalidate_valid_for_user(&pubkey, tenant_id, invalidated_by, reason_clean.as_deref())
```
NotThatKindOfDrLiz, line 2074: This invalidation closes the straight-line outstanding-link case, but it does not close the concurrent claim path. `/claim` can classify the token as valid, then this endpoint clears `verified_minor` and invalidates the token, and `claim_post` still proceeds to `claim_account` and `mark_used`. `mark_used` only checks `used_at IS NULL`; it does not re-check `invalidated_at IS NULL` or `expires_at > NOW()`, and the user update has already happened by then. Token consumption needs to be atomic with validity, and the claim flow should not mutate the user unless it successfully consumes a still-valid token, ideally in one transaction. Please add a regression test for invalidated-before-`mark_used` / concurrent revoke-vs-claim behavior.

NotThatKindOfDrLiz (changes requested):
Requesting changes for one blocking race in the claim-link closure. The straight-line invalidation path is good, and the audit/no-op fixes look sound, but concurrent claim redemption can still slip through unless token consumption is made atomic with validity.

api/src/api/http/admin.rs
mbradley: addressed in `cfe9f2e` (thank you, this was a real one). token consumption is now atomic with validity: `claim_account_consuming_token` runs a single transaction whose consume step re-checks the full predicate (`used_at IS NULL AND invalidated_at IS NULL AND expires_at > NOW()`) under the row lock, deliberately mirroring `invalidate_valid_for_user`'s predicates so exactly one side can win, and the user mutation only happens after a successful consume (rollback if the user row isn't claimable, so a failed claim never burns the token). `claim_post` treats the consume as the authoritative gate and re-classifies a token that dies mid-flight to its state-specific error page. regression tests cover your exact invalidated-before-mark_used sequence plus a concurrent revoke-vs-claim exactly-one-wins invariant. re-requested.

NotThatKindOfDrLiz:
Takeover cleanup pushed in `fae8c75` after rebasing on current `origin/main`.

What changed:
- Removed the now-unused `UserRepository::claim_account` method so the PR no longer leaves a stale non-atomic claim path next to `claim_account_consuming_token`.

Verification:
- GitHub `test`, semantic PR, and CLA checks are green on `fae8c75`.

Local targeted integration tests compiled but could not run because Docker/Postgres is unavailable on this machine (Docker socket missing). CI covered the full test job successfully.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup. The stale `claim_account` method is removed, the branch is rebased on current `origin/main`, and CI is green on `fae8c75`.

PR #281, opened 2026-07-02
fix(api): classify ATProto control-plane outages

### LINKED ISSUE #200: Fix ATProto enable dependency configuration failures
## Summary
`POST /api/user/atproto/enable` returns 503 when Keycast cannot reach the ATProto control-plane endpoint.

## Evidence
Cloud Run prod logs showed:

```text
POST /api/user/atproto/enable 503
Internal error: request failed: error sending request for url (http://127.0.0.1:3201/api/account-links/opt-in)
```

## Suspected Cause
The ATProto provisioning client defaults to `http://127.0.0.1:3201`, but that dependency appears unavailable in the Cloud Run deployment.

Relevant default:

```rust
const DEFAULT_ATPROTO_CONTROL_PLANE_URL: &str = "http://127.0.0.1:3201";
```

## Acceptance Criteria
- Production has an explicit, reachable `DIVINE_SKY_ATPROTO_CONTROL_PLANE_URL`, or the endpoint fails closed with a clearer dependency-unavailable response.
- The response explains that ATProto enablement is temporarily unavailable, not a generic internal error.
- Add a startup/config validation or health signal for the ATProto control-plane dependency.

dcadenas:
## Summary

Users enabling ATProto could see a generic Keycast outage when the ATProto control plane was unreachable or misconfigured.
This change keeps those failures scoped to ATProto: the enable/reenable/disable endpoints fail closed with a clear ATProto-specific 503, while the rest of Keycast keeps serving traffic.
Startup validation warns about missing or invalid control-plane configuration without making global readiness depend on this optional scoped dependency.

## Motivation

The root cause was that provisioning failures were not classified, so a missing or bad control-plane URL, a slow control plane, or an upstream error all became a generic response.
ATProto is an optional scoped dependency, so its failures should not look like a whole-server outage and should not make global readiness depend on the control plane.

- The provisioning client now rejects missing, non-http, ambiguous (query/fragment), and non-root-path control-plane URLs, because endpoint paths are appended to the base URL by string concatenation.
- Control-plane URL validation now has one source of truth in `keycast_api::atproto_provisioning`, so startup warnings and request-time fail-closed behavior cannot drift.
- All provisioning failures (transport, timeout, 5xx, and 4xx) now map to a scoped 503; non-dependency 4xx failures previously went through `AuthError::Internal`, whose response body discards the message, so the ATProto-specific text was lost.
- The client uses one shared reqwest client with explicit connect and request timeouts, so an unresponsive control plane fails fast into the scoped 503 instead of hanging the request.
- Scoped ATProto 503 responses preserve `Retry-After: 30` so clients still receive a retry hint.

## Related Issue

- Closes #200

## Testing

- [x] `cargo test -p keycast_api --test atproto_provisioning_test`
- [x] `cargo test -p keycast --bin keycast test_validate_environment_allows_atproto -- --test-threads=1`
- [x] `cargo test --workspace --verbose` attempted locally; non-DB tests passed before the DB-backed ATProto HTTP test binary failed to set up because local Docker/Postgres was unavailable (`PoolTimedOut` connecting to the test database). GitHub Actions runs the full CI environment for final verification.

## Visuals

- [x] No visual change

## Risks

Readiness still does not check ATProto control-plane reachability, which is intentional so an ATProto outage does not take unrelated login/signing traffic offline.
The request timeouts are fixed constants (3s connect, 10s request) rather than configurable, which is fine for the current single control-plane deployment.
Runtime ATProto outages return a scoped 503 from the affected endpoints.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -106,6 +106,18 @@ REDIS_URL=redis://redis:6379
 # true (development only): allow degraded per-instance in-memory fallback
 DPOP_REPLAY_FAIL_OPEN=false
 
+# ATProto provisioning control plane
+# Required in production. Keycast calls this service when users enable Bluesky.
+# Example: https://atproto-control-plane.internal
+# DIVINE_SKY_ATPROTO_CONTROL_PLANE_URL=
+
+# Optional: handle domain for provisioned Bluesky handles
+DIVINE_HANDLE_DOMAIN=divine.video
+
+# Shared bearer token used for Keycast <-> ATProto control-plane calls
+# Required if the control plane enforces service auth.
+# KEYCAST_ATPROTO_TOKEN=
+
 # Frontend runtime configuration (injected into HTML at runtime)
 # These are used by the SvelteKit frontend, not build-time variables
 
--- a/.env.example
+++ b/.env.example
@@ -107,7 +107,7 @@ REDIS_URL=redis://redis:6379
 DPOP_REPLAY_FAIL_OPEN=false
 
 # ATProto provisioning control plane
-# Required in production. Keycast calls this service when users enable Bluesky.
+# Optional scoped dependency. If missing or invalid, Keycast still starts and ATProto endpoints fail closed.
 # Example: https://atproto-control-plane.internal
 # DIVINE_SKY_ATPROTO_CONTROL_PLANE_URL=
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3400,6 +3400,7 @@ dependencies = [
  "tower-http",
  "tracing",
  "tracing-subscriber",
+ "url",
  "uuid",
  "zeroize",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3400,7 +3400,6 @@ dependencies = [
  "tower-http",
  "tracing",
  "tracing-subscriber",
- "url",
  "uuid",
  "zeroize",
 ]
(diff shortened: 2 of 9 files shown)


NotThatKindOfDrLiz:
Pushed cleanup commit da46303 to address the review bar before merge.

What changed:
- Extracted ATProto control-plane URL validation into `keycast_api::atproto_provisioning::validate_control_plane_base_url` and reused it from startup warnings, removing the duplicated validator and the redundant `keycast` direct `url` dependency.
- Restored `Retry-After: 30` on scoped ATProto provisioning 503 responses and updated assertions.
- Added guarded env-var removal in the missing-control-plane provisioning test so process env is restored after the test.
- Retitled the PR to `fix(api): classify ATProto control-plane outages` and refreshed the body; it still includes `Closes #200`.

Local verification:
- `cargo test -p keycast_api --test atproto_provisioning_test`
- `cargo test -p keycast --bin keycast test_validate_environment_allows_atproto -- --test-threads=1`

I also attempted `cargo test --workspace --verbose`; it reached the DB-backed ATProto HTTP test binary, then failed only because the local test DB was unavailable (`PoolTimedOut`). `bun run test` could not start dependencies because Docker is not running/reachable on this machine. CI should provide the full Postgres/Redis environment.

NotThatKindOfDrLiz (approved):
Approved after final pass. The cleanup commit removes duplicated URL validation, restores Retry-After on scoped ATProto 503 responses, guards the env-mutating test, keeps issue #200 linked, and current PR checks are green.

PR #283, opened 2026-07-09
fix(web): improve registered clients layout

dcadenas:
## Summary

The registered OAuth clients admin page could overflow its card when client IDs, redirect URIs, dates, and action buttons all competed for table width.
This change keeps the table contained, gives small screens a readable stacked layout, and makes the page use the normal Divine green theme instead of the old purple/blue fallback styling.

## What Changed

The page now constrains the registered-clients table instead of letting long values push columns outside the panel.
It also uses the shared Divine color tokens so the admin page matches the rest of the site.

- Added row labels for the responsive stacked table layout.
- Switched the table to fixed layout with explicit column widths on desktop.
- Allowed long client IDs and redirect URI chips to wrap inside their cells.
- Added mobile styles that hide the table header and render each row as labeled fields.
- Replaced local fallback colors with `--color-divine-*` design tokens for cards, inputs, buttons, icons, code chips, table borders, and modal states.

## Testing

I verified the layout in Chrome with mocked admin API responses and checked both desktop and mobile widths.
The full pre-push hook passed after resetting a stale local `keycast_test` database that had an old migration record not present on `origin/main`.

- `git diff --check`
- Playwright/system Chrome visual capture at `1440x1100` and `390x1000`
- Verified no page-level horizontal overflow and no registered-clients card overflow in Playwright
- Verified primary button and section icon color resolved to `rgb(39, 197, 139)`
- `git push --force-with-lease -u origin fix-registered-clients-ui` pre-push hook passed, including formatting, clippy, workspace tests, release build, and web build

`cd web && bun run check` still reports unrelated existing diagnostics in `QRCode.svelte` and `routes/(app)/teams/[id]/keys/[pubkey]/+page.svelte`.

## Risks

Risk is low because this only changes markup labels and CSS for one admin page.
The main behavior risk is visual: very long redirect URI values now wrap instead of expanding the table horizontally.

## Related Issue

- N/A

## Visuals

- [ ] No visual change

Playwright screenshots were captured locally for desktop and mobile verification and reviewed before opening this PR.

### DIFF
--- a/web/src/routes/admin/registered-clients/+page.svelte
+++ b/web/src/routes/admin/registered-clients/+page.svelte
@@ -494,39 +494,41 @@
 								<td class="date-cell" data-label="Created">{formatDate(c.created_at)}</td>
 								<td class="date-cell" data-label="Updated">{formatDate(c.updated_at)}</td>
 								<td class="actions-cell" data-label="Actions">
-									{#if editingId === c.id}
-										<button
-											class="btn-icon"
-											onclick={() => saveEdit(c)}
-											disabled={isSaving}
-											title="Save"
-										>
-											<FloppyDisk size={18} />
-										</button>
-										<button
-											class="btn-icon"
-											onclick={cancelEdit}
-											disabled={isSaving}
-											title="Cancel"
-										>
-											<X size={18} />
-										</button>
-									{:else}
-										<button
-											class="btn-icon"
-											onclick={() => startEdit(c)}
-											title="Edit"
-										>
-											<PencilSimple size={18} />
-										</button>
-										<button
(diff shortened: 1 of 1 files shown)


mbradley (approved):
lgtm. i pushed two small commits on top: one gives the actions column a fixed width so its buttons stop spilling past the card edge on desktop, the other groups the action buttons and edit-mode URI rows so they don't split across the label column in the stacked layout.

checked it in-browser at desktop, the 561-900px tablet band and mobile, in both display and edit mode: no page or card overflow, the fixed-layout columns resolve cleanly, and svelte-check is clean on the file.

Issue #183, opened 2026-04-29
Follow-up: add DB-state assertions for rejected registered_clients update/delete paths


NotThatKindOfDrLiz:
PR #175 proves the admin gate fires before writes on the create path by asserting that a forbidden create does not insert a row.

Follow-up:
- extend the same style of DB-state assertion to the rejected update and delete paths
- for update: create a row, attempt forbidden update, verify the row is unchanged
- for delete: create a row, attempt forbidden delete, verify the row still exists

This is non-blocking regression-hardening for the registered_clients admin gate tests.

sergey3bv:
Hey, @NotThatKindOfDrLiz, if you don't mind I will work on this issue.

cc @dcadenas

sergey3bv:
Hey, @NotThatKindOfDrLiz, I opened [a PR](https://github.com/divinevideo/keycast/pull/215), could you please take a look?

cc @dcadenas

Issue #182, opened 2026-04-29
Follow-up: add positive full-admin create/update/delete HTTP tests for registered_clients


NotThatKindOfDrLiz:
PR #175 adds the negative HTTP-layer gate coverage requested in #169 and includes positive sanity checks for list and pattern-test.

Follow-up:
- add positive full-admin HTTP tests for create, update, and delete
- keep the same real-router / real-handler test shape used in PR #175
- assert success on the mutation endpoints so the suite has full symmetry across the registered_clients admin surface

This is non-blocking test-depth follow-up work.

sergey3bv:
Hey, @NotThatKindOfDrLiz, if you don't mind I will work on the issue.

cc @dcadenas

sergey3bv:
Hey, @NotThatKindOfDrLiz, I opened [a PR](https://github.com/divinevideo/keycast/pull/217), could you please take a look?

cc @dcadenas

Issue #176, opened 2026-04-29
ci(e2e): cache Playwright browser install for the nightly workflow


NotThatKindOfDrLiz:
PR #173 runs `npx playwright install --with-deps chromium` on every execution.

Follow-up:
- evaluate caching the Playwright browser install keyed by the relevant e2e package inputs
- keep it only if the runtime savings justify the added YAML/cache complexity

This is a non-blocking optimization.

sergey3bv:
Hey, @NotThatKindOfDrLiz, I opened [a PR](https://github.com/divinevideo/keycast/pull/194) to solve the issue, could you please take a look?

cc @dcadenas

PR #284, opened 2026-07-09
feat(minor-safety): server-side verified_minor DM containment (sign/encrypt gate)

mbradley:
## Summary

Server-side DM containment for `verified_minor` accounts (divinevideo/support-trust-safety#183). For a verified_minor account, Keycast refuses to sign or encrypt DM-shaped events whose recipients fall outside the approved official set, across every signing surface. A non-minor account is unaffected.

**Why this is the real lever:** Keycast's headless signing API is reachable with just the account's own email/password (public CORS), so for a custodial minor the shipped app-side DM restriction (#176, mobile #5754 / web #474) is friction, not containment. The paired app-side change (#182) removes the raw-key escape; this PR closes the headless-signer path. Together they gate the epic's custodial-containment claim (#173).

## The binding insight (why not just check the `p` tag)

A DM's recipient is bound by whoever can *decrypt* it, not by the attacker-controlled `p` tag. The NIP-44 / NIP-04 conversation key is **symmetric ECDH**, so a colluding non-approved party can compute the minor↔them key from their own secret key plus the minor's public key, and forge minor-keyed ciphertext without ever calling Keycast. So a naive "every `p` tag must be approved" check is spoofable: hide the real payload behind a decoy approved `p` tag. Each shape is therefore gated at the point its true recipient is actually determinable.

## Enforcement points and DM shapes covered

- **`nip04_encrypt` / `nip44_encrypt`** — recipient must be the user or a pinned official. This is the primitive that produces DM ciphertext, so it blocks a non-colluding minor from encrypting to a stranger at all.
- **kind 13 (NIP-59 seal)** — the seal names no recipient in its tags; its content is NIP-44 ciphertext under the conversation key of the unstated receiver. The recipient is recovered by trial-decrypting the content against the user's conversation keys with each approved pubkey. NIP-44 v2 is authenticated (HMAC), so this is deterministic: decrypts under an approved key ⇒ the recipient is that approved party (cryptographically bound); decrypts under none ⇒ refused. The inner rumor's `p` tags and the seal's own `p` tags must also all be approved. This is the real NIP-17 containment point — the one kind a conformant client actually asks the signer to sign.
- **kind 4 (NIP-04 DM)** — a complete, normal-client-rendered DM whose recipient *is* the `p` tag (NIP-04 content is unauthenticated AES-CBC, so the reader cannot be verified by decryption). Every `p` tag must be approved, keeping a minor-signed kind-4 facially addressed to an approved party; the gated encrypt primitive blocks producing the ciphertext for a non-colluding stranger. This is a live divine send path (video-share-via-DM, NIP-04 fallback).
- **kinds 14/15 (NIP-17 rumor) and 1059 (gift wrap)** — **refused outright.** A conformant client never asks a remote signer to sign these with the user's key: rumors are unsigned (NIP-59), and gift wraps are signed by a one-time ephemeral key, not the sender's. Signing one with the minor's key only serves a covert channel, and a `p`-tag check there is pure false confidence. The legitimate NIP-17 DM to an approved official flows entirely through `nip44_encrypt` (gated) + the kind-13 seal (gated).

**Ingress (decrypt / gift-wrap unwrap) is deliberately untouched** — #183 scopes containment to egress. `nip04_decrypt`, `nip44_decrypt`, and `nip17_unwrap_batch` still work for a minor (reading inbound DMs is not egress).

The approved set is `{ the user themselves, Divine HQ, Divine Moderation }`. Self is included because NIP-17 wraps a copy of every message to the sender, and a self-DM can only be read by the user.

## Accepted ceiling (documented, not closable server-side)

A minor can freely sign public posts (kind 1), whose content is arbitrary. So a minor + a colluding recipient can always steganographically encode a payload the recipient extracts out-of-band, through a public post — no server-side gate can close that, and it is not a DM shape any normal client renders as a DM. This gate's achievable job is narrower: Keycast will not produce a DM-shaped, normal-client-deliverable artifact addressed to a non-approved recipient.

## Approved set: pin-only server-side (design decision)

The clients approve a recipient via **pin ∩ live NIP-05** (#4948): the pin blocks attacker addition, the live NIP-05 leg is a revocation lever. Server-side I enforce the **pin only**, deliberately:

- A live NIP-05 lookup in the signing hot path adds latency and a hard network dependency to a critical op, and a resolver failure would force a fail-open-vs-fail-closed decision on every single DM sign.
- The pin is what actually answers "is this recipient approved" — that needs only the pubkey set.
- Server-side revocation is handled by shipping an updated pinned set on key rotation, plus the existing account-status lifecycle for compromise — not by live-resolving NIP-05 mid-sign.

This mirrors the pinned set into a **third** location (mobile `official_accounts.dart`, web `officialAccounts.ts`, and now keycast `core/src/verified_minor_dm.rs`), maintained by hand. A shared source is a future cleanup, not in scope here; the values are kept identical (verified 2026-07-09).

## Fail-closed behavior

This is a child-safety gate, so every ambiguous case refuses: no `p` tags on a kind-4, a malformed/unparseable `p` tag, a seal whose content decrypts under no approved key, a seal whose plaintext is not event-shaped JSON, an unverifiable DM-carrier kind (14/15/1059), an unresolvable minor status, or a missing user row. Refusals surface a uniform `Operation denied by policy` message (identical to a policy denial, so account state is not leaked through error text); the specific reason is logged server-side only.

## Where it is wired

The `verified_minor` flag is read from the same per-request row as the existing account-status check on every surface, so a protection flip applies immediately despite handler caching (the BLAKE3 HTTP handler cache still runs a fresh status query per request).

- **HTTP RPC `POST /api/nostr`** (`api/src/api/http/nostr_rpc.rs`) — the primary open headless path. `sign_event`, `nip04_encrypt`, `nip44_encrypt` gated.
- **`POST /api/user/sign`** (`api/src/api/http/auth.rs`) — the minor flag is resolved once and both the fast path (cached signer handler) and the slow path (DB + decrypt) gate through a shared helper, each returning a clean 403.
- **NIP-46 relay signer** (`signer/src/signer_daemon.rs`) — a shared `check_user_sign_gates` runs on both relay-dispatched (`handle_sign_event`) and direct (`sign_event_direct`) signing; the encrypt gate lives in `validate_permissions_for_encrypt`, which every `nip04`/`nip44` encrypt dispatch already passes through.

## Review

Reviewed by two adversarial subagents (a senior code review and a bypass-hunt scoped to the acceptance criteria, fail-closed, recipient-identifiability at each shape, and any signing path that skips the gate). Two findings were confirmed against source and fixed in this branch:

1. **Spoofable `p`-tag binding for kinds 14/15/1059 (high):** the symmetric-ECDH property let a colluding recipient hide a minor-readable payload behind a decoy approved `p` tag. Fixed by refusing those kinds outright (they are never legitimately signed via a remote signer) and keeping the real binding on the seal (trial-decrypt) and the encrypt primitives.
2. **Fast-path error surface (important):** on the hot `/user/sign` fast path, a minor-gate denial mapped through the signer handler to a 503 with a retry message instead of a 403. Fixed by gating both paths in the handler before signing, with a fast-path (`signer_handlers: Some`) test asserting the clean 403.

Also folded in: kind-13 seal `p` tags are now validated (defense-in-depth), and the gated-kind list is pinned to the validator by an anti-drift test.

## Out of scope (noted per #183)

The headless client-identity weakness (self-reported `client_id` / `redirect_uri`) is separate hardening called out in #183; this PR does not expand into it. I did trace whether it's currently exploitable: the two capability decisions that consume client identity are each gated behind a stronger factor, so the weakness is latent, not live. Account deletion (`auth.rs`) requires either a user-signed UCAN or the `first_party` fact, and `first_party` is set only for headless-flow tokens, which require the account's email/password — a claimed `client_id` grants nothing without the credentials. The preloaded-user signing path (`nostr_rpc.rs`) additionally requires a server-signed UCAN, which a caller cannot forge. It would only become exploitable if a future control trusted `client_id` on its own. It also does not touch this gate: the minor DM check reads `verified_minor` from the account row per request and never consults the caller's claimed client identity, so it holds against any client, official or forged.

The `verified_minor` gate contains DM *egress via the signer*. It does not by itself close raw-key *egress*: `POST /user/export-key` (returns the nsec) and `POST /user/change-key` (swaps to a caller-supplied key) are password-authenticated with no `verified_minor` gate today, so a custodial minor could export their key via the headless API and then bypass this gate entirely from any client. That is the server-side complement to #182's app-side affordance removal and is tracked separately — this PR does not address it.

## Testing

TDD across three cycles (core unit, API integration, signer daemon), each written red-first and watched fail before implementing:

- `core/src/verified_minor_dm.rs` — 24 unit tests (kind classification + anti-drift, encrypt gate, kind-4 p-tag gating, kind-13 seal trial-decrypt + outer/inner p-tags, 14/15/1059 refused, the decoy-p-tag bypass proven closed, every fail-closed refusal).
- `api/tests/verified_minor_dm_gate_test.rs` — 18 integration tests over `/api/nostr` and `/user/sign` (refusals, allow to officials + self, non-minor unaffected, seal end-to-end, ingress-still-open, fail-closed on missing user row, and the fast-path 403).
- `signer/tests/verified_minor_dm_daemon_test.rs` — 10 tests over the NIP-46 signer.

Full workspace suite, `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`, and `cargo fmt --all -- --check` all clean.

- [x] `cargo test --workspace` (+ integration-feature suites for core/api/signer)

## Visuals

- [x] No visual change (backend only)

## Related Issue

- Closes divinevideo/support-trust-safety#183

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -3208,6 +3208,28 @@ pub struct PubkeyResponse {
     pub npub: String,   // bech32 format
 }
 
+/// Apply the verified_minor DM containment gate (support-trust-safety#183) to a
+/// `/user/sign` request, on either the fast or slow path. Refusals surface the
+/// uniform policy-denial message (no account-state leak); the specific reason is
+/// logged server-side only. `keys` is the user's signing keypair (needed to
+/// recover a kind-13 seal's recipient by trial decryption).
+fn enforce_minor_dm_sign(
+    keys: &Keys,
+    unsigned_event: &UnsignedEvent,
+    user_pubkey: &str,
+) -> Result<(), AuthError> {
+    keycast_core::verified_minor_dm::validate_minor_sign(keys, unsigned_event).map_err(|denied| {
+        tracing::warn!(
+            event = "minor_dm_gate.sign_denied",
+            user_pubkey = %user_pubkey,
+            kind = unsigned_event.kind.as_u16(),
+            reason = %denied,
+            "verified_minor DM sign refused (/user/sign)"
+        );
+        AuthError::Forbidden("Operation denied by policy".to_string())
+    })
+}
+
 /// Fast HTTP signing endpoint - sign an event without NIP-46 relay overhead
 /// This is 10-50x faster than NIP-46 for quick operations
 pub async fn sign_event(
@@ -3244,6 +3266,21 @@ pub async fn sign_event(
     )
     .await?;
 
+    // verified_minor DM containment (support-trust-safety#183) applies to BOTH
+    // the fast and slow paths. Fetch the flag once here (only for DM-shaped
+    // kinds) so the hot fast path returns a clean 403 too, rather than the
+    // signer handler's error mapping to a 503. A missing user row fails closed.
+    let verified_minor =
+        if keycast_core::verified_minor_dm::is_minor_gated_kind(unsigned_event.kind) {
+            user_repo
+                .get_verified_minor(&user_pubkey, tenant_id)
+                .await?
+                .ok_or_else(|| AuthError::Forbidden("Operation denied by policy".to_string()))?
+                .0
+        } else {
+            false
+        };
+
     // FAST PATH: Try to use cached signer handler if in unified mode
     if let Some(ref handlers) = auth_state.state.signer_handlers {
         tracing::info!(
@@ -3261,6 +3298,13 @@ pub async fn sign_event(
         if let Some(bunker_key) = bunker_pubkey {
             if let Some(handler) = handlers.get(&bunker_key).await {
                 tracing::info!("✅ Using cached handler for user {}", user_pubkey);
+
+                // DM containment on the fast path returns a clean 403 here; the
+                // signer handler also gates as a backstop, but its error would
+                // map to a 503 through the Internal wrapper below.
+                if verified_minor {
+                    enforce_minor_dm_sign(&handler.get_keys(), &unsigned_event, &user_pubkey)?;
+                }
 
                 let signed_event = handler
                     .sign_event_direct(unsigned_event)
@@ -3307,33 +3351,10 @@ pub async fn sign_event(
         .map_err(|e| AuthError::Internal(format!("Invalid secret key bytes: {}", e)))?;
     let keys = Keys::new(secret_key.into());
 
-    // verified_minor DM containment (support-trust-safety#183). The fast path
-    // above enforces this inside the signer handler; this covers the slow path.
-    // Gated here (after key decryption) because a kind-13 seal's recipient can
-    // only be recovered with the user's conversation keys.
-    if keycast_core::verified_minor_dm::is_minor_gated_kind(unsigned_event.kind) {
-        let verified_minor = user_repo
-            .get_verified_minor(&user_pubkey, tenant_id)
-            .await?
-            // Fail closed: DM-shaped signing with unresolvable minor status
-            // is refused (unreachable in practice — personal_keys FK-cascades
-            // with users — but the gate must not default open).
-            .ok_or_else(|| AuthError::Forbidden("Operation denied by policy".to_string()))?
-            .0;
-        if verified_minor {
-            keycast_core::verified_minor_dm::validate_minor_sign(&keys, &unsigned_event).map_err(
-                |denied| {
-                    tracing::warn!(
-                        event = "minor_dm_gate.sign_denied",
-                        user_pubkey = %user_pubkey,
-                        kind = unsigned_event.kind.as_u16(),
-                        reason = %denied,
-                        "verified_minor DM sign refused (/user/sign slow path)"
-                    );
-                    AuthError::Forbidden("Operation denied by policy".to_string())
-                },
-            )?;
-        }
+    // verified_minor DM containment on the slow path (support-trust-safety#183);
+    // `verified_minor` was resolved once above and shared with the fast path.
+    if verified_minor {
+        enforce_minor_dm_sign(&keys, &unsigned_event, &user_pubkey)?;
     }
 
     // Permission validation already done above (before fast path check)
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -3318,13 +3318,11 @@ pub async fn sign_event(
             // Fail closed: DM-shaped signing with unresolvable minor status
             // is refused (unreachable in practice — personal_keys FK-cascades
             // with users — but the gate must not default open).
-            .ok_or_else(|| {
-                AuthError::Forbidden("Operation denied by policy".to_string())
-            })?
+            .ok_or_else(|| AuthError::Forbidden("Operation denied by policy".to_string()))?
             .0;
         if verified_minor {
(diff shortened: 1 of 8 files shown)


dcadenas (changes requested):
The core containment holds on all three signing surfaces, but the NIP-46 relay path denies a minor's DM without telling the client. The HTTP paths return a clean 403, while a denial on the relay dispatcher returns early and never sends a reply, so the client times out with no explanation and likely retries. Please make the relay path emit a proper refusal and add a test that drives the full request handler before this merges.

---

- **Relay NIP-46 minor-gate denials time out instead of returning a JSON-RPC error** ([signer/src/signer_daemon.rs:1483](https://github.com/divinevideo/keycast/blob/127474060c6009c514b9ce613bb60394d564bf34/signer/src/signer_daemon.rs#L1483))
  The relay path denies the DM correctly, but the client never hears why. When `handle_sign_event` here (via `check_user_sign_gates`) or `validate_permissions_for_encrypt` (:1524 for nip44, :1600 for nip04) returns a denial, `?` exits `handle_nip46_request` before the encrypted `{id, error}` response is built at :1667-1719. So a minor's NIP-46 client asking to sign or encrypt a DM to a non-approved recipient gets a relay timeout, not the clean refusal the HTTP paths return as a 403.
  
  Send a JSON-RPC error on these denials the same way the tombstone branch already does:
  
  https://github.com/divinevideo/keycast/blob/127474060c6009c514b9ce613bb60394d564bf34/signer/src/signer_daemon.rs#L1376-L1414
  
  Convert the expected per-request denials (permission denied, invalid params or key) into an encrypted `{id, error}` reply through that response path instead of `?`-propagating out of the handler. This early return is pre-existing dispatcher behavior, so the same fix also covers the account-suspended and unauthorized-client denials that exit the same way today. What this PR changes is that a new, recipient-conditioned denial now fires routinely for minors, which makes the missing feedback a routine experience rather than an edge case.
  
  Add a test that drives `handle_nip46_request` for a denied minor sign or encrypt and asserts an error response event is published. The current daemon tests exercise `sign_event_direct` and `validate_permissions_for_encrypt` directly and never drive the full handler, so this relay reply path ships untested.

mbradley:
Oh. Yep: this framing is the right correction: once the minor gate denies routinely, that timeout stops being an edge case and becomes the normal experience for a whole class of accounts.

Fixed: the relay path now returns an `{id, error}` reply for expected per-request denials (permission-denied / bad params / bad key) instead of the silent drop; internal failures still propagate and log. I pulled the reply-path core into `build_nip46_response_event` so the wire handler stays thin and the path is finally testable. As you said, this also covers the pre-existing suspended and policy denials that exited the same way.

On the test: driving `handle_nip46_request` literally needs a live relay plus a Redis coordinator, and there's no mock relay at nostr-sdk 0.44, so I test the extracted seam directly (denied minor sign + encrypt each decrypt to `{id, error}`; happy path `{id, result}`; a unit test pins the classification). Glad to add a live-relay integration test on top if you'd rather have it.

dcadenas (approved):
🛡️ minor-DM containment holds across the sign, encrypt, and NIP-46 paths, ship it.

PR #288, opened 2026-07-10
feat(minor-safety): server-side verified_minor gate on key export + change-key

mbradley:
## Summary

- Server-side `verified_minor` gate on `POST /user/export-key` and `POST /user/change-key`, closing the raw-key egress path for custodial minors.
- The server-side complement to the app-side affordance removal (#182), and the key-egress twin of the DM containment (#183, keycast#284): #182 hides the export/change-key affordances in the apps; this refuses them at Keycast so the key can't be extracted over the headless API.

## Motivation

Keycast holds the signing key for a custodial minor. Both endpoints are password-authenticated with no minor gate today: `export_key` decrypts and returns the raw nsec, `change_key` swaps the account to a caller-supplied key. They sit on first-party CORS, but CORS is a browser control only, so a script with the account's own Keycast password can call either directly over the headless API. A custodial minor knows their password (they log in with it), so they could export their nsec, load it into any client, and bypass every other guardrail — including the #183/#284 DM containment. That collapses the epic's custodial-containment premise into the accepted self-custody ceiling. This PR removes that escape hatch.

## What it does

- Both handlers call a shared `refuse_key_egress_for_verified_minor` helper at the top, **before the password check and before any key material is touched**, so the least code runs between authentication and the refusal.
- **Fail closed:** only an explicit non-minor status proceeds; `verified_minor` OR an unresolvable account (no user row) is refused.
- **Uniform refusal:** `AuthError::Forbidden("Operation denied by policy")` — identical to a policy denial, so it leaks no account state; the specific reason is logged server-side only.
- **Auto-lifts:** the block clears the moment `verified_minor` is cleared (age-up / revocation via the existing lifecycle), the same flag that drives the DM gate. No special handoff path; a guardian/admin key handoff for a still-minor account, if ever needed, is future admin/service-token work.
- Non-minor accounts are unaffected.

## Design note

Gated at the top of each handler rather than after password verification, deliberately: for a child-safety key-egress gate, fail-fast (refuse before any password work or key decrypt/swap) minimizes the code path to the key material, and it makes the fail-closed-on-unresolvable case real and testable. The tradeoff is that a caller holding a valid token but not the password could infer minor status from the uniform refusal; that's low-sensitivity (they already hold account-level auth) and not a bypass.

## Testing

TDD, red-first. New `api/tests/verified_minor_key_egress_test.rs` (6 integration tests): verified_minor refused on export-key (including before the password check, on a wrong password), non-minor export returns an nsec, fail-closed on a missing user row, verified_minor refused on change-key with the original identity left intact, non-minor change-key succeeds. Existing `change_key_test` and `email_change_test` suites still green (no regression).

- [x] `cargo test -p keycast_api --features integration-tests --test verified_minor_key_egress_test` (6 pass) + change_key/email_change regression

## Related Issue

- Closes divinevideo/support-trust-safety#188

## Visuals

- [x] No visual change (backend only)

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -3841,6 +3841,45 @@ pub struct ChangeKeyResponse {
     pub message: String,
 }
 
+/// Refuse a raw-key egress operation (export-key / change-key) for a
+/// `verified_minor` account (support-trust-safety#188).
+///
+/// This is the server-side complement to the app-side affordance removal
+/// (#182): Keycast holds the key for a custodial minor, so an ungated
+/// export/change-key over the headless API would let the minor extract or swap
+/// their key with just their password and bypass every other guardrail. Called
+/// at the top of the handler, before the password check and before any key
+/// material is touched, so the least code runs between authentication and the
+/// refusal. The block lifts automatically when `verified_minor` is cleared
+/// (age-up / revocation) — the same flag that drives the DM gate.
+///
+/// Fail closed: only an explicit non-minor status proceeds; `verified_minor` OR
+/// an unresolvable account (no row) is refused. The uniform policy-denial
+/// message leaks no account state; the specific reason is logged server-side.
+async fn refuse_key_egress_for_verified_minor(
+    user_repo: &UserRepository,
+    tenant_id: i64,
+    user_pubkey: &str,
+    operation: &str,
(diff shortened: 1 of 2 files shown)


dcadenas (approved):
🔒 Fail-closed minor gate runs before the password check or any key material, clean placement.

PR #293, opened 2026-07-12
fix(auth): complete OAuth verify_email row-state retries

### LINKED ISSUE #264: OAuth verify_email conflates any existing users row with a completed registration (idempotency-by-existence)
## Summary

In `verify_email` (`api/src/api/http/auth.rs`), the OAuth/`oauth_codes` pending-registration path uses `UserRepository::exists(pubkey, tenant_id)` as its idempotency signal. `exists` is a pure existence check (`SELECT pubkey FROM users WHERE pubkey = $1 AND tenant_id = $2`) — it does not check `email`, `email_verified`, or whether `personal_keys` exist.

The `users` table is **shared** across many creation paths, several of which insert a bare row (`pubkey, tenant_id` only, no email / password / personal key), e.g.:

- team creation / membership (`core/src/repositories/team.rs` — `create_with_admin`, `add_member`)
- OAuth authorization pre-creation (`core/src/repositories/oauth_authorization.rs`)
- `personal_keys` / `oauth_code` repositories' `ON CONFLICT (pubkey) DO NOTHING` upserts
- vine_id / minor-verification user inserts (`core/src/repositories/user.rs`)

When any such row already exists for the pending pubkey, `user_already_exists` is `true`, so the retry branch only logs `"Retrying OAuth email verification for existing user"` and falls through — **it does not apply the pending email/password and does not create `personal_keys`** — yet it still proceeds to mint the OAuth code and delete the pending verification token. The registration silently no-ops while reporting success.

## Concrete failure scenario

1. Pubkey `P` gets a bare `users` row via a non-OAuth path (e.g. added as a team member): `email = NULL`, no `personal_keys`.
2. `P` later self-registers through the OAuth email/password flow with the **same** key (realistically a BYOK / imported-key registration), pending email `E`.
3. `verify_email`:
   - `find_pubkey_by_email(E)` → `None` (the existing row has a null email) → proceeds
   - `exists(P)` → `true` → retry branch → `E` / password never applied, `personal_keys` never created
   - OAuth code minted, pending verification token deleted
4. Result: `P` still has no email/password and no personal signing key, but the registration "succeeded" and the one-time token is consumed. The user cannot log in with email/password and has no personal key, with no error surfaced.

## Why this matters

The cohort most likely to hit this is exactly the set of pubkeys that exist for a non-OAuth reason **and** later authenticate personally — i.e. team members who also use personal login. Team functionality is lightly used today, so the live blast radius is currently near-zero, but that is a property of current usage, not of the code being safe: the at-risk population grows directly with team adoption.

## Reachability (production, read-only check)

- The trigger condition is already common: a large share of pending `oauth_codes` are for pubkeys that already have a `users` row, so the retry branch is exercised frequently (the large majority are benign same-user re-verifications).
- Harmful subset today is near-zero: **no** generated/import-key pending registration is currently missing its `personal_keys`; only a small, mostly-expired set of pending codes show a pending-email that differs from the existing row's email (and adding email to an existing user appears to route through the separate UPDATE path at `verify_email`/`user_repo.verify_email`, not this INSERT-retry branch).

Conclusion: **latent, not active** — but should be fixed before team usage scales.

## Proposed fix

Stop using bare `exists(pubkey)` as the finalization gate. Before treating an existing user as an idempotent retry, require that the existing row:

- owns the pending email (`find_pubkey_by_email(pending_email) == oauth_data.user_pubkey`), and
- for generated/import-key registrations, has a `personal_keys` row whose material matches the pending encrypted secret.

If those invariants do not hold: either apply the pending email/key material to the existing row, or return a conflict (`EmailAlreadyExists` / `409`) — but never silently mint the OAuth code and delete the pending verification token.

> Note: an earlier set of review cycles on PR #236 prototyped a finalization-gate approach (plus a `personal_keys (tenant_id, user_pubkey)` uniqueness migration) for this, but it was not landed. Reviewer feedback on that attempt flagged that any DB-level uniqueness migration must dedup **non-destructively** (do not `DELETE` encrypted key rows) and build the index with `CREATE INDEX CONCURRENTLY`.

## Scope

Separate from #198 / PR #236, which is correctly narrowed to adding the `EMAIL_ALREADY_EXISTS` machine-readable code to the duplicate-email 409.

## Acceptance criteria

- The OAuth pending-registration finalization in `verify_email` does not finalize (mint OAuth code / delete pending token) unless the existing user row reflects the pending email and required key material.
- A pre-existing bare `users` row (team/import/pre-created) for the pending pubkey results in the pending email/key being applied, or a conflict — never a silent success that drops the registration.
- Regression coverage for: bare-row-then-OAuth finalization, generated-key material match, and email-ownership mismatch.

dcadenas:
## Summary

OAuth email verification now distinguishes a completed retry from an incomplete same-email users row.
Same-email rows that are still unverified or missing a password hash are completed before an OAuth code is minted.
Personal-key repair inserts are backed by a unique `personal_keys(user_pubkey)` index and use `ON CONFLICT DO NOTHING`, so concurrent retries cannot create duplicate key rows.

## Motivation

Some server paths create bare or incomplete `users` rows before OAuth email verification finishes.
The old idempotency check treated any matching email as completed registration, which could report success while leaving the row unverified or incomplete.
The personal-key repair path also relied on a non-atomic `NOT EXISTS` guard without a database uniqueness constraint.

## Related Issue

- Closes #264

## Testing

- [x] `cargo test --workspace`
- [x] `cargo test -p keycast_core --features integration-tests oauth_registration_state`
- [x] `cargo test -p keycast_core --features integration-tests complete_pending_oauth_registration`
- [x] `cargo test -p keycast_core --features integration-tests backfill_personal_key_is_idempotent`
- [x] `cargo test -p keycast_api --features integration-tests verify_email_completes_same_email_unverified_row_before_minting`
- [x] `DATABASE_URL=postgres://postgres:password@127.0.0.1:5432/keycast_test REDIS_URL=redis://127.0.0.1:16379 TEST_REDIS_URL=redis://127.0.0.1:16379 cargo test -p keycast_core --features integration-tests --lib -- --test-threads=1`
- [x] `DATABASE_URL=postgres://postgres:password@127.0.0.1:5432/keycast_test REDIS_URL=redis://127.0.0.1:16379 TEST_REDIS_URL=redis://127.0.0.1:16379 cargo test -p keycast_api --features integration-tests --lib --tests -- --test-threads=1`
- [ ] Manual verification completed (no UI change; covered by integration tests)

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -1581,13 +1581,79 @@ pub async fn verify_email(
 
         // Create user with email_verified=true (they just verified!)
         let user_repo = UserRepository::new(pool.clone());
-        let user_already_exists = user_repo.exists(&oauth_data.user_pubkey, tenant_id).await?;
+        let existing_user = user_repo
+            .oauth_registration_state(&oauth_data.user_pubkey, tenant_id)
+            .await?;
 
-        if user_already_exists {
-            tracing::info!(
-                "Retrying OAuth email verification for existing user: {}",
-                oauth_data.user_pubkey
-            );
+        if let Some(existing) = existing_user {
+            match existing.email.as_deref() {
+                Some(existing_email) if existing_email == email => {
+                    // Genuine retry: a prior attempt already applied the pending
+                    // credentials. Backfill the personal key if that attempt (or
+                    // another creation path) left it missing.
+                    tracing::info!(
+                        "Retrying OAuth email verification for existing user: {}",
+                        oauth_data.user_pubkey
+                    );
+                    if oauth_data.pending_encrypted_secret.is_some() && !existing.has_personal_key {
+                        user_repo
+                            .complete_pending_oauth_registration(
+                                &oauth_data.user_pubkey,
+                                tenant_id,
+                                email,
+                                password_hash,
+                                &req.token,
+                                oauth_data.pending_encrypted_secret.as_deref(),
+                            )
+                            .await?;
+                        tracing::info!(
+                            "Backfilled missing personal key during OAuth verification retry: {}",
+                            oauth_data.user_pubkey
+                        );
+                    }
+                }
+                Some(_) => {
+                    // The pubkey is already bound to a different email. Never
+                    // overwrite an existing account's credentials from an
+                    // unauthenticated verification link.
+                    tracing::warn!(
+                        "OAuth email verification for pubkey {} conflicts with an existing account email",
+                        oauth_data.user_pubkey
+                    );
+                    oauth_code_repo
+                        .delete_by_verification_token(&req.token, tenant_id)
+                        .await?;
+                    return Err(AuthError::DuplicateKey);
+                }
+                None => {
+                    // Bare row pre-created by another path (team membership,
+                    // authorization pre-creation): apply the pending registration
+                    // instead of silently dropping it.
+                    if let Err(err) = user_repo
+                        .complete_pending_oauth_registration(
+                            &oauth_data.user_pubkey,
+                            tenant_id,
+                            email,
+                            password_hash,
+                            &req.token,
+                            oauth_data.pending_encrypted_secret.as_deref(),
+                        )
+                        .await
+                    {
+                        if matches!(err, keycast_core::repositories::RepositoryError::Duplicate) {
+                            oauth_code_repo
+                                .delete_by_verification_token(&req.token, tenant_id)
+                                .await?;
+                            return Err(AuthError::EmailAlreadyExists);
+                        }
+                        return Err(err.into());
+                    }
+                    tracing::info!(
+                        "Applied pending OAuth registration to pre-existing user row: {}",
+                        oauth_data.user_pubkey
+                    );
+                }
+            }
         } else if let Some(ref encrypted_secret) = oauth_data.pending_encrypted_secret {
             // Auto-generated or direct nsec: create user + personal_keys
             // Use a transaction to ensure atomicity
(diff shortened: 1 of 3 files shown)


mbradley (approved):
solid fix. the row-state gate closes the silent-drop path from #264 cleanly: bare rows get the pending registration applied, genuine retries only backfill a missing key without touching credentials, and conflicts 409 before minting anything. good regression coverage on all three branches, and the update-gated-on-email-is-null approach means the state read can't be raced into a credential overwrite.

one small thing: title says chore but this is a behavioral fix, fix(auth) reads better for the changelog (semantic check should rerun after a retitle).

mbradley:
pushed two small additions from my review: a fix making complete_pending_oauth_registration idempotent when a concurrent duplicate verify wins the race (plus a regression test pinning it), and a whitespace restore. both green locally against the full verify_email + repo test set.

api/src/api/http/auth.rs
```diff
@@ -1580,13 +1580,79 @@ pub async fn verify_email(
...
+            .await?;
 
-        if user_already_exists {
-            tracing::info!(
-                "Retrying OAuth email verification for existing user: {}",
-                oauth_data.user_pubkey
-            );
+        if let Some(existing) = existing_user {
+            match existing.email.as_deref() {
+                Some(existing_email) if existing_email == email => {
```
NotThatKindOfDrLiz, line 1590: This treats matching email alone as proof that the pending OAuth registration was already completed. That is not always true: normal BYOK registration can create the same pubkey/email row with `email_verified = false` and, while bcrypt is pending or failed, `password_hash = NULL`. This branch then mints the OAuth code and deletes the pending verification token, but leaves the users row unverified/incomplete, so later `/login` can still reject the account after the UI reported "Email verified!".

Please include the completion fields in the state read and make this branch explicitly complete the row. At minimum, the OAuth verification token should set `email_verified = true`; password handling should be deliberate too, e.g. preserve an existing hash but fill from the pending hash when the row has NULL.

core/src/repositories/user.rs
```diff
@@ -1315,6 +1328,136 @@ impl UserRepository {
...
+    ///
+    /// Returns [`RepositoryError::Database`] if the insert query fails.
+    pub async fn backfill_personal_key(
+        &self,
+        pubkey: &str,
+        tenant_id: i64,
+        encrypted_secret: &[u8],
+    ) -> Result<(), RepositoryError> {
+        let now = Utc::now();
+
```
NotThatKindOfDrLiz, line 1403: The `INSERT ... SELECT ... WHERE NOT EXISTS` guard is not atomic across concurrent transactions, and `personal_keys` has no unique constraint on `user_pubkey` or `(tenant_id, user_pubkey)`. The bare-row completion path is serialized by the users-row update, but this retry/backfill path is not, so two simultaneous verification clicks can both observe no key and insert duplicate `personal_keys` rows.

Please either enforce the invariant in the database, ideally with a non-destructive dedupe plus a unique index on `(tenant_id, user_pubkey)`, or add a transaction-level lock around this path and document why the schema invariant remains out of scope.

NotThatKindOfDrLiz (changes requested):
Requesting changes on two edge cases in the new idempotency logic. The row-state gate is the right shape, but the matching-email retry path can still report success without completing the user row, and the personal-key backfill still relies on a non-atomic NOT EXISTS guard without a database uniqueness constraint.

NotThatKindOfDrLiz:
Takeover update:

- Completed same-email but incomplete OAuth verification rows before minting an OAuth code.
- Added `email_verified` and password-presence state to the OAuth registration snapshot.
- Added a migration to dedupe/enforce one `personal_keys` row per `user_pubkey`.
- Switched personal-key repair inserts to `ON CONFLICT (user_pubkey) DO NOTHING`.
- Added regression coverage for same-email unverified rows and idempotent personal-key backfill.

Verification:

- `cargo test --workspace`
- targeted OAuth/user repository tests
- serial `keycast_core` and `keycast_api` integration-feature suites against local Postgres/Redis
- GitHub checks are green.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover fixes. The same-email incomplete-row path now completes before minting, personal-key repair is backed by a database uniqueness invariant plus ON CONFLICT, regression coverage was added, and CI is green.

NotThatKindOfDrLiz:
@dcadenas final re-review is clean: no blocking findings, approvals and checks are green, and the branch is current with `main`. Please merge when ready.

PR #294, opened 2026-07-13
refactor(user-repo): consolidate account-status tuple queries onto a FromRow struct

### LINKED ISSUE #267: refactor(user-repo): consolidate account-status tuple queries onto a FromRow struct
Surfaced during the keycast#263 review (#266). `core/src/repositories/user.rs` now has four overlapping `SELECT ... FROM users` methods returning positional tuples of differing arity: `get_account_status` (4), `get_account_status_with_minor` (6), `get_verified_minor` (2), `get_full_admin_status` (5). The 6-element positional tuple is error-prone to destructure and easy to mis-order.

Consider a `#[derive(sqlx::FromRow)]` row struct (e.g. `AccountStatusRow`) so these variants become field selections rather than arity-N tuples, and callers read named fields.

Non-blocking cleanup; adding the new method in #266 was the correct call given the constraint not to touch the shared `get_account_status`. Relates to the protected-minor safeguards epic (divinevideo/support-trust-safety#173).

mbradley:
## Summary

- Replace the four positional-tuple `SELECT ... FROM users` methods in `core/src/repositories/user.rs` (`get_account_status`, `get_account_status_with_minor`, `get_verified_minor`, `get_full_admin_status`) with per-method `#[derive(sqlx::FromRow)]` row structs, so every call site reads named fields instead of `.0`/`.1` or a six-element destructure.

## Motivation

- The six-element positional tuple from `get_account_status_with_minor` is easy to mis-order, which surfaced during the keycast#263 review (#266). These are the protected-minor / account-status safeguard paths, so a silent field swap here is a correctness-and-safety hazard. `FromRow` maps by column name, so the shape can no longer be destructured in the wrong order.
- Design tradeoff: I used four per-method structs rather than one struct that always selects the full column set. The single-struct option is a bit less boilerplate, but it would rewrite the `SELECT` list of all four queries (including the oauth login/session hot-path `get_account_status`, which is deliberately kept minimal) and would need `#[sqlx(default)]` on absent columns, silently fabricating defaults. Per-method structs keep every query byte-for-byte identical and never carry a column the query didn't select. The cost is four small structs with some overlapping fields, which stays local to `user.rs`.
- Pure refactor: no query text, binds, or returned values change, only the shape callers read.

## Related Issue

- Closes #267
- Part of divinevideo/support-trust-safety#173

## Testing

- [x] Full core + api integration suites (CI parity, Postgres+Redis via `docker-compose.deps.yml`): core `151` tests, api `~330` across ~40 binaries, `0` failures. Directly covers `account_status_minor_test`, `verified_minor_dm_gate_test`, `verified_minor_key_egress_test`, the oauth/session paths, and the `core::repositories::user` tests exercising `get_verified_minor` / `get_full_admin_status`.
- [x] Compiler backstop: retargeting the return types makes any unconverted call site fail to build, so a clean `--workspace` build proves all call sites were migrated.

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -15,8 +15,8 @@ use crate::api::error::{ApiError, ApiResult};
 use crate::api::extractors::UcanAuth;
 use keycast_core::repositories::{
     test_redirect_pattern, AdminAuditEventRecord, AdminAuditEventRepository, AuthEventRepository,
-    ClaimTokenRepository, OAuthAuthorizationRepository, RegisteredClient,
-    RegisteredClientRepository, RepositoryError, UserRepository,
+    ClaimTokenRepository, FullAdminStatusRow, OAuthAuthorizationRepository, RegisteredClient,
+    RegisteredClientRepository, RepositoryError, UserRepository, VerifiedMinorRow,
 };
 use keycast_core::types::claim_token::generate_claim_token;
 use keycast_core::types::user::UserStatus;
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
✅ clean account-status refactor, ship it

PR #301, opened 2026-07-18
chore: Add fuzzy + contains email search to support-admin user lookup

### LINKED ISSUE #300: Add fuzzy + contains email search to support-admin user lookup
Support agents can only find an account by email if they type the full address exactly right, so a single typo makes the account impossible to find.
This has already blocked real tickets, including one where the person's real account used a one-letter-off email and no partial search could surface it.
Add email search that matches partial addresses, matches anywhere in the address, and suggests close matches ("did you mean…") when there's no exact hit.
The search should stay fast as the user base grows.

dcadenas:
Support admins can now find accounts from partial email addresses and recover from small email typos. Exact identity matches stay first, including case-variant rows that normalize to the same email.

## Summary

Email fragments now return ordered account matches, and close misspellings return ranked "Did you mean?" choices.

- Added a partial pg_trgm GIN index for substring and similarity lookup.
- Reused the existing expandable account cards for suggestions.
- Preserved all exact email, username, name, npub, and pubkey matches as the authoritative tier.

## Motivation

The old exact-only email path ended in a dead end for fragments and typos. The lookup also treated any loose row as authoritative, which could hide better identity resolution.

- Candidate tiers are explicit, deterministic through hydration, and balanced under the 20-result cap.
- Missing trigram operators disable suggestions without hiding unrelated database errors.
- Three local INSERT trials against a 400,000-row baseline measured a median paired throughput change of -1.96% (range -0.68% to -4.20%). Local EXPLAINs showed the new index serving selective populated substring/similarity cases, but low-selectivity similarity queries can still scan and sort more broadly; the 100-candidate cap bounds rows handed to the Rust edit-distance filter, not every row PostgreSQL may inspect.

## Related Issue

- Closes #300

## Deployment notes

- Production does not have `pg_trgm` installed yet, but PostgreSQL reports version 1.6 as available and trusted. The migration role has `CREATE` on the database and `public` schema, so the migration can install it without a separate DBA step.
- Staging uses CloudNativePG rather than Cloud SQL. Its migration job connects as the `keycast` database owner after a superuser bootstrap enforces database ownership and schema privileges. The configured CloudNativePG 1.27 default PostgreSQL 17.5 image contains `pg_trgm` 1.6 marked trusted.
- Production currently has 216,549 users, including 41,122 rows with email. The measured 400,000-row workload is larger than the current email-bearing population.
- Both index statements are non-concurrent and SQLx runs the migration transactionally, so writes are blocked during the build. A 400,000-row local clone completed the two-index sequence in 1,903 ms, but that is a lower bound rather than a production latency guarantee; monitor the migration job and database write health during rollout.

## Testing

The database, repository, API, and web paths were tested locally.

- [x] Database-backed lookup tests with `integration-tests`, including case-variant normalized emails
- [x] `cd web && bun test`
- [x] 400,000-row INSERT and EXPLAIN verification

## Visuals

The 1440x1000 suggestion state was verified manually. The fuzzy row is explicitly labeled and remains collapsed until selected.

support-admin-email-suggestion

- [x] UI/web change with screenshot attached
- [ ] No visual change


core/src/repositories/user.rs
```diff
@@ -1750,45 +1822,117 @@ impl UserRepository {
...
+        let normalized_query = query.to_lowercase();
+        let (authoritative_pubkeys, primary_loose_pubkeys, email_loose_pubkeys) = if query
+            .contains('@')
+        {
+            if query.len() < 3 {
+                (vec![], vec![], vec![])
+            } else {
+                let exact_email: Vec<(String,)> = sqlx::query_as(
+                    "SELECT pubkey FROM users
+                     WHERE LOWER(email) = $1 AND tenant_id = $2
```
NotThatKindOfDrLiz: This exact-email path used to be `email = $1`, which could use the existing unique btree index on `(tenant_id, email)`. Switching to `LOWER(email) = $1` makes the lookup defensive for legacy mixed-case rows, but there is no matching functional index, and the new trigram GIN index does not serve equality. Please either add an index such as `(tenant_id, LOWER(email)) WHERE email IS NOT NULL` in this migration, or include an `EXPLAIN`/benchmark justification showing the scan is acceptable on a large tenant.

api/src/api/http/admin.rs
```diff
@@ -930,55 +1017,58 @@ pub async fn get_user_lookup(
...
+                        .into_iter()
+                        .filter(|user| seen_pubkeys.insert(user.pubkey.clone())),
+                );
+                resolved_lookup.users.truncate(ADMIN_USER_LOOKUP_LIMIT);
+                lookup = resolved_lookup;
+            }
         }
     }
 
+    let suggested_users = if should_fetch_email_suggestions(lookup.authoritative_match) {
```
NotThatKindOfDrLiz: This fetches suggestions whenever there is no authoritative match, but the frontend only renders suggestions when `results` is empty. For loose matches, this runs the trigram query and then enriches up to 5 suggestions that are never displayed. Please gate this on the merged result set being empty as well, e.g. `!lookup.authoritative_match && lookup.users.is_empty()` or an equivalent helper.

NotThatKindOfDrLiz (changes requested):
Requesting changes for two localized issues: keep the exact-email path indexed after switching to case-insensitive matching, and avoid fetching suggestions when the UI will discard them.

core/src/repositories/user.rs
dcadenas: Addressed in c535918: the migration now creates `idx_users_tenant_lower_email` on `(tenant_id, LOWER(email)) WHERE email IS NOT NULL`, and the exact query includes the partial-index predicate explicitly. A migration-backed regression test verifies the index definition, and local `EXPLAIN` selected the new index.

api/src/api/http/admin.rs
dcadenas: Addressed in c535918: `should_fetch_email_suggestions` now requires both no authoritative match and an empty merged result set. The regression test covers loose rows skipping suggestions and an empty lookup still requesting them.

dcadenas:
@NotThatKindOfDrLiz
> Requesting changes for two localized issues: keep the exact-email path indexed after switching to case-insensitive matching, and avoid fetching suggestions when the UI will discard them.

Both requested changes are addressed in c535918: normalized exact-email equality now has a matching functional index, and fuzzy suggestions are skipped whenever lookup results will be rendered.

NotThatKindOfDrLiz:
Takeover update:

- Pushed fix(tools): make email benchmark host check portable (b1829c3) to replace Bash 4-only host lowercasing with portable tr in tools/benchmark-email-trgm-inserts.sh.
- This fixes the local failure in tools/tests/email-trgm-insert-benchmark_test.sh on macOS/Bash 3.2.

Verification run locally:

- bash tools/tests/email-trgm-insert-benchmark_test.sh
- bun test web/src/routes/support-admin/page.test.ts
- cargo fmt --all -- --check

Fresh GitHub checks on b1829c3 are green. Remaining non-code item before merge: the PR body still says the support-admin suggestion-state screenshot is ready to attach, but it is not attached yet. @dcadenas please attach that visual artifact or update the Visuals section if a maintainer decides it is not required.

core/src/repositories/user.rs
```diff
@@ -2411,6 +2612,305 @@ mod tests {
...
+        sqlx::query(
+            "CREATE TABLE personal_keys (
+                user_pubkey TEXT NOT NULL,
+                tenant_id BIGINT NOT NULL
+            )",
+        )
+        .execute(&mut *connection)
+        .await
+        .unwrap();
+        drop(connection);
```
mbradley, line 2938: This test passes even if the `search_path` isolation ever stops hiding pg_trgm, because the isolated `users` table is empty either way. Seeding one trigram-matching row before the assertion makes it fail loudly if the fallback regresses (in the intended path, `email % $1` still fails at plan time before any row is scanned, so the row is inert):

```suggestion
        // Seed a row whose email is an exact trigram match for the query below. When
        // pg_trgm is correctly hidden, `email % $1` fails at plan time (42883) before any
        // row is scanned, so this row is irrelevant and the fallback still yields empty.
        // But if the `SET search_path` trick ever stops hiding pg_trgm (e.g. a future sqlx
        // release that resets session state on connection release), this row would be
        // returned and the assertion fails loudly, instead of passing only because the
        // isolated table happened to be empty.
        let matching_pubkey = Keys::generate().public_key().to_hex();
        sqlx::query(
            "INSERT INTO users (pubkey, email, tenant_id, status)
             VALUES ($1, 'missing@example.com', 1, 'active')",
        )
        .bind(&matching_pubkey)
        .execute(&mut *connection)
        .await
        .unwrap();
        drop(connection);
```

web/src/routes/support-admin/page.test.ts
```diff
@@ -0,0 +1,19 @@
+import { describe, expect, test } from "bun:test";
+
+const pageSource = await Bun.file(
+    new URL("./+page.svelte", import.meta.url),
+).text();
+
+describe("support admin user lookup", () => {
```
mbradley: This test asserts on the `.svelte` **source text**, so it can't catch a behavioral regression — inverting the `showingSuggestions` guard leaves all five strings present and the test stays green, while a harmless copy edit turns it red. Suggesting we make the results-vs-suggestions decision testable by extracting it into a small pure module (matching the existing `deeplink.ts` + `deeplink.test.ts` pattern) and unit-testing that.

This spans a new file and an import outside the diff, so it's a patch rather than one clickable suggestion — apply the pieces together (a partial apply would leave the test importing a missing module):

**New file — `web/src/routes/support-admin/lookup-view.ts`:**
```ts
/**
 * Pure selection helpers for the support-admin user lookup view.
 *
 * Extracted from +page.svelte so the "authoritative results vs fuzzy suggestions"
 * decision is unit-testable without a component/DOM harness. The page's $derived
 * values and the auto-expand logic delegate here.
 */

export interface LookupResult<User> {
	results: User[];
	suggestions: User[];
	total: number;
}

/**
 * Users to render: authoritative matches when there are any, otherwise the fuzzy
 * "Did you mean?" suggestions. Returns an empty list when there is no result yet.
 */
export function selectDisplayedUsers<User>(result: LookupResult<User> | null): User[] {
	if (!result) return [];
	return result.results.length > 0 ? result.results : result.suggestions;
}

/**
 * True only when the view is falling back to fuzzy suggestions: a completed lookup
 * with no authoritative results but at least one suggestion. Drives the "Did you
 * mean?" header, so it must never be true while authoritative results are shown.
 */
export function isShowingSuggestions<User>(result: LookupResult<User> | null): boolean {
	return result !== null && result.results.length === 0 && result.suggestions.length > 0;
}
```

**`+page.svelte`** — add the import next to the other `$lib`/local imports:
```ts
import { selectDisplayedUsers, isShowingSuggestions } from './lookup-view';
```
Replace the inline `displayedUsers` / `showingSuggestions` derivations with delegations:
```ts
let displayedUsers: UserDetails[] = $derived(selectDisplayedUsers(searchResult));
let showingSuggestions = $derived(isShowingSuggestions(searchResult));
```
And reuse the helper in `searchUser` (drops the duplicated ternary):
```ts
const returnedUsers = selectDisplayedUsers(result);
```

**`page.test.ts`** — replace the source-text test with a behavioral one:
```ts
import { describe, expect, test } from "bun:test";
import { isShowingSuggestions, selectDisplayedUsers } from "./lookup-view";

interface Row {
    pubkey: string;
}

const row = (pubkey: string): Row => ({ pubkey });

describe("support admin user lookup view", () => {
    test("shows authoritative results and hides suggestions when results exist", () => {
        const result = { results: [row("a"), row("b")], suggestions: [row("c")], total: 2 };
        expect(selectDisplayedUsers(result)).toEqual([row("a"), row("b")]);
        expect(isShowingSuggestions(result)).toBe(false);
    });

    test("falls back to fuzzy suggestions only when there are no results", () => {
        const result = { results: [], suggestions: [row("c")], total: 0 };
        expect(selectDisplayedUsers(result)).toEqual([row("c")]);
        expect(isShowingSuggestions(result)).toBe(true);
    });

    test("shows nothing and no suggestion banner when both are empty", () => {
        const result = { results: [], suggestions: [], total: 0 };
        expect(selectDisplayedUsers(result)).toEqual([]);
        expect(isShowingSuggestions(result)).toBe(false);
    });

    test("treats a null (pre-search) result as empty", () => {
        expect(selectDisplayedUsers(null)).toEqual([]);
        expect(isShowingSuggestions(null)).toBe(false);
    });
});
```
Behavior is unchanged (I left the lone-suggestion auto-expand alone — that's the UX call in point 3). With the guard inverted, the second test now fails where the old one passed.

mbradley (commented):
A few things on top of the existing review, mostly operational plus one support-UX call. The lookup logic and tenant isolation look solid, and both earlier change-requests read as resolved.

1. **pg_trgm on Cloud SQL (worth confirming before merge).** The migrate job runs as the `keycast-migrator` IAM role, and `CREATE EXTENSION` on Cloud SQL needs `cloudsqlsuperuser`, which IAM users don't get by default. Is pg_trgm already installed on `keycast-db-plus` and staging, or can that role create it? If not, the migration transaction fails and blocks the deploy — the app-side fallback doesn't cover it since the migration is a hard gate.

2. **Non-concurrent index build.** Both indexes build in-transaction, so the GIN trigram build holds a write lock on `users` for its duration (logins/signups/email changes block). That matches our existing migrations, and CONCURRENTLY isn't a clean fit with sqlx's per-migration transaction, so this is really a "how big is `users` in prod, and can the migrate job run in a low-traffic window" question rather than a code change. Rough current row count?

3. **Suggestions vs confirmed matches.** A fuzzy "Did you mean?" suggestion renders through the same card as an authoritative match, and a lone suggestion auto-expands, so a support admin can read a fuzzy hit as a confirmed identity. The claim-link gating already blocks the riskiest action on suggestions, so this is UX hardening: worth a per-row "suggested" affordance and skipping auto-expand while showing suggestions.

4. **Test quality.** Two inline suggestions below: `page.test.ts` currently asserts against the `.svelte` source text, so a real reactivity regression wouldn't fail it (inverting the suggestions guard stays green) while copy tweaks do; extracting the selection logic into a pure module and unit-testing that matches the `deeplink.ts` pattern. And the pg_trgm-unavailable repository test passes for the wrong reason if the `search_path` isolation ever stops holding.

Separately, `web/` unit tests don't run in CI today (no `bun test` step, and the sibling web tests are also orphaned) — broader than this PR, but worth a follow-up so these tests actually guard something.

database/migrations/20260717120000_email_trgm_search.sql
```diff
@@ -0,0 +1,9 @@
+CREATE EXTENSION IF NOT EXISTS pg_trgm;
```
NotThatKindOfDrLiz, line 1: This is still a deploy gate: if `pg_trgm` is not already installed in staging/prod, this migration requires a DB role that can create extensions. On Cloud SQL that means `cloudsqlsuperuser`; an app-side fallback for missing trigram operators will not help because the migration fails before the app starts. Please get/record the ops answer here: either `pg_trgm` is already installed on the target DBs, or the migrator role can create it, or the extension install is handled separately before this migration runs.

core/src/repositories/user.rs
NotThatKindOfDrLiz, line 2938: This fallback test can pass for the wrong reason because the isolated `users` table is empty. If `pg_trgm` stops being hidden by this connection setup later, the query still returns no rows and the assertion stays green. Please seed a row that would match `missing@example.com`; then the test only passes when the missing-extension fallback actually fires before row scanning.

web/src/routes/support-admin/page.test.ts
NotThatKindOfDrLiz: This test asserts on Svelte source text rather than behavior, so it will miss the regression it is meant to guard. For example, inverting the `showingSuggestions` condition leaves these strings present while breaking the UI decision. Please extract the results-vs-suggestions selection into a small pure helper, like the existing `deeplink.ts` pattern, and unit-test that helper instead.

NotThatKindOfDrLiz (changes requested):
Requesting changes before merge. The lookup implementation itself looks sound and Liz's prior two requested fixes are addressed, but this still has a deployment gate plus two test-quality issues that should be resolved before approval.

The suggestion-row UX affordance / lone auto-expand behavior is worth a maintainer call, but I am not treating that as the blocking item here. Please also attach the support-admin suggestion-state visual or update the Visuals section before merge.

core/src/repositories/user.rs
dcadenas: Addressed in 39958d3: the isolated users table now contains an exact trigram-matching row before the connection is released, so restored pg_trgm visibility would make the assertion fail instead of passing on an empty table. The database-backed test and full integration suite pass.

web/src/routes/support-admin/page.test.ts
dcadenas: Addressed in 39958d3: lookup selection and auto-expansion now delegate to pure helpers in lookup-view.ts, with behavioral tests covering authoritative results, suggestions, empty results, and pre-search state. Lone fuzzy suggestions remain collapsed and each suggested row is labeled explicitly.

NotThatKindOfDrLiz:
@dcadenas please finish the remaining release-readiness items on this PR:

1. Verify and document that the actual prod/staging migration runner can install `pg_trgm` via `CREATE EXTENSION IF NOT EXISTS pg_trgm`, or that the extension is already installed before this migration runs. The runtime fallback does not cover migration failure.
2. Fix the Visuals section: the current raw.githubusercontent screenshot link returns 404. Please attach/commit a resolving support-admin suggestion-state screenshot, or update the section if a maintainer decides the visual is not required.

The code review looks otherwise clean, so these are the items blocking approval/merge.

dcadenas:
Thanks @NotThatKindOfDrLiz. Both addressed.

**1. pg_trgm migration.** Verified on the Cloud Run prod path: the migrate job runs before deploy (and fails the deploy on error), and its DB user has the privileges to create the extension. `pg_trgm` is available and trusted, so `CREATE EXTENSION IF NOT EXISTS pg_trgm` succeeds as-is. No infra change needed.

**2. Visuals 404.** Removed the committed screenshots from the repo (they belong as PR attachments) and replaced the broken link with an attached screenshot in the description.

database/migrations/20260717120000_email_trgm_search.sql
dcadenas: Recorded the deployment gate in 1dbe4a0. Live production preflight confirms pg_trgm 1.6 is available/trusted and the migration connection has CREATE on the database and public schema; live staging verification is still blocked on expired GKE auth, so the rollout doc explicitly keeps that query as a hard pre-deploy requirement. Leaving this thread unresolved until the live staging evidence is available.

core/src/repositories/user.rs
dcadenas: Addressed in 188f459: the isolated users table now contains an exact trigram-matching row before the extension is hidden, so restored pg_trgm visibility would return that row and fail the assertion. The focused missing-extension fallback test passes.

web/src/routes/support-admin/page.test.ts
dcadenas: Addressed in 188f459 and e9c75d8: result selection, suggestion state, and auto-expansion now delegate to pure helpers with behavioral tests, including authoritative, loose, fuzzy-fallback, empty, and null states. The full web unit suite passes (26 tests).

api/src/api/http/admin.rs
```diff
@@ -930,55 +1062,68 @@ pub async fn get_user_lookup(
...
+    // the response, but still resolve username-shaped queries so an exact external name match can
+    // be promoted ahead of those candidates.
+    if should_attempt_authoritative_name_promotion(q, lookup.authoritative_match)
         && crate::divine_names::is_enabled()
     {
-        if let Ok(Some(hex_pubkey)) = crate::divine_names::lookup_by_name(q).await {
-            users = user_repo
+        if let Some(Ok(Some(hex_pubkey))) = await_name_promotion_within(
+            crate::divine_names::lookup_by_name(q),
+            ADMIN_NAME_PROMOTION_TIMEOUT,
```
NotThatKindOfDrLiz: This 250 ms cap now applies to the zero-local-result fallback too. On `origin/main`, a username-shaped query with no local rows awaited `divine_names::lookup_by_name` under that function's existing 3 s reqwest timeout; this change silently drops any cold DNS/TCP/TLS/name-server response that takes over 250 ms and returns `No user found`. The short budget makes sense when local candidates are already available, but when `lookup.users` is empty it regresses the primary fallback path. Please make the timeout conditional on having local candidates, or preserve the old fallback budget for empty results.

core/src/repositories/user.rs
```diff
@@ -1864,14 +2035,87 @@ impl UserRepository {
...
+                u.suspended_reason,
+                u.created_at,
+                u.updated_at
+             FROM users u
+             LEFT JOIN personal_keys pk ON pk.user_pubkey = u.pubkey AND pk.tenant_id = u.tenant_id
+             WHERE u.email IS NOT NULL
+               AND u.email % $1
+               AND similarity(u.email, $1) >= $3
+               AND u.tenant_id = $2
+             ORDER BY similarity(u.email, $1) DESC, u.pubkey",
```
NotThatKindOfDrLiz: This suggestion query still has no SQL row bound. `fetch_all` will return every trigram candidate before the Rust Levenshtein filter and `.take(5)`, which can fan out badly for tenants with many addresses sharing a long domain. The benchmark doc's ranked-similarity plan includes a `Limit`, so the verified plan is not the shipped query plan. Please add a generous SQL `LIMIT` before the Rust filter and recapture the EXPLAIN so the doc matches the code.

web/src/routes/support-admin/lookup-view.ts
```diff
@@ -0,0 +1,181 @@
...
+		result.results.length === 0 &&
+		result.suggestions.length > 0
+	);
+}
+
+/** Select a lone authoritative result for automatic expansion. */
+export function selectAutoExpandedPubkey<User extends { pubkey: string }>(
+	result: LookupResult<User> | null
+): string | null {
+	if (!result?.authoritative_match || result.results.length !== 1) return null;
```
NotThatKindOfDrLiz: This still keys auto-expand off the total number of primary rows, not the number of authoritative rows. For a search like exact username `bob`, `find_users_for_admin` can return the authoritative username row plus unrelated email-substring candidates; `authoritative_match` is true, but `results.length !== 1`, so the old single-hit auto-expand behavior no longer fires. Please preserve auto-expand when there is exactly one authoritative match, even if loose candidates are also displayed.

web/src/routes/support-admin/page.test.ts
```diff
@@ -0,0 +1,128 @@
+import { describe, expect, test } from "bun:test";
```
NotThatKindOfDrLiz, line 1: These new Bun tests are not run by the PR gate: `build-test-push.yaml` has no web/Bun test step, and `web/package.json` does not define a test script. Please wire `cd web && bun test` (or an equivalent `web` test script) into CI so this new coverage can actually catch regressions.

tools/tests/email-trgm-insert-benchmark_test.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
NotThatKindOfDrLiz: Same issue for the benchmark safety test: nothing in the workflow or package scripts runs `tools/tests/*_test.sh`. Since this PR adds this test as part of the migration safety story, please add a CI step to execute it instead of leaving it as orphaned coverage.

docs/EMAIL_TRGM_INSERT_BENCHMARK.md
```diff
@@ -0,0 +1,130 @@
...
+  --clients 4 \
+  --transactions 10000 \
+  --baseline-rows 400000
+```
+
+The script refuses non-local database URLs.
+
+## Deployment Gate
+
+The migration must run in a low-traffic maintenance window. Do not deploy it through a normal unattended rollout: both index statements are non-concurrent and block writes while their transaction remains open. `CREATE INDEX CONCURRENTLY` is not a drop-in replacement because PostgreSQL forbids it inside the transaction SQLx uses for a migration.
```
NotThatKindOfDrLiz: This deployment gate does not match the actual deploy path. `cloudbuild.yaml` executes the `keycast-migrate` Cloud Run Job during normal unattended deploys, and that job runs `./keycast --migrate` without the `PGOPTIONS` timeouts prescribed below. Either enforce the lock/statement timeout story in the real migration job, or revise this doc to describe the measured unattended rollout risk that the pipeline actually accepts.

NotThatKindOfDrLiz (changes requested):
Requesting changes. The main lookup tiers and tenant isolation look careful, but the current head still regresses the existing Divine-name fallback, leaves the fuzzy suggestion query unbounded, and adds tests/deployment guidance that are not enforced by the actual merge/deploy path. Details inline.

mbradley (commented):
Concur with the current change-request review; nothing to add on those items.

Two minor, non-blocking extras:

1. `get_user_lookup` hand-rolls a pubkey dedup/extend/truncate that duplicates the repository's `merge_admin_lookup_candidates` helper. Worth folding into the shared helper so the two paths can't drift.

https://github.com/divinevideo/keycast/blob/225f2003e349510ea5d6afaa1819b9c5d07c97e1/api/src/api/http/admin.rs#L1082-L1096

2. The `find_users_for_admin` doc comment says queries without `@` "merge email substring matches with the existing username, vine ID, and pubkey paths," but the `npub` branch returns before any merge, so the comment slightly overstates the behavior.

https://github.com/divinevideo/keycast/blob/225f2003e349510ea5d6afaa1819b9c5d07c97e1/core/src/repositories/user.rs#L1858-L1861

database/migrations/20260717120000_email_trgm_search.sql
dcadenas: Confirmed and recorded in the deployment notes: pg_trgm 1.6 is available/trusted, the production migrator has CREATE privileges, and staging's database-owner migration role can install the trusted extension. Commit 5ca67e7 aligned the rollout description with the actual migration path, and the PR body retains the target-environment evidence.

api/src/api/http/admin.rs
dcadenas: Fixed in 3323422: the 250 ms promotion budget is used only when local candidates already exist; a zero-local-result lookup preserves the existing upstream timeout for the primary Divine-name fallback. The API regression tests cover both paths.

core/src/repositories/user.rs
dcadenas: Fixed in 668c53d: the similarity-ranked SQL candidate query now applies LIMIT 100 before Rust's edit-distance filter and final five-result cap. The one-off benchmark artifact was later removed in 4a29e99, while the corrected populated-plan result remains in the PR description.

web/src/routes/support-admin/lookup-view.ts
dcadenas: Fixed in 19e6a49 and preserved through c185a04: auto-expansion keys off authoritative_count, and per-row authority survives merging and serialization. The web regression test verifies one authoritative row auto-expands while the accompanying loose row remains visible and labeled Suggested.

web/src/routes/support-admin/page.test.ts
dcadenas: CI wiring was not added in this feature PR; that broader repository-level proposal was explicitly triaged out of scope. The suite is still verified locally (`bun test web`, 27/27 passing), so I am leaving this thread unresolved rather than claiming the CI concern is fixed.

tools/tests/email-trgm-insert-benchmark_test.sh
dcadenas: Resolved by 4a29e99, which removes the one-off benchmark script, workload, shell test, and long-form document. The actual migration, migration regression test, and durable repository/API/UI behavior tests remain.

docs/EMAIL_TRGM_INSERT_BENCHMARK.md
dcadenas: The mismatched one-off rollout document was removed in 4a29e99. The PR deployment notes now describe the actual unattended SQLx migration path, the transactional non-concurrent write lock, the measured 1,903 ms local lower bound, and the required rollout monitoring.

dcadenas:
@NotThatKindOfDrLiz
> Requesting changes. The main lookup tiers and tenant isolation look careful, but the current head still regresses the existing Divine-name fallback, leaves the fuzzy suggestion query unbounded, and adds tests/deployment guidance that are not enforced by the actual merge/deploy path. Details inline.

All blocking code and deployment-documentation findings are addressed in 3323422, 668c53d, 19e6a49, and 5ca67e7. The full Rust checks, CI-equivalent database-backed suite, 27 Bun tests, and benchmark safety test pass; broader CI wiring remains tracked in #305 and #306.

mbradley (commented):
Re-reviewed the delta since the last round (`225f2003`..`4a29e99`). The four code fixes hold up under trace plus compile/clippy/unit tests and mutation-checking the web tests; no regressions.

- Name-promotion timeout is now conditional: empty-result lookups keep the upstream ~3s budget, 250ms cap applies only when local candidates exist (covered by `empty_local_lookup_preserves_upstream_name_timeout`).
- Suggestion query bounded by `LIMIT 100` after `ORDER BY similarity`, before the Rust filter; EXPLAIN-asserted test confirms the `Limit` node.
- Per-row `authoritative` flag now threads through the merge, API response, and frontend; auto-expand keys off `authoritative_count` and the "Suggested" badge is per-row. Tests fail under inverted conditions.
- Merge policy moved to the shared core helper and the `find_users_for_admin` docstring is corrected.

Remaining items are ops/process rather than code, and already tracked: web-unit-test CI (#305) and the live staging `pg_trgm` preflight. On migration deploy-safety: both indexes are non-concurrent `CREATE INDEX`, so the build write-locks `users`, and the `keycast-migrate` job runs `./keycast --migrate` with no `lock_timeout`/`statement_timeout`. Moving that note from the deleted doc into the PR body is fine, but the write-lock gate should be honored at deploy (or add a `lock_timeout` to the migrate path) rather than lost at merge.

Optional hardening: `selectAutoExpandedPubkey` returns `results[0].pubkey`, correct only because authoritative rows sort first; `results.find(u => u.authoritative)?.pubkey` would remove that implicit dependency.

NotThatKindOfDrLiz:
Takeover update pushed in `755cb3e`.

What changed:
- Snapshotted the executed support-admin query so email highlights and Did-you-mean diffs do not drift when the input changes after results render.
- Made service-token batch email lookup case-insensitive for stored mixed-case emails and added a regression test.
- Removed the dead `find_user_for_admin` wrapper/tests and the unused public `ADMIN_USER_LOOKUP_LIMIT` re-export.
- Recomputed authoritative metadata after appending loose promoted-name results.
- Updated the PR body to avoid overstating trigram index plan behavior for low-selectivity similarity queries.

Verification:
- `cd web && bun test src/routes/support-admin/page.test.ts`
- `cargo test -p keycast_api --features integration-tests --test batch_lookup_test --no-run`
- GitHub `test` and semantic PR checks are green on `755cb3e`.

Local runtime integration tests could not use the repo Postgres stack because port 5432 is already owned by `config-postgres-1` with different credentials; I cleaned up the repo Docker containers/volumes created during that attempt.

Remaining pre-merge gate: @dcadenas, please decide accepting the non-concurrent users-index migration write lock during rollout, or add a deliberate timeout/migration-path mitigation before merge. The PR body now records the write-lock behavior, but this still needs an affirmative rollout decision.

web/src/routes/support-admin/page.test.ts
dcadenas: This broader CI wiring was intentionally kept out of PR #301. The web suite is verified locally (`bun test web`, 27/27), but no CI change landed here, so I am leaving this thread unresolved.

Issue #49, opened 2026-03-04
Support admin search: case-insensitive matching and name server fallback


mbradley:
## User Story

As a support admin looking up a creator's account, I want to search by the username I can see on their divine.video profile and find their Keycast account, so I don't have to know internal identifiers or hunt for hex pubkeys.

## Current Behavior

User Lookup at login.divine.video/support-admin uses exact string matching (`WHERE username = $1`). This creates a frustrating experience when the Keycast username (imported from Vine) differs from the divine.video username (managed by the name server).

**Example:** A creator's public profile is visible at `lelepons.divine.video`. A support admin searches `lelepons` in Keycast and gets no result. The Keycast username is `Lele.Pons` (original Vine capitalization and punctuation). The admin has to:

1. Go to names.admin.divine.video, search for `lelepons`, copy the hex pubkey
2. Return to login.divine.video/support-admin, paste the full 64-character hex pubkey
3. Now they find the account

This is a common scenario for Vine creators whose original usernames had dots, mixed case, or other formatting that was normalized when imported into the name server.

### 1. Case-insensitive, punctuation-normalized search

When searching by username, normalize both the query and the stored value before comparing. At minimum use case-insensitive matching (`LOWER()` or `ILIKE`). Ideally also strip dots and common punctuation so that `lelepons` matches `Lele.Pons`.

### 2. Name server fallback

If local search returns no results, try resolving the query against the name server (`divine.video/.well-known/nostr.json?name=<query>`). If that returns a pubkey, search Keycast by pubkey. This would let support staff search by any divine.video username without needing to know it differs from the Keycast username.

## Context

- Search logic is in `core/src/repositories/user.rs`, `find_user_for_admin()` function
- The name server and Keycast have independent username stores with no synchronization
- This came up during VIP creator onboarding documentation and testing

mbradley:
Update after tracing current `main`: the two improvements this issue proposed have
largely landed. `find_users_for_admin` already does case-insensitive,
punctuation-normalized username matching, and `get_user_lookup` already falls back to the
Divine name server (`divine_names::lookup_by_name`), enabled in production
(`ENABLE_DIVINE_NAMES=true`).

But the underlying problem is still live (reproduced in a support check-in). Re-scoping to
the three specific gaps that remain:

1. **Labeling.** The search hint reads "Search by email, Vine username, vine_id, hex
   pubkey, or npub" (`web/src/routes/support-admin/+page.svelte:273`). It says "Vine," so
[...]

PR #312, opened 2026-07-23
feat(support-admin): surface minor/age-review terminal status in user lookup

### LINKED ISSUE #309: feat(support-admin): surface verified_minor + account status in user lookup
## User story

As a support admin looking up an account, I want to see whether it is an approved
protected minor and its current account status, so I can tell someone the terminal
outcome of the age-review process without hunting across tools or sending them from
pillar to post.

## Background

Surfaced in a support check-in: when a user says "I can't post / features don't work,"
support has no way to tell from the lookup tool whether the account went through age
review and where it landed. Today they guess or escalate.

#263 already exposed `verified_minor` on the client-facing `GET /user/account`. This
issue does the equivalent for the human support tool: the same durable signals, shown
in login.divine.video/support-admin.

## Scope (keycast-only, terminal state)

Surface the account's terminal status in the support-admin user lookup:
- `verified_minor` + `verified_minor_at` (approved protected minor, 13 to 15)
- `status` (active / suspended / banned) + `suspended_reason`

`status` and `suspended_reason` are already returned by the lookup API
(`UserLookupDetails`) but silently dropped by the frontend, so surfacing account status
is a frontend-only change. `verified_minor` needs adding to the query and response.

Out of scope: the "reported / currently in age review" pending state. That queue
(`age_review_cases`, the 13-15 band, reviewer decision) lives in divine-relay-manager,
tracked by divinevideo/divine-relay-manager#141. This issue covers only the terminal
signals keycast owns.

## Implementation notes

- `core/src/repositories/user.rs` (`find_users_for_admin` / `AdminUserDetails`): add
  `verified_minor`, `verified_minor_at` to the SELECT and struct. Columns already exist,
  no migration. Pattern already present in `get_full_admin_status`.
- `api/src/api/http/admin.rs` (`get_user_lookup` / `UserLookupDetails`): add the two
  fields to the response.
- `web/src/routes/support-admin/+page.svelte`: add `status`, `suspended_reason`,
  `verified_minor`, `verified_minor_at` to the `UserDetails` interface and render them
  (the status strip is the natural home). Present the pair so support can read the
  outcome, e.g. cleared+active, verified_minor + suspended("age_review") = pending,
  banned = confirmed-and-rejected.
- Keep it UCAN-gated (support-admin). Do not route the service-token status endpoint to
  browser sessions; extend the existing UCAN lookup query instead.

Caveat: `suspended_reason` is unstructured free text with no enum tying it to
"age_review". If support needs to reliably distinguish suspension causes later, that is a
separate data-quality change.

## Acceptance

Looking up an approved-minor account in support-admin shows `verified_minor: true`
(+ timestamp), and every account shows its status / suspended reason. No new migration.

Part of the protected-minor safeguards epic (divinevideo/support-trust-safety#173);
sibling of #263 (client exposure) and divinevideo/divine-relay-manager#141
(pending-review visibility).

mbradley:
## Summary

- Surfaces an account's terminal age-review status in the support-admin user lookup: `verified_minor` + `verified_minor_at` (approved protected minor, 13-15), plus the account `status` / `suspended_reason` that the API already returned but the UI dropped.
- Backend: adds the two `verified_minor` columns to `AdminUserDetails` and all three admin-lookup SELECTs (`find_users_for_admin`, the email-suggestion query, and the batch/email lookup), threaded through the shared `enrich_user_lookup_details` helper so both results and "did you mean" suggestions carry them.
- Frontend: renders account status (active / suspended / banned + reason) and a protected-minor badge on the existing lookup rows and status strip.
- Builds on the lookup rework in #301 (merged); no changes to its search/authoritative logic.

## Motivation

- From a support check-in: when a user says "I can't post / features don't work," support had no way to tell from the lookup tool whether the account went through age review or where it landed, so they guessed or escalated ("it's an opaque process to me"). This gives them the terminal outcome at a glance.
- Sibling of #263 (which exposed `verified_minor` on the client-facing `/user/account`); this does the equivalent for the human support tool.

## Related Issue

- Closes #309
- Out of scope by design: the reported / in-review (non-terminal) age-review state, which lives in divine-relay-manager (`age_review_cases`), not keycast. Divine-username search is tracked separately in #49.

## Testing

Targeted (scope recorded per AGENTS.md; the full `bun run test` suite runs in CI):
- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated` - clean
- `cargo test -p keycast_core --features integration-tests --lib find_users_for_admin` - 8 pass, incl. `test_find_users_for_admin_surfaces_verified_minor` (mutation-verified: flipping the fixture to `FALSE` makes it fail)
- `cargo test -p keycast_api --features integration-tests --test user_lookup_test` - 4 pass: `verified_minor` on exact results, `verified_minor` on the "did you mean" suggestion path, suspended `status`/`suspended_reason`, and the non-admin auth gate
- `cargo test -p keycast_api --features integration-tests --test batch_lookup_test` - 12 pass (regression guard for the shared `AdminUserDetails` SELECT)
- #301's `user_lookup_response_tests` still green (field additions didn't break serialization/tier tests)
- `npm run check` (svelte-check): `support-admin/+page.svelte` clean

## Visuals

- UI/web change. A browser screenshot was not captured locally (no bun dev server in this environment); svelte-check passes on the page. The change adds, to each lookup result row and the expanded status strip: an account-status label (active / suspended / banned + reason) and a "Minor" badge for approved protected minors. Recommend capturing a screenshot before merge.
- No sensitive external brand/partner names.

### DIFF
--- a/api/src/api/http/admin.rs
+++ b/api/src/api/http/admin.rs
@@ -915,6 +915,11 @@ pub struct UserLookupDetails {
     pub status: String,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub suspended_reason: Option<String>,
+    /// Approved protected-minor (13-15) flag. Terminal age-review signal; the in-review
+    /// state lives in relay-manager, not keycast.
+    pub verified_minor: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub verified_minor_at: Option<String>,
     pub active_sessions: i64,
     pub created_at: String,
     pub last_active: Option<String>,
@@ -947,6 +952,8 @@ async fn enrich_user_lookup_details(
         has_personal_key: details.has_personal_key,
         status: details.status.as_str().to_string(),
         suspended_reason: details.suspended_reason,
+        verified_minor: details.verified_minor,
+        verified_minor_at: details.verified_minor_at.map(|at| at.to_rfc3339()),
         active_sessions: sessions.len() as i64,
         created_at: details.created_at.to_rfc3339(),
         last_active,
@@ -996,6 +1003,8 @@ mod user_lookup_response_tests {
             has_personal_key: false,
             status: UserStatus::Active,
             suspended_reason: None,
+            verified_minor: false,
+            verified_minor_at: None,
             created_at: now,
             updated_at: now,
         }
@@ -1044,6 +1053,8 @@ mod user_lookup_response_tests {
             has_personal_key: false,
             status: "active".to_string(),
             suspended_reason: None,
+            verified_minor: false,
+            verified_minor_at: None,
             active_sessions: 0,
             created_at: "2026-07-17T12:00:00+00:00".to_string(),
             last_active: None,
--- a/api/tests/user_lookup_test.rs
+++ b/api/tests/user_lookup_test.rs
@@ -1,0 +1,280 @@
+// ABOUTME: HTTP-handler tests for the support-admin user-lookup endpoint (get_user_lookup)
+// ABOUTME: Locks the response contract for minor/age-review terminal status + the auth gate (#309)
+
+#![cfg(feature = "integration-tests")]
+
+mod common;
+
+use axum::extract::{Query, State};
+use chrono::Utc;
+use keycast_api::{
+    api::{
+        extractors::UcanAuth,
+        http::{
+            admin::{get_user_lookup, UserLookupQuery},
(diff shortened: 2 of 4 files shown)


dcadenas (approved):
✅ clearer age-review status in user lookup

PR #313, opened 2026-07-23
fix(support-admin): make Divine username search authoritative

### LINKED ISSUE #49: Support admin search: case-insensitive matching and name server fallback

mbradley:
## Summary

Builds on #301, which consults the Divine name server only when the local search finds nothing authoritative. Three gaps remained; each item is new on top of #301:

- **A stale exact `username` no longer shadows the real account.** #301 skips the name server when a local row exactly matches, so an outdated local `username` could win (the "wrong person" case). Now the name server is consulted for any handle-shaped query even on an exact local match; when it resolves to a keycast account, that account is the sole authoritative match and the disagreeing local row is demoted to a candidate.
- **Handle forms are parsed** — `mjb`, `@mjb`, `mjb.<domain>`, `@mjb.<domain>`, `mjb@<domain>` reduce to the bare handle (#301 routes anything with `@` to email).
- **Name resolution is case-insensitive** (#301's `nostr.json` lookup was exact-key only).

Plus a label fix: search hint/placeholder "username" → "Divine username".

## Motivation

A support check-in surfaced that searching a creator's Divine handle could return the wrong account. The name server owns handles, so the lookup should defer to it, not a stale local `username`.

## Related Issue

Closes #49.

## Review note

Reworks #301's name-promotion gating and merge (which dcadenas reviewed): replaces `should_attempt_authoritative_name_promotion` with handle-shape gating, adds `AdminUserLookup::demote_to_candidates`, and renames `append_loose_results` to `merge_ranked_results` now that it merges lookups of any authority. Tradeoff: the name server is consulted for every handle-shaped query, bounded by #301's existing 250ms budget when local results exist. `is_handle` is syntactic, so a coincidental `vine_id` match is demoted too. That is intentional, since the name server owns a registered handle, and the row stays visible as a candidate.

Resolved across three review rounds from dcadenas:

1. Canonicalize against the public handle domain rather than the tenant login host, so `mjb@divine.video` and `mjb.divine.video` actually reduce to `mjb` in production.
2. A name-server timeout no longer leaves a stale local row marked authoritative. Demotion drops the `match_kind` tier too, so a timed-out match cannot render as the top-tier identity.
3. A real `<local>@divine.video` mailbox keeps its exact-email authority. That form has two honest readings, a mailbox and the NIP-05 spelling of the handle, and they can belong to different accounts. Both are searched and merged. When they disagree both owners stay confirmed, which takes `authoritative_count` past 1 and stops `selectAutoExpandedPubkey` picking one. When a single account owns both, it dedupes to one confirmed row and stays auto-expandable. The merge runs after name promotion because the name server is authoritative for handles, not mailboxes.

Round 3 narrows the round 2 invariant: a timeout now demotes the handle reading only, leaving an exact local email match confirmed. That distinction is pinned by a test and called out in the review thread for an explicit read.

## Testing

fmt and clippy clean. Unit: canonicalization including `agent@qa.<domain>` staying on the email path and the NIP-05 form retaining its address, the promote/demote merge, merge behavior when the two readings disagree and when they dedupe, case-insensitive extraction. Integration (`api/tests/divine_username_lookup_test.rs`): all five handle forms resolve to one account through `get_user_lookup`; a mailbox whose local part is another account's handle keeps both owners confirmed; a single owner of both readings stays one confirmed row. Also `api/tests/divine_username_email_collision_test.rs`, which runs the collision against a live mock name server so the merge is exercised through `apply_name_promotion` rather than only the local path: one case where the name server confirms the handle, one where it times out. Full core, API and signer suites pass locally, and new guards were mutation-checked rather than just observed green.

Not run locally: the `web` suite uses `bun test` and bun is not installed on this machine. It is unchanged by this PR's later rounds and passes in CI.

## Visuals

Text-only: hint/placeholder now say "Divine username". svelte-check clean; no screenshot (no bun dev server locally).


api/src/api/http/admin.rs
```diff
@@ -1122,16 +1318,25 @@ pub async fn get_user_lookup(
         return Err(ApiError::bad_request("Query parameter 'q' is required"));
     }
 
+    // Reduce the several ways a Divine handle is written (@mjb, mjb.<domain>, mjb@<domain>) to the
+    // bare handle, and mark whether this is a handle-shaped query at all.
+    let nip05_domain = crate::api::http::auth::resolve_nip05_domain(&tenant.0.domain);
```
dcadenas: Use the public Divine handle domain when canonicalizing support lookups.
At `api/src/api/http/admin.rs:1323`, `resolve_nip05_domain` supplies the tenant host, which is `login.divine.video` in production, while public handles use `divine.video`. As a result, `mjb@divine.video` remains an email search and `mjb.divine.video` is not reduced to `mjb`. Pass an explicit public handle or name-server domain here and add a production-shaped test covering `mjb`, `@mjb`, `mjb.divine.video`, `@mjb.divine.video`, and `mjb@divine.video`, while keeping unrelated emails on the email path.

api/src/api/http/admin.rs
```diff
@@ -1122,16 +1318,25 @@ pub async fn get_user_lookup(
...
+    // query even when a local row already matches: a stale local `username` must not shadow the
+    // real account. When the name server resolves to a keycast account, that account becomes the
+    // sole authoritative match and every local row is demoted to a candidate. Note `is_handle` is
+    // syntactic, so a local match on some other column (e.g. a coincidental `vine_id`) is demoted
+    // too; that is intentional — the name server owns a registered handle regardless — and the
+    // demoted row stays visible as a candidate rather than being hidden.
+    if canonical.is_handle && crate::divine_names::is_enabled() {
+        let name_lookup = crate::divine_names::lookup_by_name(&canonical.query);
         let promoted_name =
             if let Some(timeout) = authoritative_name_promotion_timeout(lookup.users.len()) {
```
dcadenas, line 1565: Do not keep a stale local row marked authoritative when the name-server check times out.
At `api/src/api/http/admin.rs:1341`, the 250 ms timeout returns the existing lookup unchanged, so a stale exact username can still be selected as the authoritative match when the name server is slow. Keep the local row visible if needed, but demote it or expose authority as unavailable until an authoritative response arrives, and add a delayed lookup test with a disagreeing owner.

dcadenas (dismissed):
The support lookup still fails for production-qualified public handles because it derives the public suffix from the login tenant host.
It also preserves stale local authority when the name-server check exceeds the 250 ms timeout.
Address both paths before merging so qualified handles resolve through the authoritative service and timeouts cannot select the wrong account.

mbradley:
marking this back to draft for a moment. i got ahead of myself pushing the review fixes before reconciling with main, which moved under the branch and touched the same files. resolving that conflict now and re-running the tests, then i'll re-request review once it's clean.

mbradley:
resolved. merged main (it had moved under the branch and touched the same files) and kept both review fixes. finding 2 needed re-integrating into main's new match_kind ranking: demotion now drops the tier too, not just the authoritative flag, so a stale timeout match can't render as the top-tier identity. all rust tests + clippy + fmt green, and the PR's test + web-tests CI are green. re-requesting review.

api/src/api/http/admin.rs
```diff
@@ -996,13 +992,123 @@ async fn await_name_promotion_within<T>(
...
+    }
+
+    // A leading '@' is a common way to write a handle.
+    let s = q.strip_prefix('@').unwrap_or(q);
+    let domain = nip05_domain.trim().to_lowercase();
+    let lower = s.to_lowercase();
+
+    if !domain.is_empty() {
+        // NIP-05 email form `handle@domain` reduces to the bare handle.
+        if let Some(handle) = strip_ascii_suffix(s, &lower, &format!("@{domain}")) {
```
dcadenas, line 1073: Guard this canonicalization so a real `<local>@divine.video` mailbox is not treated only as a handle. Stripping the public handle domain here removes `@` before `find_users_for_admin` can take its exact-email authoritative path, so a different account that owns the bare handle can be promoted as the sole confirmed match. Run the lookup against the original email as well as the reduced handle and merge the results, then add a test covering these two different owners and verify the handle owner is not auto-expanded in place of the email owner.

dcadenas (changes requested):
Please preserve the exact-email lookup when canonicalizing the public handle domain. A real `<local>@divine.video` address is currently reduced to a bare handle, so the email owner loses authoritative status and a different handle owner can become the sole confirmed result. Before merge, handle both interpretations and add integration coverage for a mailbox whose local part belongs to another account's Divine handle, while keeping the existing pure-handle forms unchanged.

api/src/api/http/admin.rs
mbradley: Fixed. Confirmed the diagnosis first: stripping the public handle domain removed the `@` before `find_users_for_admin` could take its exact-email path. The consequence is worse than a ranking problem, because `selectAutoExpandedPubkey` auto-expands whenever exactly one row is confirmed, so support landed inside the handle owner's account without ever seeing the mailbox owner.

`CanonicalLookup` now carries an `email_form`, set only for the NIP-05 shape since that is the one query that is also a valid address. The handler runs both readings and merges them. When they resolve to different accounts both stay confirmed, which pushes `authoritative_count` to 2 and turns auto-expand off so support disambiguates. When one account owns both readings it dedupes back to a single confirmed row and stays auto-expandable.

The merge runs after name promotion on purpose: the name server is authoritative for handles, not for mailboxes, so a slow or unregistered handle must not demote an exact email match.

Coverage in `api/tests/divine_username_lookup_test.rs`: the two-owner case you asked for (a mailbox whose local part is another account's Divine handle), a single-owner case guarding the dedupe, and assertions that all four pure-handle forms still confirm only the handle owner and stay auto-expandable. The two-owner test fails on the previous commit with exactly the symptom you described. I also mutation-checked the new guards rather than trusting green: reversing the merge order, dropping the authority recompute, and removing dedupe each fail a specific assertion.

Separately, `api/tests/divine_username_email_collision_test.rs` runs the collision against a live mock name server. The tests above have the name server disabled, so they never exercise `apply_name_promotion` demoting local rows before the email reading is merged back in, which is the production shape and the path your first two rounds were about.

One consequence worth your explicit sign-off, since it narrows what you asked for in the first round. A name-server timeout no longer leaves the lookup with nothing confirmed: the mailbox owner stays confirmed, because an exact local email match is a fact the name server has no say over. The handle reading is still demoted exactly as you asked. I pinned the distinction in `name_server_timeout_demotes_the_handle_but_not_the_mailbox_owner` so it cannot regress quietly, and verified it fails both ways (removing the merge drops the timeout case to zero confirmed rows). If you read the original invariant as covering the whole lookup rather than the handle reading, say so and I will restrict the merge to the confirmed path.

One deliberate call worth naming: when the readings disagree I ranked the mailbox owner first, reasoning that it is the query exactly as typed while the handle reading required stripping a suffix. Both are confirmed either way, so this only affects display order. Happy to flip it if you read it the other way.

Also renamed `append_loose_results` to `merge_ranked_results`, since it merges lookups of any authority rather than just loose candidates and the old name would have been misleading at the new call site.

Unrelated to this finding but surfaced by the new coverage: the first `reqwest::Client` built in a process costs roughly 335ms, which exceeds the 250ms promotion budget, so the first handle lookup on a fresh instance can time out and demote a legitimate match. It predates this PR (the budget and the per-call clients are both on `main`), so I filed #324 rather than widening this change.

mbradley:
Third review round resolved, detail in the thread on `admin.rs`. Short version: `<local>@divine.video` is now searched as both a mailbox and a handle, and when those are two different accounts both stay confirmed so the UI stops auto-expanding either one. New integration coverage for the two-owner case and the single-owner dedupe, with the pure-handle forms pinned unchanged. fmt, clippy, core, all API integration suites and signer green locally, and CI is green.

I call out a behavior change in thread, that I want your headcheck on, thanks sir.

dcadenas (approved):
✅ authoritative Divine username lookup. The accepted findings from the prior cycle were verified. No in-scope repair regression was established this cycle. The full change was not re-reviewed this cycle.

PR #316, opened 2026-07-23
feat: nip17_wrap_batch RPC (send-side gift-wrap batch) — mirror of #254, cut 4 RPCs/DM to 1

### LINKED ISSUE #292: feat: nip17_wrap_batch RPC (send-side gift-wrap batch) — mirror of #254, cut 4 RPCs/DM to 1
## Summary

The send-side counterpart to #254 (`nip44_decrypt` batch). A remote-signer NIP-17 **1:1 DM send makes 4 serial RPCs** — 2 per gift wrap (`nip44_encrypt` the kind-13 seal + `sign_event` the seal), for the recipient wrap and the self-addressed wrap (the outer kind-1059 wrap uses an ephemeral local key, so it never touches Keycast). Add a wrap-batch verb so one RPC seals + signs both wraps.

## Problem / Opportunity

4 serial round trips per send, each paying the same per-call fixed overhead #254 documents (auth / BLAKE3 lookup, the uncached user-status `SELECT`, DPoP JTI Redis, ECDH derivation) and **amplifying the pool contention in #291** (~4N connection demand per N sends). Group DM to N recipients is `2(N+1)` RPCs. On a degraded server (see #291) this serial fan-out is what turns a slow signer into an undelivered message.

## Proposed approach

- New `nip17_wrap_batch` RPC taking a rumor + the recipient set (recipient + self), returning the kind-1059 gift wraps in order with **per-item error slots** (one bad recipient must not fail the batch). The server holds the key and `nostr_sdk::EventBuilder::gift_wrap(sender, receiver, rumor, tags)` is already in the crate (`http_rpc_handler.rs:681`, `nostr_rpc.rs:1063`); reuse one `HttpRpcHandler` / `SigningSession` per batch. Mirror the receive-side `nip17_unwrap_batch` (`nostr_rpc.rs:220`). Bound batch size; keep the single-op `nip44_encrypt` / `sign_event` path for back-compat.
- **Mobile (divine-mobile):** add `nip17WrapBatch` to `KeycastRpc` mirroring `nip17UnwrapBatch`, wired through `nip17_message_service._buildBothWraps` for remote signers, **behind the existing method-not-found fallback** (`keycast_rpc.dart`) so it degrades to today's 4-call path until Keycast ships the verb.

## Impact

- 1:1 DM: **4 → 1** RPC. Group to N: **2(N+1) → 1**.
- Helps even while the server stays slow (1 slow call beats 4 serial), and cuts the per-send pool demand that drives #291.

## Alternatives / interim (mobile-only, no server change)

- Build the two wraps concurrently (`4 → 2` concurrent rounds).
- Defer the self-addressed wrap out of the send's critical path (`selfWrapStatus` / `publishSelfWrap` already exist), driven by the retry sweep.

Both are smaller wins that overlap with this verb; the batch verb supersedes them.

## Related

- #254 (decrypt / receive batch) — this is the encrypt / send symmetric counterpart.
- #291 (RPC latency / pool contention) — this reduces the per-send pool demand behind it.
- #257 (spawn_blocking pool bound).

cc @dcadenas @NotThatKindOfDrLiz

dcadenas:
## Summary

Sending one NIP-17 message through a remote signer currently takes four HTTP calls for the recipient and self copies.
The new `nip17_wrap_batch` method sends one shared rumor to up to 100 recipients in one authenticated call and returns one ordered result per recipient.
Duplicate recipients stay aligned, while malformed recipient keys get isolated error slots.

## Motivation

The server must not bypass the permissions and minor-account safeguards that protect the existing encrypt and sign methods.
The new builder reuses `nip44_encrypt` and `sign_event`, validates rumor authors and IDs, and checks both batch targets and embedded `p` tags for protected accounts.
Wrapping runs sequentially under the recipient cap to keep server work bounded.
This is HTTP-only and does not change NIP-46, the frontend, the database, or deployment settings.

## Related Issue

- Closes #292

## Testing

The full workspace gates passed, along with PostgreSQL round-trip tests for ordering, partial failures, account restrictions, revocation, activity logging, and protected-account containment.

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -15,7 +15,7 @@ use keycast_core::repositories::{
 };
 use keycast_core::signing_session::{parse_cache_key, CacheKey, SigningSession};
 use keycast_core::traits::CustomPermission;
-use nostr_sdk::{Event, Keys, PublicKey, UnsignedEvent};
+use nostr_sdk::{Event, JsonUtil, Keys, PublicKey, UnsignedEvent};
 use secrecy::ExposeSecret;
 use serde::{Deserialize, Serialize};
 use serde_json::Value as JsonValue;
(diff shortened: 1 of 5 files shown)


NotThatKindOfDrLiz:
@dcadenas I took over the final cleanup pass on this PR and pushed `15a5d90`.

Changes made:
- collapsed the duplicate wrap/unwrap error-slot helpers into one shared helper
- hoisted shared rumor id normalization/serialization out of the per-recipient wrap loop
- corrected the sequential wrapping comment and the `WrapError` wire-surface docs
- added assertions that successful outer gift wraps include the recipient `p` tag, both in the non-DB unit coverage and the RPC integration test

Verification:
- `cargo test -p keycast_api handlers::http_rpc_handler::tests::nip17_wrap --verbose`
- `cargo test -p keycast_api api::http::nostr_rpc::tests::wrap_gift_wrap_batch_preserves_order_duplicates_and_partial_errors --verbose`
- `cargo test --workspace --lib --verbose`
- GitHub checks are green: `test`, `semantic-pull-request / build`, and CLA

Local note: the DB-backed integration test compiles with `--features integration-tests`, but I could not run it locally because port 5432 is currently served by a different Postgres container with incompatible credentials. CI ran the full test job successfully after the push.

Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover cleanup. The RPC keeps the existing encrypt/sign policy surfaces, minor-account containment remains fail-closed, the duplicated slot helper/new doc/test gaps are fixed, and CI is green.

PR #320, opened 2026-07-25
fix(signer): enforce authorization expiry on cached NIP-46 handlers

realmeylisdev:
## Summary

- `Nip46Handler.status` was a snapshot taken at load time, and the handler cache has **no TTL**. A handler cached while its authorization was still valid kept a stale `Active` status indefinitely — it went on signing, encrypting, and decrypting past its `expires_at` for as long as the process lived.
- Only the cache-miss path in `handle_nip46_request` computed expiry, so whether `expires_at` was honoured depended entirely on whether the handler happened to be cached.
- `load_single_authorization` compounded it: both branches hard-coded `HandlerStatus::Active`, and the team branch never even selected `expires_at`. Team authorizations are hard-deleted rather than revoked, so expiry is their only tombstone signal.
- Fix: carry `expires_at` on the handler and re-evaluate it against the clock per request via `effective_status()`. Revocation still outranks expiry. Duplicated expiry arithmetic is folded into `compute_status_from_expiry`.

## Motivation

The tombstone mechanism exists so a revoked or expired authorization returns a clean error instead of a silent timeout (see the `HandlerStatus` doc comment). Expiry was falling outside that guarantee whenever the handler was warm.

This matters most for shared operator identities, where bunker URLs are handed out per person and `expires_at` is the mechanism meant to time that access out. Under the old behaviour a bunker URL that had formally expired kept full signing and NIP-44 encrypt/decrypt authority over the shared identity until the signer process restarted.

Found while investigating divinevideo/divine-moderation-service#169's neighbouring area; it is independent of that issue and stands on its own.

## Related Issue

- Closes #

No existing issue — filing this as the record. Happy to open one and link it if you'd prefer that order.

## Testing

- [x] `cargo test -p keycast_signer` — 26 passed, 0 failed
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] Manual verification completed — see below
- [ ] `cargo test --workspace --verbose` — **not run**; the workspace suites behind `integration-tests` need `DATABASE_URL` and `TEST_REDIS_URL`, which I don't have locally. They compile clean under the clippy `--all-features` run above. Please confirm in CI.

New regression coverage in `signer/tests/authorization_expiry_tests.rs` (5 tests, no database required — `connect_lazy` builds the pool without opening a connection, matching the in-memory state the bug lived in):

- an expired authorization cached as `Active` is a tombstone and returns `"Authorization has expired"`
- a future `expires_at` stays active
- a null `expires_at` stays active
- revoked-and-expired reports revoked, not expired
- expiry takes effect without reloading the handler

Manual verification that the tests actually catch the defect: temporarily reverting `effective_status()` to return `self.status` fails exactly the two regression tests (`expired_authorization_cached_as_active_is_a_tombstone`, `expiry_takes_effect_without_reloading_the_handler`) and leaves the three guard tests passing. Restored before committing.

## Visuals

- [x] No visual change

### DIFF
--- a/signer/src/lib.rs
+++ b/signer/src/lib.rs
@@ -7,5 +7,5 @@ pub mod work_queue;
 
 // Re-export main types for convenience
 pub use error::{SignerError, SignerResult};
-pub use signer_daemon::{Nip46Handler, UnifiedSigner};
+pub use signer_daemon::{HandlerStatus, Nip46Handler, UnifiedSigner};
 pub use work_queue::{Nip46RpcItem, RelayQueue, RelaySender};
--- a/signer/src/signer_daemon.rs
+++ b/signer/src/signer_daemon.rs
@@ -57,9 +57,17 @@ pub struct Nip46Handler {
     is_oauth: bool,
     pool: PgPool,
     /// Handler status for tombstone support (Active, Revoked, or Expired)
+    ///
+    /// This is a snapshot taken when the handler was loaded. Time-based expiry is
+    /// re-evaluated per request from `expires_at` — see [`Self::effective_status`].
     status: HandlerStatus,
     /// When this handler became a tombstone (for cleanup after 24h)
     tombstone_at: Option<DateTime<Utc>>,
+    /// The authorization's `expires_at`, carried so expiry can be re-evaluated
+    /// against the clock on every request rather than only at load time. The
+    /// handler cache has no TTL, so a handler cached while valid would otherwise
+    /// keep signing, encrypting, and decrypting forever past its expiry.
+    expires_at: Option<DateTime<Utc>>,
 }
 
 impl Nip46Handler {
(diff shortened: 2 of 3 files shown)


dcadenas (approved):
Takeover handback for the current head.

The takeover closes the cached-authorization expiry gaps before signing and encryption key use, including rechecks after awaited validation, and ensures clock-expired handlers age out through the existing tombstone cleanup.

Focused signer coverage, workspace checks, the repository pre-push gate, and current-head CI passed.

Remaining uncertainty: the standalone local full-test bootstrap could not create its Docker network because the local address pools were exhausted; the repository gate and current-head CI still completed successfully.

Please review the takeover changes and merge when ready, or follow up with any questions.

PR #322, opened 2026-07-26
fix(atproto): reconcile app-claimed Divine usernames

### LINKED ISSUE #321: fix(atproto): reconcile app-claimed Divine usernames for crosspost enable
## Problem

Divine mobile users can claim a username directly through divine-name-server. That makes NIP-05 resolve correctly, but keycast never learns the username, so GET /api/user/atproto/status returns username: null and PUT /api/account/{pubkey}/crosspost rejects enable with 400 "Username must be claimed before enabling ATProto".

This is the backend half of divinevideo/divine-mobile#6357. The mobile fix can stop the repeated setup prompt, but keycast still needs to treat divine-name-server as authoritative or sync its local users.username column before Bluesky crosspost enable will work for app-claimed handles.

## Evidence

- keycast ATProto status and crosspost enable read UserRepository::get_username from users.username.
- Divine mobile claims usernames against https://names.divine.video/api/username/claim through ProfileRepository.claimUsername.
- divine-name-server exposes GET /api/username/by-pubkey/{pubkey}, which can confirm the active handle, but keycast does not currently fall back to it.
- keycast update_profile checks divine-name-server availability, but that endpoint path is not used by mobile's NIP-05 flow.

## Acceptance criteria

- A user with an active divine-name-server username but null keycast users.username can enable ATProto crossposting.
- GET /api/user/atproto/status reports the resolved username for that user.
- The resolved username is persisted or otherwise consistently available so the provisioning path receives the username.
- update_profile allows a same-owner re-submit instead of treating the user's own active name as unavailable.
- Add regression coverage for app-claimed username fallback and same-owner profile update.

NotThatKindOfDrLiz:
## Summary

- Reconcile app-claimed Divine usernames into `users.username` when ATProto status/enable finds a null local username.
- Preserve Divine name-server owner metadata for availability checks and allow same-owner active usernames through profile update.
- Add a shared timeout-bearing Divine name-server HTTP client and regression coverage for fallback, conflict, lookup failure, crosspost enable, and same-owner profile update.

## Motivation

- Mobile can claim usernames directly against Divine name-server, leaving Keycast's local username null.
- That made ATProto status return `username: null` and blocked crosspost enable with "Username must be claimed before enabling ATProto".
- The profile workaround also failed because Keycast treated the user's own active Divine username as unavailable.
- The ATProto status path now performs idempotent write-on-read reconciliation from Divine name-server when the local username is missing.

## Related Issue

- Closes #321
- Related: divinevideo/divine-mobile#6357

## Testing

- [x] `cargo test --workspace --no-run`
- [x] `cargo test -p keycast_api --lib`
- [ ] `cd api && cargo test --test atproto_http_test --test update_profile_divine_name_test` attempted locally; compilation succeeded, but runtime DB-backed tests could not connect because Docker/Postgres is not running on this machine (`docker ps` cannot connect to Docker daemon; targeted tests timed out connecting to `keycast_test`).

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/atproto.rs
+++ b/api/src/api/http/atproto.rs
@@ -48,12 +48,21 @@ pub struct AtprotoStatusResponse {
     pub username: Option<String>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum UsernameResolution {
+    Resolved(String),
+    NotClaimed,
+    Unavailable,
+}
+
 #[derive(Debug, thiserror::Error)]
 pub enum AtprotoControlError {
     #[error("user not found")]
     UserNotFound,
     #[error("username must be claimed before enabling atproto")]
     UsernameNotClaimed,
+    #[error("username resolution is temporarily unavailable")]
+    UsernameResolutionUnavailable,
     #[error("requested username does not match claimed username")]
     UsernameMismatch,
     #[error("{}", .0.public_message())]
@@ -89,7 +98,7 @@ pub async fn resolve_username_with_fallback<F, Fut>(
     tenant_id: i64,
     user_pubkey: &str,
     lookup: F,
-) -> Result<Option<String>, AtprotoControlError>
+) -> Result<UsernameResolution, AtprotoControlError>
 where
     F: FnOnce(String) -> Fut,
     Fut: Future<Output = Result<Option<PubkeyLookupResponse>, DivineNameError>>,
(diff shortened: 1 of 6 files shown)


dcadenas (approved):
App-claimed Divine usernames now reconcile in ATProto status and enable flows. Temporary lookup outages remain retryable, concurrent local claims cannot be overwritten by stale write-on-read results, and same-owner profile submissions remain accepted without weakening ownership checks.

Focused coverage, workspace checks, the pre-push gate, and current-head CI passed.

Please review the takeover changes and follow up with any questions.

PR #326, opened 2026-07-31
feat(auth): rate-limit, audit, and code the raw-key export endpoint

### LINKED ISSUE #325: feat(auth): rate-limit, audit, and code the raw-key export endpoint
## Context

`POST /api/user/export-key` is the one endpoint that hands a caller an account's
raw `nsec`. It is now reachable natively from divine-mobile (divinevideo/divine-mobile#6501),
where the Key Management screen fetches the key with the session bearer token
plus the account password, so this endpoint's hardening is no longer only a
web-surface question.

Review on that PR (divinevideo/divine-mobile#6501, @NotThatKindOfDrLiz) asked for the same level of
server-side clarity the earlier BYOK custody work got in #242 and
divinevideo/divine-mobile#4665: raw-key egress is an intended product contract, but it needs
explicit refusal semantics, rate limiting, and audit logging around it. Three
gaps against `origin/main` (`api/src/api/http/auth.rs:4046`, routed
`routes.rs:159`):

## 1. No rate limit or lockout

`export_key` verifies the password with `bcrypt::verify` and returns
`AuthError::InvalidCredentials` on failure, with nothing counting attempts —
no per-account lockout, no per-IP limit, no backoff. An attacker with a valid
bearer token (a borrowed unlocked phone, an exfiltrated session) can guess the
account password against an endpoint that pays out the private key on success.

The mobile client caps its own retries at five per sheet, but that is friction
in one client, not a control: the cap resets when the sheet reopens, and it does
not exist for any other caller of the HTTP API.

## 2. No audit record on success

The refusal path logs — `refuse_key_egress_for_verified_minor` emits
`event = "minor_key_egress_denied"`. A *successful* export writes nothing, so
there is no record that an account's raw key left the service, when, or from
where. `record_auth_event_and_log`
(`api/src/api/http/auth_observability.rs:115`) is already available and
infallible at the call site.

This is the one operation where the absence of a record is most costly: after a
key compromise there is no way to tell whether the key was exported, or how
often.

## 3. The two 403s are indistinguishable to a client

`export_key` can return 403 for two unrelated reasons:

- `AuthError::EmailNotVerified` — recoverable, the user verifies their email.
- `AuthError::Forbidden("Operation denied by policy")` — the `verified_minor`
  custody refusal, deliberately uniform so it leaks no account state.

Neither carries a machine-readable discriminator, so a client that wants to show
the right next step has to match on the English prose in `{"error": ...}` —
wording this service owns and can reword or translate per deployment. divine-mobile
shipped exactly that and has now replaced it with a second round trip to
`GET /user/account` to read `email_verified` / `verified_minor` as fields.

`coded_error_response` (`auth.rs:492`) already exists and is used for
`INVALID_EMAIL` and `EMAIL_ALREADY_EXISTS`. Extending it to these two 403s would
let clients branch on a code and drop the extra request.

## Proposal

- Rate-limit `export_key` per account and per IP, with a lockout after repeated
  password failures. Return 429 with `Retry-After` — `AuthError::ServiceUnavailable`
  already demonstrates the `Retry-After` shape.
- Emit an auth event on successful export via `record_auth_event_and_log`, and
  keep the existing denial warn.
- Give both 403 branches a `code` through `coded_error_response`, without
  changing the uniform human message on the policy denial.

## Notes

Response bodies for this endpoint must stay out of request/response logging —
the 200 body is the key itself.

Related: #242, divinevideo/divine-mobile#3359, divinevideo/divine-mobile#4665,
divinevideo/divine-mobile#6501.

realmeylisdev:
## Summary

- Protects `export-key` and `change-key` with one distributed Redis budget: five failed passwords per current tenant/pubkey over an exact 15-minute sliding window.
- Uses one atomic Redis script and owner-token leases so concurrent requests cannot over-admit or finalize another request's reservation.
- Runs password verification behind a bounded CPU permit and outside database transactions.
- Requires durable audit commits before returning raw key material, and commits change-key custody rotation and its success audit in one database transaction.
- Adds focused coverage for concurrency, pool pressure, Redis failure and expiry behavior, audit-write failure, change-key rollback, sliding-window boundaries, and policy gates.

## Motivation

The previous database limiter serialized admission, but held a transaction while the handler acquired another pooled connection and performed bcrypt and KMS work. Under ordinary pool pressure, one key-egress request could retain the last free connection while waiting up to 60 seconds for another, starving unrelated work on that instance.

Redis keeps refusal-heavy traffic off the database and makes admission atomic across Cloud Run instances. Its 15-minute limiter state is intentionally volatile: a Redis data restart resets the current budget after recovery, while Redis command failures return `503`. An abandoned request can retain its lease for at most 60 seconds; five abandoned leases can temporarily refuse that tenant/pubkey for that bounded interval.

The budget is scoped to the current pubkey, so a successful key rotation starts a fresh budget. Repeated rate-limit denials use metrics and tracing instead of writing one durable row per refusal; the fifth durable wrong-password audit event still records entry into lockout.

This also corrects earlier description claims: not every extractor or internal-error path writes an audit row, Redis-refused requests intentionally avoid Postgres, and `change-key` does not enforce `email_verified` even though `export-key` does. That difference predates this PR and is unchanged here.

## Related Issue

- Closes #325
- Public-login brute-force hardening remains separate in #328

## Testing

- [x] `cargo test -p keycast_api --features integration-tests --lib key_egress -- --nocapture`
- [x] `cargo test -p keycast_api --features integration-tests --lib verifier -- --nocapture`
- [x] `cargo test -p keycast_api --features integration-tests --test key_egress_hardening_test -- --nocapture`
- [x] `cargo test -p keycast_api --features integration-tests --test verified_minor_key_egress_test -- --nocapture`
- [x] CI test, web, semantic-title, and CLA checks passed on the takeover head

## Visuals

- [x] No visual change


NotThatKindOfDrLiz:
@realmeylisdev I pushed one takeover commit, `4152015`, to close out the remaining control gaps.

What changed:
- Serialized raw-key egress attempt accounting with a Postgres transaction-scoped advisory lock per tenant/pubkey, held across the limit read, password verification outcome, and authoritative audit insert.
- Made key-egress audit writes fail closed for this endpoint pair by adding a result-returning observability path while preserving best-effort behavior for existing callers.
- Moved the limiter count and key-egress audit inserts onto the same transaction connection to avoid pool deadlocks under concurrent bursts.
- Audited the previously silent branches: missing password, invalid export format, invalid BYOK nsec, duplicate BYOK key, encryption failure, and change-key transaction failure.
- Fixed `Retry-After` for over-budget windows by basing it on the oldest failure in the active five-attempt budget.
- Filed #327 for `auth_events` retention now that the table is a lockout control input.

Verification:
- `cargo check -p keycast_api --features integration-tests`
- `cargo test -p keycast_api --features integration-tests --test key_egress_hardening_test --no-run`
- `cargo test -p keycast_core --features integration-tests --test auth_event_repository_test --no-run`
- `cargo test --workspace --all-features --no-run`
- GitHub checks are green: test, Web Tests, semantic PR, CLA.

Local DB-backed test execution was blocked because Docker/Postgres was unavailable in this environment, but CI ran the full test job successfully on the pushed commit. Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover commit 4152015. The lockout accounting is now serialized per account, key-egress audit writes are authoritative for the control paths, previously silent post-auth branches are audited, Retry-After handles over-budget windows, CI is green, and retention is tracked in #327. @realmeylisdev please review the takeover commit and merge when ready.

dcadenas (dismissed):
Takeover incoming. Leaving this to block merges

dcadenas:
I'm taking over after re-reviewing head `4152015`. That commit fixed the original over-admission bug and made successful export auditing authoritative. Two problems remain: an availability bug in the limiter, and two change-key correctness gaps.

**The availability bug: one request needs two database connections at once.**

The handler opens a transaction, then acquires a second pooled connection — via `find_with_password_and_verified`, `find_encrypted_key_for_tenant`, or a nested `pool.begin()` in `change_key_transaction`.

This needs no concurrent traffic to reproduce. With `max_connections(1)`, one wrong-password export waits the full acquire timeout and returns a server error. With two connections it returns in milliseconds.

The pool is per instance and shared with every endpoint. So if one connection is free, key egress takes it, then waits for another while still holding the first. Requests can fail, and they can block login, OAuth and the signer for the 60-second acquire timeout.

The advisory lock is not the cause. It only accelerates occupancy, because waiters hold their connections. The existing burst-of-10 test misses this: it sits exactly on the default pool boundary, and pre-transaction work staggers acquisition. Burst 11 fails.

**The fix: move the limiter to Redis.**

One atomic script reserves against the current tenant/pubkey budget using Redis server time. It keeps the exact 15-minute failure window plus short-lived owner-token leases.

Requests already over budget stop early — before password verification and before any user or key lookup. A wrong password converts its own lease into a counted failure. Every other outcome releases it. Reserved work is bounded by a deadline below the lease lifetime, so a lease cannot expire while its request is still running. Redis command failures return 503.

**The tradeoff.** A Redis data restart resets the volatile budget after recovery. That bounded reset is accepted. Redis being unavailable fails key egress closed. `#325` requires distributed throttling, not restart-durable state.

**What Postgres still guarantees.** Admitted password outcomes stay durably audited. No raw key and no custody mutation is delivered unless its audit commit succeeds. Change-key rotation and its success audit now commit together.

**Other fixes.** Password verification runs behind a bounded CPU permit. Under verifier saturation a caller may get a load-shed 503 before the limiter can return 429 — intentional and cheap. Change-key builds its Origin and UCAN inputs before mutating custody, eliminating the post-commit failure paths. Repeated denials use metrics and tracing instead of a durable row each; the fifth wrong-password record still proves lockout entry.

**One user-visible effect.** In-flight attempts hold a lease. So a client that abandons five or more export attempts mid-request can be refused for up to 60 seconds without submitting a wrong password. It is bounded, self-inflicted, reported accurately by `Retry-After`, and fails safe.

**Description corrections.** The old description overstated audit coverage — extractor and internal-error paths don't all write rows, and Redis-refused requests intentionally skip Postgres. Change-key doesn't enforce `email_verified`; that predates this PR, so "equivalent coverage" is corrected. Claims that the budget survives restarts, and that `auth_events` retention or indexing affects limiter correctness, are removed.

Public-login brute-force protection stays separate, tracked in `#328`.

dcadenas (approved):
Approving the takeover state at `faed210`. Handing back to @realmeylisdev to review and merge.

The detail is in the takeover comment above. In short: the limiter moved off Postgres because the handler required two simultaneous pool connections, which stalls a single request for the full 60-second acquire timeout when fewer than two are free and starves other endpoints on the same instance while it waits.

**Validation.** CI is green on this head. Beyond that, the fix was verified independently of the session that wrote it: a one-connection pool now completes both a successful export and a change-key end to end, where the previous head timed out and returned 5xx. The same probe was run under injected pool pressure at two pool sizes, and a burst of 60 against a pool of 10 returns clean where the previous head took 60 seconds with most requests failing. Redis-absent and Redis-slow paths were both exercised: absent fails closed with a 503 rather than returning key material, slow degrades latency without corrupting the budget.

**Remaining uncertainty, stated rather than buried.** These runs were performed by tooling under my direction, not by hand. The Redis restart tradeoff is deliberate and accepted: losing limiter state grants an identity a fresh five attempts, and a later state-loss event can do so again. And a client that abandons five or more in-flight export attempts can be refused for up to 60 seconds without having submitted a wrong password — bounded, self-inflicted, and reported accurately by `Retry-After`.

The earlier approval on this pull request predates the takeover and applies to `4152015`, which this replaced. Worth a fresh look at the diff rather than relying on it.

PR #331, opened 2026-08-01
fix(auth): preserve first-party deletion on refresh

### LINKED ISSUE #323: fix(auth): account deletion 403s for every user whose token has been refreshed (first_party fact not preserved)
## Summary

`DELETE /user/account` requires a UCAN that is user-signed **or** carries the `first_party: true` fact (`api/src/api/http/auth.rs:4288`). The refresh grant does not preserve that fact, and divine-mobile refreshes immediately before deleting — so any returning divineOAuth user is denied with **403**.

This is distinct from #296 (claim-token FK, which produces a 503) and accounts for the large majority of failures.

## Evidence

The limitation is acknowledged in-tree:

- `api/src/api/http/oauth.rs:2666` — `// Note: refresh tokens don't preserve first_party status - users need fresh headless login`
- `api/src/api/http/oauth.rs:2682` — `false, // Refresh tokens are not first-party`
- `api/src/api/http/oauth.rs:3135` — `is_headless, // first_party fact for headless flow tokens`, minted only on the original code exchange

Client side, `mobile/lib/services/auth_service.dart:2886` calls `getSessionOrRefresh()` immediately before `deleteAccount`, and access tokens expire in roughly a day — so the client reliably triggers the refresh that strips its own authorization.

Reproduced against local keycast — same account, two tokens, one call:

| token | UCAN `fct` | `DELETE /api/user/account` |
|---|---|---|
| original (register → login → code exchange) | `"first_party":true` present | 200 |
| after one `grant_type=refresh_token` | `first_party` absent | 403 |

## Production impact

Cloud Logging, project `openvine-co`, service `keycast`, 30 days to 2026-07-26:

| event | count | distinct users |
|---|---|---|
| `account_deletion_denied` | **275** | **57** |
| `account_deletion_started` | 71 | — |
| `account_deletion_completed` | 63 | 63 |
| `account_deletion_failed` | 8 | 1 |

- **48 distinct users were denied and never completed.** Of the 111 distinct users who attempted deletion, 43% never got through.
- **All 71** `account_deletion_started` events carry `is_user_signed=False, is_first_party=True` — no exceptions, and zero user-signed deletions. Success ⟺ the token has not been refreshed.
- 9 users were denied and later completed, consistent with re-login re-minting the fact.
- Denials per user reach 32, 18, 12, 12 and 10 — users retry because the client tells them to check their connection.
- Most recent denial: **2026-07-26 07:13:52Z**.

Present since `f1bea21` (2026-01-16), which introduced the `first_party` gate and the refresh-path comment in the same commit.

## Why this is worse than a plain 403

divine-mobile publishes the irreversible NIP-62 request to vanish **before** calling this endpoint, so a denial lands after the user's content has already been broadcast for deletion. Content destroyed, account alive, user still signed in. Client-side context is in divinevideo/divine-mobile#4881 and divinevideo/divine-mobile#6335.

## Suggested directions

1. **Accept NIP-98 proof-of-key on this endpoint.** `api/src/nip98.rs:82` `extract_and_validate(auth_header, expected_url, expected_method)` already verifies kind 27235, the Schnorr signature, a 60-second window, and `u`/`method` binding — and is already wired into a POST route at `auth.rs:788`. divine-mobile already has the matching minting side (`nip98_auth_service.dart:82`, which omits the payload hash for bodyless requests). This binds authorization to key control rather than token age, and the endpoint's own documentation already treats user-signed proof as valid.
2. **Preserve `first_party` across refresh**, or mint a narrowly-scoped deletion capability.
3. **Add a side-effect-free preflight** (`GET /user/account/deletable`) so a client can verify deletability *before* publishing anything irreversible. This matters specifically because deleting the account cascade-destroys `personal_keys`, so a client cannot safely delete first and publish after.

## Adjacent finding: #201's 503 narrowing did not reach this endpoint

`auth.rs:4318-4326` still collapses `NotFound`, FK-integrity violations, and genuine database failures into `AuthError::Database(sqlx::Error::Protocol(...))` → **503 "Service temporarily unavailable. Please try again in a few minutes."** A client therefore cannot distinguish "already deleted" from "server broken", so a retry can never be made idempotent, and the #296 cohort is told to retry something that can never succeed. #201 was closed as completed but covered other routes.

## Adjacent finding: `delete_account` never invalidates `http_handler_cache`

There is no `.invalidate()` call site in production code at all — only the fast-path get/insert in `nostr_rpc.rs` and tests. Post-delete signing still refuses, because `check_user_status_active` runs in the caller for every RPC method except `get_public_key`, so this appears latent rather than exploitable. Flagging it as a loose end.

realmeylisdev:
## Summary

- Preserve first-party account deletion authorization across OAuth refresh-token rotation.
- Add an `oauth_authorizations.is_first_party` flag so the refresh grant can mint replacement UCANs with `first_party: true` only for authorizations that originally came from a first-party flow.
- Gate `/api/headless/authorize` so a third-party OAuth bearer token cannot mint a first-party authorization.
- Preserve first-party status through silent OAuth re-authentication when replacing an existing first-party authorization.
- Keep `DELETE /user/account` on the existing UCAN authorization path; it no longer gains a NIP-98 bypass.
- Stop truncating account-deletion pubkeys in structured logs.
- Add integration coverage for refresh rotation, headless authorize gating, and silent re-auth first-party preservation.

## Motivation

Authorization for account deletion depends on a UCAN fact (`first_party`) that was previously present only on the original headless-flow access token. Clients refresh immediately before deletion, so a returning first-party user could lose the fact and be denied after the client had already published an irreversible NIP-62 vanish.

The safer fix is to persist first-party status on the OAuth authorization and preserve it during refresh-token rotation while keeping third-party OAuth apps outside the account-deletion boundary.

## Related Issue

- Closes #323
- Client-side context: divinevideo/divine-mobile#4881, divinevideo/divine-mobile#5756

## Review notes

- This remains an authorization change in a security-critical path and should get a human security/platform review.
- Existing authorizations created before this migration have `is_first_party = false`; users with old sessions may need to re-authenticate once to get a first-party authorization carrying the new persisted flag. There is no safe server-side backfill, because `client_id` and `redirect_uri` are caller-supplied at `/api/headless/login`.
- `first_party` now persists for the life of eligible first-party authorizations rather than for a single access token, so deletion authority is renewable via refresh where it previously expired with the access token.
- Deploy ordering is safe with the current mobile fallback to bearer-token deletion, and `cloudbuild.yaml` runs migrations before the deploy step.

## Testing

- [x] `cargo check -p keycast_api --tests --features integration-tests`
- [x] `cargo test -p keycast_api server_signed_ucan_includes_first_party_only_when_requested`
- [x] Local DB-backed targeted tests were attempted but could not run because Docker/Postgres is unavailable here (`PoolTimedOut`):
  - `cargo test -p keycast_api --features integration-tests --test headless_auth_test headless_authorize`
  - `cargo test -p keycast_api --features integration-tests --test oauth_refresh_first_party_test`
  - `cargo test -p keycast_core --features integration-tests --lib repositories::oauth_authorization`
- [x] GitHub CI DB-backed checks passed on commit `40ae627`.

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -1501,6 +1501,7 @@ pub async fn create_bunker(
             secret_hash,
             relays: relays_json.clone(),
             policy_id,
+            is_first_party: false,
             client_pubkey: None,
             authorization_handle: None,
             handle_expires_at,
@@ -4903,7 +4904,7 @@ pub async fn delete_account(
         tracing::warn!(
             event = "account_deletion_denied",
             tenant_id = tenant_id,
-            user = &user_pubkey[..8],
+            user_pubkey = %user_pubkey,
             redirect_origin = %redirect_origin,
             "Denied: not user-signed and not first-party"
         );
@@ -4916,7 +4917,7 @@ pub async fn delete_account(
     tracing::info!(
         event = "account_deletion_started",
         tenant_id = tenant_id,
-        user = &user_pubkey[..8],
+        user_pubkey = %user_pubkey,
         is_user_signed = is_user_signed,
         is_first_party = is_first_party,
         redirect_origin = %redirect_origin,
@@ -4932,7 +4933,7 @@ pub async fn delete_account(
             tracing::error!(
                 event = "account_deletion_failed",
                 tenant_id = tenant_id,
-                user = &user_pubkey[..8],
+                user_pubkey = %user_pubkey,
                 error = %e,
                 "Database error during deletion"
             );
@@ -4960,7 +4961,7 @@ pub async fn delete_account(
     tracing::info!(
         event = "account_deletion_completed",
         tenant_id = tenant_id,
-        user = &user_pubkey[..8],
+        user_pubkey = %user_pubkey,
         teams_removed = result.teams_removed,
         oauth_auths_deleted = result.oauth_authorizations_deleted,
         bunkers_notified = result.bunker_pubkeys.len(),
@@ -4975,9 +4976,63 @@ pub async fn delete_account(
 
 #[cfg(test)]
 mod tests {
+    use super::generate_server_signed_ucan;
     use super::validate_origin;
     use super::AccountStatusResponse;
     use axum::response::IntoResponse;
+    use ucan::Ucan;
+
+    #[tokio::test]
+    async fn server_signed_ucan_includes_first_party_only_when_requested() {
+        let user_keys = Keys::generate();
+        let server_keys = Keys::generate();
+
+        let first_party_token = generate_server_signed_ucan(
+            &user_keys.public_key(),
+            1,
+            "user@example.test",
+            "https://first-party.example.test",
+            None,
+            &server_keys,
+            true,
+            None,
+            None,
+        )
+        .await
+        .expect("first-party UCAN");
+        let first_party_ucan =
+            Ucan::try_from_token_string(&first_party_token).expect("decode first-party UCAN");
+        assert!(
+            first_party_ucan
+                .facts()
+                .iter()
+                .any(|fact| fact.get("first_party").and_then(|v| v.as_bool()) == Some(true)),
+            "first-party sessions must carry the deletion authorization fact"
+        );
+
+        let third_party_token = generate_server_signed_ucan(
+            &user_keys.public_key(),
+            1,
+            "user@example.test",
+            "https://third-party.example.test",
+            None,
+            &server_keys,
+            false,
+            None,
+            None,
+        )
+        .await
+        .expect("third-party UCAN");
+        let third_party_ucan =
+            Ucan::try_from_token_string(&third_party_token).expect("decode third-party UCAN");
+        assert!(
+            third_party_ucan
+                .facts()
+                .iter()
+                .all(|fact| fact.get("first_party").is_none()),
+            "third-party sessions must not gain account deletion authority"
+        );
+    }
 
     #[cfg(feature = "integration-tests")]
     #[tokio::test]
--- a/api/src/api/http/headless.rs
+++ b/api/src/api/http/headless.rs
@@ -561,10 +561,7 @@ pub async fn headless_authorize(
     let pool = &auth_state.state.db;
     let tenant_id = tenant.0.id;
 
-    // Extract user from UCAN Bearer token with tenant validation
-    let user_pubkey = super::auth::extract_user_from_token(&headers, tenant_id)
-        .await
-        .map_err(|_| HeadlessError::Unauthorized)?;
+    let user_pubkey = extract_first_party_or_user_signed_user(&headers, tenant_id).await?;
 
     tracing::info!(
         event = "headless_authorize",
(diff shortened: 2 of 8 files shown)


NotThatKindOfDrLiz:
@/private/tmp/pr331-comment.md

NotThatKindOfDrLiz:
@realmeylisdev I took over the branch and replaced the NIP-98 deletion path with the safer fix we discussed:

- `DELETE /user/account` stays on the existing UCAN authorization path.
- OAuth code exchange now stores whether the authorization came from the first-party headless flow.
- Refresh-token rotation preserves `first_party: true` only for those persisted first-party authorizations.
- Third-party apps, including apps with signing access, do not gain deletion authority.
- Account-deletion structured logs now emit the full pubkey instead of a truncated prefix.

Verification:

- CI is green on `50d9e74`.
- Local: `cargo fmt --all -- --check`
- Local: `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- Local: `cargo check --workspace --all-targets --all-features`
- Local: `cargo test -p keycast_api server_signed_ucan_includes_first_party_only_when_requested`

Local DB-backed tests could not run here because Docker/Postgres was unavailable (`PoolTimedOut`), but the GitHub `test` job passed with the repo test stack.

Please review the takeover changes and merge when you are comfortable with them.

NotThatKindOfDrLiz (approved):
Approved after takeover. The PR now preserves first-party deletion authority through refresh-token rotation without adding a NIP-98 deletion bypass, and CI is green on the current head.

realmeylisdev:
@NotThatKindOfDrLiz Reviewed the takeover — the persisted-flag approach is the right shape and I'm happy with it. I traced every site it touches: both explicit-column selects into `OAuthAuthorization` were updated (the signer's two use `SELECT *`), all three `create` call sites set the flag and the compiler enforces it, and the other `generate_server_signed_ucan` call sites still pass `false`. The migration sorts newest, is additive, and `cloudbuild.yaml` runs `run-migrations` before the deploy step.

I pushed one test-only commit on top (b78b5c8), so CI will re-run against a new head.

**Why:** the refresh grant is the path this PR actually fixes, but nothing exercised it. `server_signed_ucan_includes_first_party_only_when_requested` covers the UCAN builder and `test_create_and_find_preserves_first_party_flag` covers the column — both still pass if `handle_refresh_token_grant_inner` stops reading `oauth_auth.is_first_party`.

`api/tests/oauth_refresh_first_party_test.rs` drives the real `/oauth/token` route with `grant_type=refresh_token` against a seeded authorization and asserts the `first_party` fact on the rotated access token: present for a first-party authorization, absent for a third-party one. I confirmed the positive case fails when the grant is reverted to a hardcoded `false`, then restored the fix and re-ran.

Docker was available here, so I also ran the DB-backed tests that were blocked on your side:

- `cargo test -p keycast_core --features integration-tests --lib repositories::oauth_authorization` — 5 passed, including `test_create_and_find_preserves_first_party_flag`
- `cargo test -p keycast_api --features integration-tests --test oauth_refresh_first_party_test` — 2 passed
- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated` clean

Two things I'd like your read on before I merge:

1. Pre-migration authorizations all get `is_first_party = false`, and I don't see a safe server-side backfill — `client_id` and `redirect_uri` are caller-supplied at `/api/headless/login` with no allowlist, so there's no reliable signal for which existing rows were first-party. Existing mobile sessions lose deletion authority until the user logs out and back in. Does the bearer-token fallback cover that gap in practice, or do we want a re-auth prompt?

2. `first_party` now means "presented email+password to `/api/headless/*`" for the life of the authorization rather than for one access token. `headless_login` does no registered-client validation and `authorize_get` documents open OAuth, so a stolen refresh token can renew deletion authority for as long as the authorization lives, where previously it could never obtain it. That widening is inherent to the fix rather than a flaw in it, but I'd rather it be an explicit accept than a side effect.

NotThatKindOfDrLiz:
@realmeylisdev I took over the remaining auth-boundary issues and pushed two commits on top of the branch:

- `9bd2874` gates `/api/headless/authorize` so a third-party server-signed OAuth bearer cannot mint a first-party authorization, and preserves first-party status through handle/origin silent re-auth replacement paths.
- `40ae627` fixes the new re-auth integration test setup so the authorization-code exchange has a seeded secret pool.

Verification:

- Local: `cargo fmt --all -- --check`
- Local: `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- Local: `cargo check -p keycast_api --tests --features integration-tests`
- Local: `cargo test -p keycast_api server_signed_ucan_includes_first_party_only_when_requested`
- Local DB-backed targeted tests were attempted here, but Docker/Postgres is unavailable and they timed out at pool setup.
- GitHub CI is green on `40ae627` (`test`, `Web Tests`, semantic PR, CLA).

Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Final pass is clean after the takeover commits.

The first-party authorization boundary now matches the PR description: third-party OAuth bearers cannot use `/api/headless/authorize` to mint durable deletion authority, refresh rotation preserves only persisted first-party authorizations, and silent re-auth replacement keeps the first-party flag from the authorization it replaces. GitHub CI is green on the current head.

NotThatKindOfDrLiz:
@realmeylisdev one more merge-gate reminder: because this is a security-sensitive Keycast auth change, the PR should ultimately be approved for merge by @dcadenas before it lands.

Issue #276, opened 2026-07-01
Correct deployment documentation for current Cloud Run and GKE paths


NotThatKindOfDrLiz:
## Summary

Rewrite deployment documentation to reflect Keycast's current production and non-production deployment paths.

## Motivation

The existing deployment documentation incorrectly describes production as already served from GKE/ArgoCD. Production currently runs on Cloud Run, while staging, POC, and test use the GKE workflow. Incorrect deployment docs can mislead incident response and release work.

## Acceptance Criteria

- `docs/DEPLOYMENT.md` accurately describes Cloud Run production deployment.
- The document separately describes GKE/ArgoCD deployment for non-production environments.
- Known migration gaps are called out clearly.
- Log queries, migration notes, and environment tables match the current infrastructure state.

PR #332, opened 2026-08-01
docs: record deployment source of truth and transaction-mode pooling constraints

dcadenas:
## Summary

Two things a future reader or agent currently has to rediscover the hard way.

First, where production actually runs. Keycast is mid-migration, and a resource named `prod` or `production` is not evidence that it serves production traffic. This has already caused one documented correction (#276), which fixed `docs/DEPLOYMENT.md` but never reached the agent-facing entry point.

Second, what transaction-mode connection pooling forbids when a Keycast environment is configured behind a pooler. The backend serving a connection can change between transactions, so code and operational docs need to avoid session-continuity assumptions.

## What Changed

`AGENTS.md` gains a pointer telling readers to check `docs/DEPLOYMENT.md` before doing anything that depends on where production runs, with enough of the answer inline that someone who skips the pointer still does not get it wrong.

`AGENTS.md` also gains a short `Transaction-Mode Pooling` subsection under `Database And Migrations` covering what breaks: session-level `SET`, `LISTEN`/`NOTIFY`, `WITH HOLD` cursors, session-scoped advisory locks, and streaming results across a transaction boundary. It records why `SQLX_STATEMENT_CACHE` is safe only when poolers set `max_prepared_statements`, and why these failures are load-dependent.

The migration guidance now matches the current deploy paths: serving instances do not run migrations on startup; dedicated Cloud Run and GKE jobs execute `./keycast --migrate`, which runs the embedded `sqlx::migrate!` migrations. Migration database URLs must bypass any transaction-mode pooler because SQLx migrations use a session-scoped advisory lock.

The takeover commit also removes package scripts that pointed at deleted production helper scripts and fixes the same stale migration command in an existing implementation plan.

## Testing

- `bun run test:deploy-preflight`
- `jq -e . package.json`
- `git diff --check origin/main...HEAD`
- `cargo test --workspace --lib`

Attempted `cargo test --workspace`, but the local Docker dependency stack was unavailable (`/Users/lizsw/.docker/run/docker.sock` missing), and the DB-backed integration tests failed with `PoolTimedOut`.

## Risks

No runtime behavior change. This updates docs, comments, and removes two broken package scripts for deleted production helpers.

## Follow-ups

`docs/MONITORING.md` has nothing on connection-pool exhaustion, and Cloud Run has no readiness probe configured, so the DB-aware `readyz` handler is unused there. Both are separate from this change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -114,17 +114,17 @@ When touching OAuth, auth/session behavior, NIP-05/profile behavior, signer flow
 
 - New schema changes live in `database/migrations/` as a new timestamped migration. Do not edit shipped migrations.
 - Use SQLx for queries so compile-time verification stays meaningful. If you change a query, regenerate `sqlx-data.json` (where applicable) and commit it with the source change.
-- Locally, `bun run db:reset` recreates the dev database; `bun run db:migrate` applies new migrations. Production migrations are run via `tools/run-migrations.sh` from the deploy pipeline.
+- Locally, `bun run db:reset` recreates the dev database; `bun run db:migrate` applies new migrations. Production migrations are run by the deploy path as a dedicated one-shot job before serving rollout: the `keycast-migrate` Cloud Run Job for Cloud Run and the `keycast-db-migrate` ArgoCD sync hook for GKE. Both execute `./keycast --migrate`, which runs the embedded `sqlx::migrate!` migrations.
 
 ### Transaction-Mode Pooling
 
-Every environment runs a transaction-mode connection pooler in front of Postgres. The backend serving a connection can change between transactions, which breaks anything that assumes session continuity.
+When Keycast runs behind a transaction-mode connection pooler, the backend serving a connection can change between transactions, which breaks anything that assumes session continuity. Local dev and CI connect directly to Postgres unless explicitly using the loadtest pooler harness.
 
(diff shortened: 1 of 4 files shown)


NotThatKindOfDrLiz:
@dcadenas I took this over and pushed `af821facf12c5df241bb3dc00cd8fc163dcebf40`.

What changed:

- Corrected the migration guidance to point at the current one-shot paths: Cloud Run `keycast-migrate` and GKE `keycast-db-migrate`, both executing `./keycast --migrate` / `sqlx::migrate!`.
- Added the missing invariant that migration DB URLs must bypass a transaction-mode pooler because SQLx migrations use a session-scoped advisory lock.
- Scoped the transaction-mode pooling warning to environments that actually run behind a pooler, with local dev and CI called out as direct Postgres unless the loadtest harness is used.
- Removed the deleted production helper scripts from `package.json` and fixed the same stale migration command in the self-serve email-change plan.
- Updated the PR body to match the new diff and verification.

Verification:

- `bun run test:deploy-preflight`
- `jq -e . package.json`
- `git diff --check origin/main...HEAD`
- `cargo test --workspace --lib`
- GitHub checks are green on the pushed head: `test`, `Web Tests`, `semantic-pull-request`, and CLA.

I also attempted `cargo test --workspace` locally, but Docker was not running (`/Users/lizsw/.docker/run/docker.sock` missing), so the DB-backed integration tests failed with `PoolTimedOut`. CI ran those successfully. Please review the takeover changes and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover. The migration and pooling invariants now match the current Cloud Run/GKE deploy paths, stale deleted-script references are removed from active guidance, and CI is green on the pushed head.

PR #333, opened 2026-08-01
test(core): detect nested pool acquisition with a one-connection probe

### LINKED ISSUE #330: test(core): detect nested pool acquisition with a one-connection probe
## Summary

Add an integration test that catches nested pool acquisition: a repository method that acquires its own connection while the caller already holds one.

## Why

One request needing two simultaneous pool connections stalls for the full `ACQUIRE_TIMEOUT_SECS` (60s) whenever fewer than two are free, and keeps holding its first connection while it waits, so it starves other endpoints on the same instance. Measured on a one-connection pool: 5s stall and a 5xx before the fix, ~4ms after.

`main` has no instances of this today. The point of the test is that nothing currently prevents one being introduced — which is how it reached `keycast#326`, on a codebase whose `AGENTS.md` already documented the invariant.

## Approach

`core/tests/pool_nesting_test.rs`. Build a pool with `max_connections(1)` and a short acquire timeout, run each transaction-owning repository method against it, and assert it completes. A path needing a second connection cannot get one, so it stalls and fails. The assertion is completion, not query result — a `NotFound` still passes, since the only question is whether a second connection was required.

Include a control test that deliberately nests and must be caught. Without it, a harness that stopped working looks identical to a clean run.

Coverage is a sample of the hold sites, not all of them, since each needs a fixture. Make that visible in the test output rather than only in a comment, so a green run cannot be read as a clean bill of health.

## Acceptance criteria

- The test fails on a deliberately nesting path and passes on current `main`.
- It runs in the existing `keycast_core --features integration-tests` CI step, with no new pipeline machinery.
- Handler invariants are documented in `AGENTS.md`.

## Not doing

Converting repository methods to take `&mut PgConnection` so the mistake is unrepresentable. That is the stronger fix and remains correct in principle, but the compiler only identifies the call sites, not what to do at each: handlers interleave database work with bcrypt, KMS and email sends, so every site needs a judgment about where the connection region begins and ends. Roughly 109 for `UserRepository` alone. Done carelessly it increases connection hold time, which is the thing that causes the outage. Not worth it against zero current defects.

dcadenas:
Closes #330

## Summary

Adds an integration test that catches a bug class nothing currently prevents: a repository method acquiring its own database connection while the caller already holds one.

A request that needs two connections at once stalls for the full 60-second acquire timeout whenever fewer than two are free, and keeps holding its first connection the whole time, so it starves every other endpoint on that instance. It is not an abuse case; ordinary load reaches that condition.

`main` has no instances of this today. The point is that nothing stopped one being introduced, which is how it reached #326 on a codebase whose guidance already documented the invariant.

## What Changed

`core/tests/pool_nesting_test.rs` builds a pool with `max_connections(1)` and runs selected transaction-owning repository paths against it. A path that needs a second connection cannot get one, so SQLx returns a pool timeout and the test fails with a nested-acquisition diagnosis.

The probe now asserts each sampled path reaches its expected success outcome, so fixtures cannot pass by returning early from `NotFound`, `Duplicate`, or token-not-consumable branches. The claim-token and pending-OAuth cases seed live rows that reach the mutation tails they are meant to protect.

The key-rotation sample covers `change_key_in_transaction`, the transaction-scoped method used by the real handler, rather than the unused convenience wrapper.

The coverage disclosure was updated to the reproducible 8-of-21 hold-site count and now asserts both the covered count and total count match the test name. Database cases are serialized and clean up committed test rows.

`package.json` now runs `keycast_core` integration test binaries under `bun run test` and `bun run test:ci` by adding `--tests` to the existing core command.

## Testing

- `cargo clippy -p keycast_core --all-features --tests -- -D warnings`
- `cargo test -p keycast_core --features integration-tests --lib --tests --no-run`

Attempted database execution with:

`DATABASE_URL=... cargo test -p keycast_core --features integration-tests --test pool_nesting_test -- --nocapture`

That compiled and started the test binary, but local execution could not complete because the Docker-backed Postgres stack was unavailable from this shell. The first sandboxed run failed with `Operation not permitted` on DB connect; the unsandboxed run then failed at pool setup with `PoolTimedOut`, and `docker ps` could not reach the Docker daemon.

Also attempted `cargo test --workspace --no-run`; it failed while compiling unrelated workspace crates with `No space left on device`, after which the worktree build artifacts were cleaned.

## Risks

Low. The production code path is unchanged. The behavior change is that the probe now fails on early-return fixtures instead of treating any non-timeout result as sufficient.

The probe samples 8 of the 21 functions that hold a connection, because each case needs a hand-written fixture. A green run is therefore not a clean bill of health for the whole workspace, and the file says so in both the module docs and a test name.

## Follow-ups

Making this unrepresentable rather than detectable, with repository methods taking caller-owned connection/transaction handles, remains the stronger fix. It stays out of scope here because those call sites need handler-by-handler judgment about connection hold time around bcrypt, KMS, email, and audit work.


NotThatKindOfDrLiz:
@dcadenas I took over the review fixes on this branch.

What changed:
- `package.json` now runs `keycast_core` integration test binaries with `--tests` in both `test` and `test:ci`, so this probe is covered by the documented local gate.
- `pool_nesting_test` now asserts each sampled path reaches its expected success path instead of accepting early-return errors.
- The claim-token and pending-OAuth fixtures now seed live rows that reach the mutation tails.
- The key-rotation sample now covers `change_key_in_transaction`, the transaction-scoped method used by the real handler.
- Pool-timeout detection is exact instead of a 2s wall-clock heuristic.
- Coverage disclosure is updated to 8 of 21 hold sites and guarded in the test name; committed test rows are cleaned up.

Verification:
- GitHub checks are green on `c67c435`.
- Local: `cargo fmt --all -- --check`
- Local: `cargo clippy -p keycast_core --all-features --tests -- -D warnings`
- Local: `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- Local: `cargo test -p keycast_core --features integration-tests --lib --tests --no-run`

Local database execution could not complete because Docker/Postgres was unavailable from this shell; CI did run the database-backed test successfully. Please review the takeover commit and merge when you're ready.

NotThatKindOfDrLiz (approved):
Takeover fixes address the review blockers. The probe now exercises success paths, the local test gate includes core integration test binaries, coverage disclosure is corrected, and CI is green on the updated head.

PR #334, opened 2026-08-04
fix(atproto): stop a failed enable from trapping crossposting on

hm21:
## Summary

- `enable_user_atproto_with_trigger` / `reenable_user_atproto_with_trigger` now roll the opt-in back to `enabled=false` when the control-plane trigger fails, instead of leaving `enabled=true, state='failed'`.
- `disable_user_atproto_with_trigger` completes the local disable when the trigger fails **and** the account has no DID. A provisioned account (DID present) still returns the error, unchanged.
- The lifecycle writes now carry the existing `atproto_did` forward instead of nulling it, so the DID that gates the disable above survives a failed enable.
- The web settings card keys its failure phase on the state alone, since a rolled-back enable no longer reports `enabled=true`.
- Tests: the existing enable-failure test now pins the rollback; new tests cover a provisioned enable failure (DID kept, disable still refused), the reenable rollback, and disabling a stuck unprovisioned account both directly and through `set_user_atproto_crosspost`.

## Motivation

`enable_user_atproto` persists `enabled=true, state='pending'` before it calls the control plane, and on trigger failure the old code wrote `enabled=true, state='failed'`. That leaves an opt-in that exists nowhere except that row — no gateway record, no repo, no DID — while `/api/user/atproto/status` reports crossposting as on.

The second half is what makes it a trap rather than a cosmetic wrong flag: the disable path calls the control plane *first* and only writes `disabled` after it succeeds. So while the control plane is unavailable, the same outage that refused the enable also refuses the disable, and the account stays switched on with no way for the user to switch it off.

This is reproducible in production right now. In the mobile client (`divinevideo/divine-mobile`, Settings → Bluesky Publishing) the toggle reads as on with status "Account provisioning failed", and turning it off returns the scoped 503 from #281 ("Bluesky publishing is temporarily unavailable"). Verified on a real Samsung Galaxy against production `login.divine.video`; the handle it advertises (`<username>.divine.video`) resolves nowhere — no `/.well-known/atproto-did`, no `_atproto` TXT record, `Profile not found` from the public Bluesky API.

Design notes for review:

- **Why roll back rather than keep the intent.** If the trigger never reached the control plane, the intent exists only here and is safe to drop; the user can retry. If it *did* land before the response failed, provisioning reports back through `internal_sync_atproto` and the state converges there — so the rollback is self-healing in that race, whereas today's stuck flag is not.
- **Why the disable escape hatch is gated on the DID.** Without a DID there is no repo to publish from, so honouring "off" locally cannot produce a false "off". With one, refusing is correct: a live account could keep publishing while the user believes it is off. That boundary is the existing `disable_trigger_failure_preserves_existing_state` test, which stays green.
- **Why the enable path stopped nulling the DID.** `set_atproto_state` writes `atproto_did` unconditionally, and every lifecycle write passed `None` — so the enable path erased the exact signal the gate above depends on. An enable against an already-provisioned account, or a control-plane sync landing mid-write, left `did=NULL` with a live repo behind it, and the next disable would then take the unprovisioned path. Carrying the DID through is safe: a stale DID authorizes nothing on its own (`ready_atproto_identity` also requires `state='ready'`), and provisioning overwrites it through the internal sync.
- **Known limit of the escape hatch.** It is a local release, not a remote one. If the original enable did reach the control plane, a late `internal_sync_atproto` can report `enabled=true` and switch the account back on after the user turned it off — the mirror image of the convergence that makes the enable rollback safe. Recovering from that needs the gateway to be reachable again, which is the same condition that makes the normal disable work.
- **Out of scope.** Whatever is making `DIVINE_SKY_ATPROTO_CONTROL_PLANE_URL` (`divine-handle-gateway`) fail in production is an operational issue this PR does not touch — it only stops that outage from leaving accounts in an unrecoverable state. A durable retry queue for remote disables is likewise out of scope; the gateway remains the source of truth for the remote link.

## Related Issue

- Closes #none

## Testing

- [ ] `cargo test --workspace --verbose` — **not run locally**: this machine has neither Docker nor a local Postgres, and the ATProto suites need `common::setup_test_db`. Please let CI confirm the five touched tests in `api/tests/atproto_http_test.rs`.
- [ ] `e2e/tests/bluesky-settings.spec.ts` — **not run locally** (needs the Playwright stack); the fixture change is mock-only.
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `cargo test -p keycast_api --lib` — 190 passed
- [x] `cargo test -p keycast_api --test atproto_http_test --no-run` — the new tests compile
- CI note: run 30899938228 first failed in `test_suggest_users_for_admin_uses_the_email_trigram_index` (`keycast_core`, untouched here). Re-running the identical commit passed; the plan assertion depends on ambient row counts and is tracked in #335. Every test in this PR passed in both runs.
- [x] Manual verification: reproduced the stuck state on a production account from the mobile client (see Motivation). The fix itself is not manually verified — it needs a deploy.

## Visuals

- [x] No visual change to the mobile client. In the keycast web card, an account whose enable failed now renders the existing "Failed" state (badge, last-error text, retry button) instead of the plain enable card — no new UI.

### DIFF
--- a/api/src/api/http/atproto.rs
+++ b/api/src/api/http/atproto.rs
@@ -304,8 +304,9 @@ enum OptIn {
 
 /// Writes the pre-trigger opt-in and reports the resulting status.
 ///
-/// Shared by [`enable_user_atproto`] and [`reenable_user_atproto`] so both go
-/// through the same conditional write. See
+/// Shared by [`enable_user_atproto`] and
+/// [`reenable_user_atproto_with_trigger`] so both go through the same
+/// conditional write. See
 /// [`UserRepository::begin_atproto_enable`] for why the write is conditional and
 /// why it leaves `atproto_did` alone.
 async fn begin_atproto_opt_in(
@@ -371,8 +372,7 @@ where
     Ok(response)
 }
 
-/// Clears the opt-in that [`enable_user_atproto`] or [`reenable_user_atproto`]
-/// wrote ahead of the trigger.
+/// Clears the opt-in that [`begin_atproto_opt_in`] wrote ahead of the trigger.
 ///
 /// The enabled flag is set before the control plane is called, so a failed
 /// trigger leaves an opt-in that exists nowhere but this row. Keeping it makes
@@ -428,25 +428,6 @@ async fn roll_back_failed_enable(
     }
 
     Ok(())
-}
-
-pub async fn reenable_user_atproto(
-    repo: &UserRepository,
-    tenant_id: i64,
-    user_pubkey: &str,
-) -> Result<AtprotoStatusResponse, AtprotoControlError> {
-    let claimed_username = require_resolved_username(
-        resolve_username_with_fallback(repo, tenant_id, user_pubkey, |pubkey| async move {
-            divine_names::lookup_by_pubkey(&pubkey).await
-        })
-        .await?,
-    )?;
-
-    Ok(
-        begin_atproto_opt_in(repo, tenant_id, user_pubkey, claimed_username)
-            .await?
-            .0,
-    )
 }
 
 pub async fn reenable_user_atproto_with_trigger<F, Fut>(
--- a/api/src/api/http/atproto.rs
+++ b/api/src/api/http/atproto.rs
@@ -365,10 +365,15 @@ where
 /// control plane still owns, and [`disable_user_atproto_with_trigger`] depends
 /// on it to keep refusing a local-only disable for a provisioned account.
 ///
-/// The rollback only applies while the row still holds the `pending` state this
-/// request wrote. If the trigger did reach the control plane and provisioning
-/// reported back through the internal sync endpoint before the local call
-/// failed, that sync has already moved the row and must not be overwritten.
+/// The rollback only applies while the row is still `pending`. If the trigger
+/// did reach the control plane and provisioning reported back through the
+/// internal sync endpoint before the local call failed, that sync has already
+/// moved the row and must not be overwritten.
+///
+/// That guard is a state check and not a claim of request ownership: the row
+/// records no operation identity, so a rollback cannot tell its own `pending`
+/// write apart from one a later concurrent request left behind, and it can
+/// still compensate a newer opt-in.
 async fn roll_back_failed_enable(
     repo: &UserRepository,
     tenant_id: i64,
--- a/api/src/api/http/atproto.rs
+++ b/api/src/api/http/atproto.rs
@@ -283,8 +283,17 @@ pub async fn enable_user_atproto(
         return Err(AtprotoControlError::UsernameMismatch);
     }
 
-    repo.set_atproto_state(user_pubkey, tenant_id, true, Some("pending"), None, None)
-        .await?;
+    let existing_did = current_atproto_did(repo, tenant_id, user_pubkey).await?;
+
+    repo.set_atproto_state(
+        user_pubkey,
+        tenant_id,
+        true,
+        Some("pending"),
+        existing_did.as_deref(),
+        None,
+    )
+    .await?;
 
     let state = repo
         .get_atproto_state(user_pubkey, tenant_id)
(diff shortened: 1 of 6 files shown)


core/src/repositories/user.rs
```diff
@@ -1470,6 +1498,172 @@ impl UserRepository {
...
+    ///
+    /// `atproto_did` is deliberately left out of the `SET` clause, for the same
+    /// reason as [`Self::roll_back_atproto_enable`]: reading the DID and writing
+    /// it back leaves a window in which provisioning can attach one, and the
+    /// write then erases the only local record that a live repo exists. That
+    /// record is what [`Self::disable_atproto_if_unprovisioned`] reads to decide
+    /// whether a local-only release is safe, so losing it would let a later
+    /// disable report a live account as switched off. Omitting the column
+    /// preserves any DID without a read to go stale.
+    pub async fn begin_atproto_enable(
```
dcadenas, line 1525: **The old opt-in write read the DID and wrote it back, so a sync landing between those two awaits erased it.**

`roll_back_atproto_enable` already documents the right technique — leave `atproto_did` out of the `SET` clause and it survives without a read to go stale. That reasoning applies here too, and `current_atproto_did` did the opposite.

The consequence is the failure the disable guard exists to prevent: with the DID gone, a later disable during the same outage takes the unprovisioned path and reports the account off while a live repo can still publish.

This write now omits the column, and `current_atproto_did` is gone along with the extra round trip it cost.

core/src/repositories/user.rs
```diff
@@ -1470,6 +1498,172 @@ impl UserRepository {
...
+                 SET atproto_enabled = true,
+                     atproto_state = 'pending',
+                     atproto_error = NULL,
+                     atproto_updated_at = $1,
+                     updated_at = $1
+                 WHERE pubkey = $2
+                   AND tenant_id = $3
+                   AND NOT (
+                       atproto_enabled
+                       AND atproto_state IS NOT DISTINCT FROM 'ready'
```
dcadenas, line 1545: **Without this predicate, a redundant enable could report a publishing account as switched off.**

`POST /api/user/atproto/enable` has no `ready` short-circuit of its own, unlike the crosspost route. So a client that re-asserts the opt-in — a retry, or a settings screen that enables on open — moved a `ready` row to `pending`, and when the control plane was down the rollback then wrote `enabled=false, state='failed'` over a live account.

`IS NOT DISTINCT FROM` rather than `=` so that a NULL state does not make the whole predicate NULL and silently skip the write.

`pending` is deliberately still eligible: a stuck opt-in should still be retryable against the control plane.

api/src/api/http/atproto.rs
```diff
@@ -305,29 +340,96 @@ where
...
+    tenant_id: i64,
+    user_pubkey: &str,
+    error: &crate::atproto_provisioning::AtprotoProvisioningError,
+) -> Result<(), AtprotoControlError> {
+    match repo
+        .roll_back_atproto_enable(user_pubkey, tenant_id, Some(error.public_message()))
+        .await?
+    {
+        ConditionalWrite::Applied => {
+            session_repo.revoke_sessions_for_pubkey(user_pubkey).await?;
```
dcadenas, line 416: **A rollback reports the account off, so it owes the same revocation the local-release path does.**

The refresh grant at `api/src/api/http/atproto_oauth.rs:1217` validates only the session's own DPoP and client binding — it never re-reads `ready_atproto_identity`. A session left behind here would keep minting access tokens for up to the 30-day refresh lifetime against an account the status endpoint calls off.

Revoking only on `Applied` keeps it a no-op when the rollback did not land. With the `ready` exclusion in place this is rarely reachable, but it is the invariant your local-release test states, so it should hold on every path that reports off.

dcadenas (approved):
✅ The rollback and the DID-gated local release are the right shape, and the interleaving tests pin them properly.

Three gaps remained around the pre-trigger write. I pushed fixes rather than handing them back; detail is inline.

### Changes I pushed

- The opt-in write no longer reads `atproto_did` and writes it back — it omits the column, so there is no window for a provisioning sync to be overwritten.
- A confirmed `ready` account is now excluded from that write, so a redundant enable during an outage cannot report a publishing account as switched off.
- A rollback that lands now revokes ATProto OAuth sessions, matching the invariant the local-release test states.

The `ready` exclusion is the widest of the three: enabling an already-ready account now returns current status without calling the control plane. `pending` stays eligible, so a stuck opt-in can still be retried.

### Checks

- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- `cargo test --workspace`, plus the `integration-tests` suites for core, api and signer
- `cargo test -p keycast_api --test atproto_http_test`: 35 passed
- `cargo build --release --workspace`
- CI is green on the pushed head.

The two behaviour changes are pinned by tests that fail when the fix is removed — I removed each one and confirmed only the matching test failed. The two DID tests are regression guards, not race reproductions; the window they close has no seam to interleave into, and the fix removes it structurally rather than detecting it.

`enable_failure_keeps_the_did_so_disable_still_refuses` now seeds `pending` with a DID instead of `ready`, because the `ready` exclusion short-circuits before the trigger runs. Its intent and assertions are unchanged.

Merge stays with you.

Issue #338, opened 2026-08-05
feat(oauth): require PKCE on the redirect-based authorization-code flow


dcadenas:
## Summary

For the redirect-based OAuth flow, PKCE is verified only when a code challenge happens to have been stored. Make it required there, and fix the error code returned on failure.

## Problem or Opportunity

`code_challenge` is optional on both `AuthorizeRequest` (`api/src/api/http/oauth.rs:241`) and `ApproveRequest` (`api/src/api/http/oauth.rs:257`), and the exchange only validates when one is present (`api/src/api/http/oauth.rs:2623`). A redirect-flow client that omits it receives a code that can be redeemed without a verifier.

PKCE (RFC 7636) protects the authorization code while it travels back through the user agent or an OS URL handler. That is exactly the redirect flow's exposure, and RFC 9700 (OAuth 2.0 Security Best Current Practice) makes PKCE mandatory for it.

Two smaller conformance problems sit on the same path:

- An omitted `code_challenge_method` defaults to `plain` (`api/src/api/http/oauth.rs:2624`), while the discovery document added in #166 advertises `S256` only.
- `validate_pkce` returns `invalid_request` on a verifier mismatch (`api/src/api/http/oauth.rs:489`, `api/src/api/http/oauth.rs:497`). RFC 7636 §4.6 requires `invalid_grant`.

## Is this already effectively enforced? No, and the distinction matters

The first reaction to this issue is reasonably "the redirect flow already uses PKCE, so this is a no-op." It is worth being precise about why it is not.

Keycast never generates the challenge. It arrives from the client on the authorize URL, and the login page relays whatever was sent (`api/src/api/http/oauth.rs:1477`, `api/src/api/http/oauth.rs:2165`, `api/src/api/http/oauth.rs:2224`, all `code_challenge: codeChallenge || undefined`). Nothing in the authorize path rejects its absence: both request types hold `Option<String>` (`api/src/api/http/oauth.rs:240`, `api/src/api/http/oauth.rs:256`), every call site passes `params.code_challenge.as_deref()` through to storage (`api/src/api/http/oauth.rs:696`, `api/src/api/http/oauth.rs:753`), and `oauth_codes.code_challenge` is nullable (`database/migrations/0001_initial_schema.sql:87`, with a partial index `WHERE code_challenge IS NOT NULL` that anticipates rows without one).

[...]

Issue #282, opened 2026-07-03
feat(oauth): require a registered client for the web-redirect OAuth flow (default-deny unknown client_ids)


dcadenas:
## Problem

For the redirect-based OAuth flow (`GET /oauth/authorize`), any **unregistered** `client_id` currently accepts **any HTTPS `redirect_uri`**. See `core/src/repositories/registered_client.rs` — `validate_redirect_uri` returns `Ok(())` on the `None` (unregistered) arm:

```rust
match allowed {
    None => Ok(()), // Unregistered client — allow any HTTPS (existing behavior)
    Some(patterns) => { /* enforce allowlist */ }
}
```

This is an open-redirect / auth-code-interception risk. An attacker can craft a link using a **trusted** `client_id` (e.g. `divine-web`) but an attacker-controlled `redirect_uri`:

```
https://login.divine.video/oauth/authorize?client_id=divine-web&redirect_uri=https://evil.example/cb
```

The victim logs in on the genuine `login.divine.video` and the authorization code is delivered to the attacker's URL — account takeover, with nothing visibly wrong in the address bar. The consent screen currently labels the app by its raw `client_id` (`api/src/api/http/oauth.rs`), so it reads as legitimate.

Mainstream OAuth providers (Google, GitHub) require pre-registration and reject unknown `client_id`s (`invalid_client`) with no open fallback. We should match that for our redirect flow.

## Scope — read carefully

**In scope:** the redirect-based `/oauth/authorize` flow only. It is the sole production caller of the `registered_clients` gate (`api/src/api/http/oauth.rs:576`).

**Explicitly OUT of scope — must NOT be gated:**

- **NIP-46 nostrconnect / bunker flow.** ~115 distinct `client_id`s in prod are free-form *display labels* on bunker connections (Amethyst, Damus, Primal, nak, and Divine's own mobile bunker labeled "Divine"). They have no `redirect_uri`, never reach the gate, and requiring registration would break third-party remote signing — a core feature.
[...]

PR #339, opened 2026-08-05
fix(oauth): bind the authorization code to the requesting client_id

### LINKED ISSUE #337: fix(oauth): bind the authorization code to the requesting client_id
## Summary

The authorization-code exchange does not check that the `client_id` in the token request matches the `client_id` the code was issued to. Add that comparison.

## Problem or Opportunity

`handle_authorization_code_grant` reads `client_id` off the stored code record at `api/src/api/http/oauth.rs:2603` and uses it for the rest of the exchange. The `client_id` the caller sent is present on `TokenRequest` (`api/src/api/http/oauth.rs:265`) but is never compared against it.

RFC 6749 §4.1.3 requires this check for public clients: the server must "ensure that the authorization code was issued to the authenticated confidential client, or if the client is public, ensure that the code was issued to `client_id` in the request." Keycast's token endpoint performs no client authentication, so every client is public and the check always applies.

The exchange does bind the code to `redirect_uri` (`api/src/api/http/oauth.rs:2616`). That is not a substitute — the spec asks for both.

## Proposed Approach

In `handle_authorization_code_grant`, compare `req.client_id` against the `client_id` on the code record before issuing anything, and reject a mismatch with `invalid_grant`.

Add a test alongside the existing coverage in `api/tests/` that redeems a code with a different `client_id` and expects a rejection.

Blast radius is small: a correctly implemented client already sends the same `client_id` it used at authorization.

## Alternatives Considered

Relying on `redirect_uri` matching alone, which is already enforced. Not equivalent, and not what RFC 6749 asks for.

## Mockups, Logs, or References

- RFC 6749 §4.1.3
- Raised during review of #166 and deliberately kept out of it, since the behavior predates that branch and is untouched by it.

dcadenas:
## Summary

- The token endpoint now checks that the authorization code being redeemed was issued to the `client_id` in the token request, and returns `invalid_grant` when it was not.
- The new rejection path now emits structured diagnostic fields for the OAuth authorization-code flow.
- The public identity integration guide documents the new `invalid_grant` response and matching `client_id` requirement.
- Shared OAuth integration-test state setup now lives in `api/tests/common/mod.rs` instead of being copied across test files.

## Motivation

- RFC 6749 §4.1.3 requires the authorization server to ensure the authorization code was issued to the `client_id` in the request. RFC 6749 §5.2 defines `invalid_grant` for that case. This change adds that check.
- The comparison runs before redirect URI validation, PKCE handling, pending-registration completion, code deletion, and token issuance, so a rejected redemption leaves the code unspent.

## Related Issue

- Closes #337

## Testing

- [x] `cargo test -p keycast_api --test oauth_code_client_binding_test --features integration-tests --no-run`
- [x] `cargo test -p keycast_api --test oauth_verifier_nsec_scope_test --features integration-tests --no-run`
- [ ] Runtime integration tests in this takeover environment: attempted targeted test execution, but local Postgres was unavailable and Docker was not running, so the tests timed out while connecting to the database before assertions ran.

New `api/tests/oauth_code_client_binding_test.rs` covers both directions: redemption presenting a different `client_id` is rejected with `invalid_grant` and leaves the authorization code and its pending registration intact, and redemption by the issuing client still completes. It drives the token handler through the `TokenRequestBody` extractor added in #166, matching the existing `oauth_verifier_nsec_scope_test.rs`.

Daniel's original PR run completed `cargo test --workspace --verbose`, `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`, `cargo fmt --all -- --check`, and the `integration-tests` suites for `keycast_core`, `keycast_api`, `keycast_signer`, and `cluster-hashring` against local Postgres and Redis. CI will re-run after the takeover push.

## Visuals

- [x] No visual change

## Scope

PKCE enforcement (#338) and requiring a registered client (#282) are separate and are not touched here. The `redirect_uri` mismatch response is likewise left as is.

### DIFF
--- a/api/src/api/http/oauth.rs
+++ b/api/src/api/http/oauth.rs
@@ -2854,6 +2854,18 @@ async fn handle_authorization_code_grant(
     let previous_auth_id = auth_code.previous_auth_id;
     let is_headless = auth_code.is_headless;
 
+    // RFC 6749 §4.1.3 requires the token endpoint to ensure the authorization
+    // code was issued to the client_id in the request. RFC 6749 §5.2 specifies
+    // invalid_grant when a code was issued to another client.
+    if client_id != req.client_id {
+        tracing::warn!(
+            "Rejecting authorization code redemption: code was issued to a different client_id"
+        );
+        return Err(OAuthError::InvalidGrant(
+            "Authorization code was not issued to this client".into(),
+        ));
+    }
+
     // Validate redirect_uri matches
     if stored_redirect_uri != *redirect_uri {
         return Err(OAuthError::InvalidRequest(
--- a/api/src/api/http/oauth.rs
+++ b/api/src/api/http/oauth.rs
@@ -2859,6 +2859,14 @@ async fn handle_authorization_code_grant(
     // invalid_grant when a code was issued to another client.
     if client_id != req.client_id {
         tracing::warn!(
+            event = "oauth_authorization_code_rejection",
+            grant_flow = "authorization_code",
+            outcome = "rejected",
+            reason_code = "client_id_mismatch",
+            http_status = StatusCode::BAD_REQUEST.as_u16(),
+            tenant_id,
+            stored_client_id = client_id.as_str(),
+            presented_client_id = req.client_id.as_str(),
             "Rejecting authorization code redemption: code was issued to a different client_id"
         );
         return Err(OAuthError::InvalidGrant(
--- a/api/tests/common/mod.rs
+++ b/api/tests/common/mod.rs
@@ -1,7 +1,26 @@
 // ABOUTME: Shared test utilities and safety guards
 // ABOUTME: Ensures tests never accidentally connect to production database
 
+use chrono::Utc;
+use keycast_api::{
+    api::{
+        http::routes::AuthState,
+        tenant::{Tenant, TenantExtractor},
+    },
+    bcrypt_queue::BcryptQueue,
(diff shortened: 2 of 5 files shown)


NotThatKindOfDrLiz:
@dcadenas I took over the small cleanup from review and pushed it to this PR branch.

Changes added on top of your commit:
- documented the new token endpoint `invalid_grant` response and the matching `client_id` requirement in `docs/IDENTITY_INTEGRATION_GUIDE.md`
- added structured fields to the authorization-code client mismatch warning
- moved the duplicated OAuth integration-test state helpers into `api/tests/common/mod.rs` and updated both affected tests to use them

Verification:
- `cargo test -p keycast_api --test oauth_code_client_binding_test --features integration-tests --no-run`
- `cargo test -p keycast_api --test oauth_verifier_nsec_scope_test --features integration-tests --no-run`
- CI is green: test, Web Tests, semantic PR, CLA

I attempted to run the targeted integration tests locally, but this environment had no reachable Postgres and Docker was not running, so they timed out before assertions. CI ran the actual test job successfully.

Please review the takeover commit and merge when you are ready.

NotThatKindOfDrLiz (approved):
Approved after takeover cleanup. The authorization-code client binding is correctly placed before redirect URI validation, PKCE, registration finalization, and code deletion; docs, observability, and duplicated test setup have been addressed; CI is green.

PR #341, opened 2026-08-06
fix(oauth): reject a refresh grant whose client_id is not the one the token was issued to

### LINKED ISSUE #340: fix(oauth): bind the refresh token to the requesting client at the token endpoint
## What should change

The refresh-token grant at `/api/oauth/token` should establish that the refresh token presented belongs to the client making the request, and fail closed with `invalid_grant` when that cannot be established, before rotating the refresh token and issuing a new access token.

## Why this is separate from #337

#337 covers the authorization-code grant, where RFC 6749 §4.1.3 defines an explicit comparison for public clients: the authorization server ensures the code was issued to the `client_id` present in the request. That made the fix there a direct string comparison against the value stored with the code.

RFC 6749 §6 does not define the equivalent for the refresh grant. It establishes that a refresh token is bound to the client it was issued to, but it does not specify a public-client `client_id` parameter comparison as §4.1.3 does; the binding it describes is expressed through client authentication. RFC 6749 §10.4 addresses the binding obligation for tokens issued to clients that are not authenticated.

So the two grants need different reasoning and different tests, and the correct change here is not the same shape as #337. That is why it is filed on its own rather than folded into that pull request.

## Where it applies

- `api/src/api/http/oauth.rs` — `handle_refresh_token_grant_inner`. The owning `OAuthAuthorization` is already loaded, and its `client_id` is already in scope as `stored_client_id`, currently used for rejection diagnostics. `TokenRequest.client_id` is a required field, so the requesting client's identifier is available on the same path.
- `handle_authorization_code_grant`, in the same file, is the shape #337 landed and is worth reading first for the response semantics — `OAuthError::InvalidGrant` already renders the RFC 6749 §5.2 error body.

## Prior art in this repository

The ATProto token endpoint already performs the equivalent stored-versus-request comparison on both of its paths, including the refresh path, in `api/src/api/http/atproto_oauth.rs`. That is a useful reference for the shape, though its client-authentication model differs and the semantics should not be copied across without checking.

## Suggested scope

- Decide the correct semantics for this grant given that the token endpoint does not authenticate clients, rather than assuming the §4.1.3 comparison transfers.
- Fail closed before rotation and issuance, so a rejected request does not consume or rotate the refresh token.
- Integration test under `api/tests/`, covering both a request that should be rejected and one that should still succeed. Confirm the rejection test fails when the new check is removed, so it cannot pass vacuously.

## Provenance

Found during work on #337 and independently confirmed by a separate reviewing session on a different model that was given only the commit and the requirements. Both reached the same conclusion that it is real and that it does not belong in #337.

dcadenas:
## Summary

- The refresh-token grant at `/api/oauth/token` now compares the request's
  `client_id` with the `client_id` recorded for the authorization that owns the
  refresh token, and answers `invalid_grant` (RFC 6749 §5.2) when they differ or
  when the recorded binding is absent.
- The comparison runs before the refresh token is consumed, through a new
  tenant-scoped `RefreshTokenRepository::find_binding` that reads the binding
  without mutating it.
- Refresh-token consumption is also tenant-scoped, so a token presented on the
  wrong tenant is rejected without spending the owning tenant's token.
- Rejection reason codes `refresh_token_client_mismatch` and
  `refresh_token_binding_lookup_failed` join the existing structured rejection
  log, which continues to log only the recorded `client_id` and never the
  presented one.
- Database failures on this grant now answer with the generic 503 that other
  dependency failures already use, rather than a 400 carrying a repository
  diagnostic. `OAuthError::Database` now holds a rendered string, which let the
  old synthetic `sqlx::Error::Protocol` wrappers go away.

## Motivation

RFC 6749 §6 states that a refresh token is bound to the client it was issued to.
§10.4 requires the authorization server to maintain that binding, to verify it
whenever the client identity can be authenticated, and — where client
authentication is not possible — to deploy other means of detecting abuse,
giving refresh token rotation as its example.

This endpoint does not authenticate clients, so the requirement it can meet is
the one §10.4 states unconditionally: maintain the binding. The step beyond
recording it is to compare it, which is what this change adds. Rotation,
implemented in `RefreshTokenRepository::consume`, remains the §10.4
abuse-detection mechanism and is now constrained to the request tenant.

Scope note for reviewers. Because this endpoint does not authenticate clients,
the presented `client_id` is self-asserted. The comparison is a consistency
check between the request and the recorded binding; it is not a
client-authentication control and should not be reviewed or relied on as one.
Nothing issued on this path derives from the request's `client_id` — the access
token's origin, the policy, the bunker key and the authorization handle all come
from the stored authorization.

This is deliberately not the RFC 6749 §4.1.3 comparison that #339 makes on the
authorization-code grant. §4.1.3 defines that comparison for the code grant;
§6 defines no equivalent parameter for this one, so the two grants need
different reasoning and different tests.

## Ordering, and its consequences

The comparison runs before `consume` so that a rejected mismatched-client request
does not spend the caller's refresh token. Tenant-scoping `consume` separately
ensures a cross-tenant request cannot spend a token before being rejected.

Two consequences are deliberate and worth reviewing as such:

1. A refresh token presented with a non-matching `client_id` is answered with
   the mismatch result even when that token is expired or already consumed, so
   its lifecycle state is not disclosed to a caller that does not own it.
2. With the database unreachable, the first failing call on this path is now the
   binding lookup rather than `consume`, reported as
   `refresh_token_binding_lookup_failed`.

## Compatibility

Each first-party client sends one constant `client_id` for authorize, code
exchange and refresh alike — `keycast-login/src/oauth.ts` uses its configured
client id on all three, and the mobile app's Flutter OAuth client does the same.
`web/`, `e2e/` and `examples/` do not exercise the refresh grant. Refresh tokens
are only ever attached to a newly created authorization, whose `client_id` is
written from the authorization code on the single insert path, so the recorded
binding is present for any live refresh token.

## Related Issue

- Closes #340

## Testing

- [x] `cargo test -p keycast_core refresh_token`
- [x] `cargo test -p keycast_api --test oauth_refresh_client_binding_test`
- [ ] `cargo test -p keycast_api --features integration-tests --test oauth_refresh_client_binding_test` (local Docker/Postgres unavailable in takeover environment; CI covers this)
- [ ] `cargo test -p keycast_core --features integration-tests refresh_token` (local Docker/Postgres unavailable in takeover environment; CI covers this)
- [ ] `cargo test --workspace --verbose` (local run reached DB-backed suites and failed on `PoolTimedOut`; CI covers this)
- [x] The original rejection tests were confirmed non-vacuous: with the new comparison
      removed, `refresh_grant_rejects_a_client_id_other_than_the_one_the_token_was_issued_to`
      fails by receiving HTTP 200, and
      `refresh_grant_rejects_a_mismatched_client_before_disclosing_token_state`
      fails by receiving `"Invalid or expired refresh token"` in place of the
      mismatch description. The success test still passes with the comparison
      removed, so it is not what makes the rejection tests meaningful.

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/oauth.rs
+++ b/api/src/api/http/oauth.rs
@@ -2619,7 +2619,7 @@ async fn handle_refresh_token_grant_inner(
 
     // Consume refresh token atomically (validates + marks as consumed)
     // This implements one-time use per RFC 9700 token rotation
-    let token_record = match refresh_token_repo.consume(refresh_token).await {
+    let token_record = match refresh_token_repo.consume(refresh_token, tenant_id).await {
         Ok(Some(token_record)) => token_record,
         Err(error) => {
             // Same reasoning as the binding lookup above: the token's state is
@@ -2722,7 +2722,7 @@ async fn handle_refresh_token_grant_inner(
         .await
         .map_err(|error| {
             refresh_grant_rejection(
-                OAuthError::from(error),
+                OAuthError::Database(error.to_string()),
                 RefreshTokenRejectionReason::StoredKeyLookupFailed,
                 stored_client_id.as_deref(),
             )
@@ -2798,7 +2798,7 @@ async fn handle_refresh_token_grant_inner(
         .await
         .map_err(|error| {
             refresh_grant_rejection(
-                OAuthError::from(error),
+                OAuthError::Database(error.to_string()),
                 RefreshTokenRejectionReason::RotatedTokenInsertFailed,
                 stored_client_id.as_deref(),
             )
--- a/api/src/api/http/oauth.rs
+++ b/api/src/api/http/oauth.rs
@@ -391,7 +391,10 @@ pub enum OAuthError {
     InvalidEmail,
     InvalidRequest(String),
     InvalidGrant(String), // RFC 6749 - for invalid/expired refresh tokens or auth codes
-    Database(sqlx::Error),
+    /// A database operation failed, so the request could not be judged. Holds
+    /// the rendered diagnostic: callers reach this from `sqlx::Error` and from
+    /// `RepositoryError`, which has already rendered its own.
+    Database(String),
     Encryption(String),
     ServerError(String),
 }
@@ -471,7 +474,7 @@ impl IntoResponse for OAuthError {
 
 impl From<sqlx::Error> for OAuthError {
     fn from(e: sqlx::Error) -> Self {
-        OAuthError::Database(e)
+        OAuthError::Database(e.to_string())
     }
 }
 
@@ -545,7 +548,7 @@ async fn resolve_policy_from_scope(pool: &sqlx::PgPool, scope: &str) -> Result<i
                 "Unknown policy '{}'. See GET /api/policies for available options.",
                 policy_slug
             )),
-            _ => OAuthError::Database(sqlx::Error::Protocol(e.to_string())),
+            _ => OAuthError::Database(e.to_string()),
         })?;
 
     Ok(requested_policy.id)
@@ -2449,6 +2452,8 @@ async fn handle_refresh_token_grant(
 #[derive(Clone, Copy, Debug)]
 enum RefreshTokenRejectionReason {
     MissingRefreshToken,
+    BindingLookupFailed,
+    ClientMismatch,
     ConsumeFailed,
     DiagnosticLookupFailed,
     TokenNotFound,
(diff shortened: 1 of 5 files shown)


NotThatKindOfDrLiz:
@dcadenas I pushed a takeover commit on top of this PR: `aad1be9` (`fix(oauth): scope refresh token consumption by tenant`).

What changed:
- Made `RefreshTokenRepository::consume` tenant-scoped so a cross-tenant refresh request cannot spend another tenant's token before rejection.
- Converted the remaining refresh-grant repository failures on stored-key lookup and rotated-token insert to `OAuthError::Database(...)`, preserving the generic 503 response instead of leaking repository diagnostics as 400s.
- Added cross-tenant regression coverage and updated the stale silent-reauth doc snippet.
- Updated the PR body to remove stale wording and document the takeover verification.

Verification:
- `cargo test -p keycast_core refresh_token`
- `cargo test -p keycast_api --test oauth_refresh_client_binding_test`
- GitHub checks are green on the current head: test, Web Tests, Semantic PR, and CLA.

Local DB-backed integration commands could not run in my takeover environment because Docker/Postgres is unavailable and connection attempts timed out with `PoolTimedOut`; CI covered the DB-backed path successfully.

Please review the takeover changes and merge when ready.

NotThatKindOfDrLiz (approved):
Final takeover review is clean. The tenant-scoped consume fix closes the cross-tenant token burn hole, the remaining refresh-grant repository failures now use the same generic 503 path, and CI is green on the current head.

PR #342, opened 2026-08-06
fix(oauth): return invalid_grant on PKCE verifier mismatch

dcadenas:
## Summary

A PKCE verifier mismatch was reported with the wrong OAuth error.
RFC 7636 requires `invalid_grant` there, and the token endpoint returned `invalid_request` — not even in the registered form, since the `error` member carried an English sentence rather than an error code, so a client matching on it could not classify the failure.
The mismatch is now returned as `invalid_grant` in the RFC 6749 error shape, for both the `S256` and the `plain` comparison.

A second, separable commit hardens the failure log on the same statement: it truncated a client-supplied challenge by byte range, which could panic for values the server stores verbatim.
The log now truncates on character boundaries instead.
Nothing outside that log line changes.

The unsupported-transform branch is deliberately left alone. RFC 7636's `invalid_grant` requirement covers the verifier comparison only, and its `invalid_request` requirement for an unsupported transform is addressed to the authorization endpoint rather than to token exchange.

## Motivation

This is the smallest of the three problems in #338 and the only one independent of the others, so it ships on its own.
It carries no rollout risk, because it changes how an already-failing exchange is reported and not whether any exchange succeeds.

Requiring `code_challenge` on the redirect flow and rejecting `plain` are deliberately not here.
Those can reject clients that work today, and unlike a registered-client gap they have no runtime toggle, so they need client-coverage evidence before they land.

This changes OAuth token-endpoint behavior, which is sensitive operational surface.
The change is confined to the PKCE failure path; the success path and every other grant are untouched.

## Related Issue

- Part of #338. This lands step 3 only — steps 1 and 2 remain open, so this should not close the issue.

## Testing

The new tests cover the protocol result, the serialized response body, and the same failure through the real token handler against a database, so the assertion is on what a client actually receives rather than on an internal value.
The `S256` success case uses the worked example published in RFC 7636 rather than a locally recomputed hash, so a mistake in this module's own hashing cannot make the failure cases pass by accident.
Each test was confirmed to fail when the corresponding source change is reverted.

- [x] `cargo nextest run --workspace` — 500 passed, 15 skipped
- [x] `cargo test -p keycast_api --features integration-tests --lib --tests` — green

Not tested: `tests/qa/` is excluded from the workspace and needs a running server, so it did not run here. Its existing PKCE assertion is satisfied by the new response and needed no change.

## Visuals

- [x] No visual change

### DIFF
--- a/api/src/api/http/oauth.rs
+++ b/api/src/api/http/oauth.rs
@@ -507,7 +507,9 @@ fn validate_pkce(
                     &computed_challenge[..16],
                     &code_challenge[..16]
                 );
-                return Err(OAuthError::InvalidRequest(
+                // RFC 7636 section 4.6: when the values are not equal the token endpoint
+                // MUST return `invalid_grant`.
+                return Err(OAuthError::InvalidGrant(
                     "Invalid code_verifier: PKCE validation failed".to_string(),
                 ));
             }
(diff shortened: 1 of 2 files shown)


NotThatKindOfDrLiz (approved):
Approved. The PKCE mismatch response now uses the registered `invalid_grant` shape for both `S256` and `plain`, and the log-prefix change fixes the same failure branch without changing successful exchanges.

I also verified the sibling ATProto endpoint has the same response-shape issue, but that is pre-existing and belongs in a separate follow-up rather than expanding this ready PR. Filed that separately.

Additional local verification:

- `cargo test -p keycast_api validate_pkce --all-features`
- `cargo test -p keycast_api pkce_mismatch_serialises --all-features`

PR #345, opened 2026-08-06
fix(atproto-oauth): return invalid_grant for PKCE mismatch

### LINKED ISSUE #344: fix(atproto-oauth): return invalid_grant on PKCE verifier mismatch
## Summary

The ATProto OAuth token endpoint should return the registered OAuth error code `invalid_grant` when an authorization-code exchange fails the PKCE verifier comparison.

Today `api/src/api/http/atproto_oauth.rs` returns `AuthError::BadRequest("Invalid PKCE verifier")`, which serializes the prose message into the `error` member instead of returning the RFC 6749 error shape.

## Expected

- Status: `400 Bad Request`
- Body includes:
  - `"error": "invalid_grant"`
  - `"error_description": "..."`

## Scope

- Update the ATProto token endpoint PKCE mismatch branch.
- Add a focused ATProto token endpoint test that asserts the wire response body for a mismatched `code_verifier`.
- Keep unrelated ATProto OAuth validation behavior unchanged.

## Context

Follow-up from #342. That PR fixes the same response-code class for the standard OAuth token endpoint and intentionally keeps this sibling ATProto endpoint separate.

NotThatKindOfDrLiz:
## Summary

- Add an OAuth protocol error shape to AuthError for RFC 6749-style token endpoint responses.
- Return invalid_grant with error_description when the ATProto token endpoint rejects a mismatched PKCE code_verifier.
- Add a focused ATProto OAuth token endpoint regression test for the wire response.

## Motivation

- Clients need the registered OAuth error code to classify PKCE verifier mismatch failures instead of parsing prose.
- The change is limited to an already-failing ATProto token exchange path; successful token exchange behavior is unchanged.

## Related Issue

- Closes #344

## Testing

- [ ] cargo test --workspace --verbose
- [x] cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated
- [x] cargo fmt --all -- --check
- [x] cargo test --workspace --no-run
- [ ] cd api && cargo test --test atproto_oauth_http_test (blocked locally: Docker daemon unavailable, so tests timed out connecting to keycast_test Postgres)

## Visuals

- [x] No visual change

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -60,6 +60,9 @@ components:
         code:
           type: string
           description: Stable application error code, when available
+        error_description:
+          type: string
+          description: Human-readable detail accompanying an RFC 6749 OAuth error code, when available
       required:
         - error
 
@@ -1046,7 +1049,7 @@ paths:
                   sub:
                     type: string
         '400':
-          description: Invalid grant or DPoP proof
+          description: Invalid grant or DPoP proof. A PKCE `code_verifier` that does not match the stored `code_challenge` returns the RFC 6749 error code `invalid_grant` with an `error_description`.
           content:
             application/json:
               schema:
--- a/api/src/api/http/atproto_oauth.rs
+++ b/api/src/api/http/atproto_oauth.rs
@@ -1,6 +1,6 @@
 use axum::{
     extract::{Query, State},
-    http::{HeaderMap, HeaderValue},
+    http::{HeaderMap, HeaderValue, StatusCode},
     response::{IntoResponse, Redirect},
     Form, Json,
 };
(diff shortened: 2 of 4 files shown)


api/openapi.yaml
```diff
@@ -60,6 +60,9 @@ components:
         code:
           type: string
           description: Stable application error code, when available
+        error_description:
```
dcadenas, line 63: I added this member, and the endpoint's `400` description below, in a separate commit.

The token endpoint now returns `error_description`, but the `400` response points at the shared `Error` schema, which declared only `error` and `code`. The spec is embedded with `include_str!` and served at `/docs/openapi.json`, so it is shipped contract rather than a side document.

Adding it to the shared schema follows what cafdd2 did for the `code` member, in the same PR that introduced it. `error` stays the only required member, so existing consumers are unaffected.

A dedicated OAuth error schema would be more precise, but that is a separate refactor — the shared schema already mixes a Keycast-specific `code` with the generic `error`.

dcadenas (approved):
✅ The `invalid_grant` mapping is correct and the new test genuinely pins it — reverting the handler branch makes the test fail, so it cannot pass for an unrelated reason.

The PKCE comparison runs before DPoP validation, so the test's DPoP-less token POST reaches the branch it claims to cover rather than short-circuiting on an earlier guard.

### Changes I pushed

- Documented the new `error_description` member in the OpenAPI spec, which was still describing this endpoint's 400 with a schema that has no such member. Detail inline.

That is the only change I made; the implementation and test are the author's as written.

### Checks

- `cargo test --workspace`: no failures
- `cargo test --test atproto_oauth_http_test`: 9 passed — this covers the run the description marks as blocked on a local Docker daemon

### Follow-ups

Neither blocks this branch, and both are outside the scope #344 sets.

- The endpoint's other 400s still return prose in the `error` member rather than a registered OAuth error code.
- No token-endpoint response sets `Cache-Control: no-store`. That predates this change and applies to the success path too.

Issue #291, opened 2026-07-11
fix(nostr-rpc): track and reduce HTTP signer RPC tail latency


realmeylisdev:
## Summary

Parent tracker for the HTTP `nostr_rpc` latency work surfaced by Divine mobile DM delivery failures.

OAuth remote-signer users saw individual HTTP `sign_event` / `nip44_encrypt` RPCs take 20-30s. The evidence points at request-path database pool contention rather than cryptography latency: warm mutating RPCs still perform account-status database checks, activity logging uses the same pool, and the previous SQLx acquire timeout allowed long waits under saturation.

Mobile has already shipped a client-side mitigation in divinevideo/divine-mobile#6046 so stalled signer calls fail visibly and can be retried. This issue tracks the Keycast server-side cure.

## Current status

- PR #354 is in flight for the first slice: shorten the shared SQLx acquire timeout, keep acquire below the HTTP RPC budget, and add saturated-pool regression coverage for retryable 503 behavior.
- #355 tracks the highest-leverage follow-up: cache account-status checks off the warm mutating RPC hot path.
- #356 tracks request timeout coverage plus pool/acquire/operation latency metrics so the fix is measured in production.
- #357 tracks optional activity-log coalescing if metrics still show that `spawn_log_activity` amplifies pool pressure.

## Child issues

- [ ] #354 / PR: `fix(core): shorten database acquire timeout`
- [ ] #355: `fix(nostr-rpc): cache account status checks off hot path`
- [ ] #356: `fix(nostr-rpc): add request timeout and latency metrics`
- [ ] #357: `perf(nostr-rpc): coalesce RPC activity logging under pool pressure`

## Done when

- HTTP `nostr_rpc` tail latency is bounded by explicit request and acquire budgets.
- Warm mutating RPCs no longer require a database checkout solely for account-status validation.
- Metrics can distinguish DB acquire wait, RPC operation latency, and total route latency.
- Production verification confirms the 20-30s latency cliff is gone or identifies the remaining bottleneck in a follow-up.

## Original evidence

- Mobile logs showed `nip44_encrypt` calls completing in roughly 23s and 30s over HTTP 200.
- A kind-14 DM send hit the mobile 30s timeout and did not deliver.
- The HTTP 200 shape ruled out the NIP-46 bunker/relay path for this symptom.
- Code inspection found a per-call `SELECT status FROM users` on warm mutating RPCs plus background activity updates on the same Postgres pool.

## Related

[...]

Issue #257, opened 2026-06-22
Add a global bound on concurrent blocking crypto (spawn_blocking pool contention)


mbradley:
Surfaced during review of #256 (nip17_unwrap_batch).

`nip17_unwrap_batch` fans out up to `UNWRAP_BATCH_CONCURRENCY` (16) `spawn_blocking` tasks per request, but that cap is per-request only. With Cloud Run concurrency=50 per instance, the worst case is roughly 50 × 16 = 800 concurrent blocking tasks against tokio's default 512-thread blocking pool, which is shared with every other `spawn_blocking` in the process (`sign_event`, nip04/nip44 encrypt/decrypt). `max_blocking_threads` is not overridden.

Because each task is short and pure-CPU (two Schnorr verifies plus two NIP-44 decrypts), the pool queues rather than deadlocks, so this is not a correctness problem. But under a burst of large batches it could add head-of-line latency for unrelated signing on the same pool.

Options to consider:
- A process-wide `Semaphore` gating blocking crypto.
- Deliberately sizing `max_blocking_threads` and documenting the choice.
- Documenting why 16 was chosen as the per-request fan-out.

Not blocking #256. Filing to track.

dcadenas:
Password and bcrypt admission is now tracked by #366 (unified bounded bcrypt CPU admission). This issue remains focused on non-password blocking cryptography and should compose with #366 and #370 (expensive HTTP route admission) without bypassing either bound or double-counting shared capacity. Parent coordination: #372 (layered overload readiness).

PR #351, opened 2026-08-08
fix(api): bound HTTP RPC latency

NotThatKindOfDrLiz:
## Summary

- Adds HTTP RPC latency, status-check, DB acquire, pool state, and activity-drop metrics with buckets that cover 30s/60s tails.
- Bounds `/api/nostr` with an 8s handler timeout (504) behind a wider tower backstop, and bounds the GCP KMS retry loop inside that request budget.
- Replaces per-request detached OAuth activity updates with a bounded batching writer drained after HTTP shutdown.
- Maps pool exhaustion to a retryable 503 on the warm status check, the cold handler-load path, and shared API repository-error paths without echoing connect-level detail into pre-auth responses.
- Keeps account status and minor-protection flags live per request; this intentionally does not add a short TTL status cache.
- `KMS_ATTEMPT_TIMEOUT_SECS = 2` is process-wide for GCP KMS encrypt/decrypt calls. It is sized for the cold HTTP RPC signing path, but also affects key creation, signer cold loads, and admin paths that use the GCP key manager.

## Motivation

- Prevent HTTP Nostr RPC calls from sitting behind shared resource contention long enough to break DM delivery.
- Make pool acquire waits, live status checks, and long-tail RPC latency visible before future capacity decisions.
- Remove out-of-band per-request DB acquisitions from the hot path.

## Related Issue

- Refs #291

Remaining work for #291: the per-request `SELECT status, verified_minor` pool checkout is still on every mutating RPC, because a suspension or minor-protection flip has to take effect on the next request. #291 ranked caching that first. This bounds and instruments the symptom rather than removing that demand, so it does not close the issue on its own.

## Testing

- [x] `cargo test -p keycast_api activity_log::tests`
- [x] `cargo test -p keycast_api repository_unavailable_maps_to_503`
- [x] `cargo test -p keycast_api cold_path_pool_exhaustion_is_a_retryable_503_without_leaking_detail`
- [x] `cargo test -p keycast_core kms_retry_loop_fits_inside_the_http_rpc_request_bound`
- [x] `cargo test --workspace --verbose` attempted; local run failed before exercising this diff because `api/tests/common/mod.rs` could not connect to the test database (`PoolTimedOut`) in `atproto_http_test`.
- [x] `bun run test` attempted; local run could not start Docker dependencies because Docker was unavailable at `unix:///Users/lizsw/.docker/run/docker.sock`.
- [x] Earlier reviewer-run checks on this PR: `bun run test:ci` with Docker Postgres + Redis (`520 tests run: 520 passed, 16 skipped` plus integration-feature suites), `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`, and `cargo fmt --all -- --check`.

## Visuals

- [x] No visual change

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -41,6 +41,7 @@ subtle = "2.6"
 sqlx = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
+tower = { version = "0.5", features = ["timeout", "util"] }
 tower-http = { version = "0.6", features = ["trace", "util", "cors", "fs"] }
 tracing = { workspace = true }
 tracing-appender = { workspace = true }
@@ -63,7 +64,6 @@ redis = { version = "0.27", features = ["tokio-comp", "connection-manager", "aio
 cluster-hashring = { path = "../cluster-hashring" }
 
 [dev-dependencies]
-tower = { version = "0.5", features = ["util"] }
 hyper = { version = "1.5", features = ["full"] }
 http-body-util = "0.1"
 aes-gcm = "0.10"
--- a/api/src/activity_log.rs
+++ b/api/src/activity_log.rs
@@ -1,6 +1,7 @@
 // ABOUTME: Bounded OAuth activity logger for HTTP RPC requests
 // ABOUTME: Coalesces authorization activity updates so request handlers never spawn DB work
 
+use keycast_core::metrics::METRICS;
 use sqlx::PgPool;
 use std::{
     collections::BTreeMap,
@@ -19,12 +20,21 @@ const DEFAULT_QUEUE_CAPACITY: usize = 4096;
 const DEFAULT_FLUSH_INTERVAL: Duration = Duration::from_secs(1);
 const DEFAULT_MAX_BATCH_IDS: usize = 256;
 
+/// Ceiling on retained-but-unwritten authorization ids after failed flushes.
+/// A failed flush is retried on the next tick rather than discarded, but a
+/// database that stays unhappy must not grow this map without bound.
+const MAX_RETAINED_IDS: usize = 4096;
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ActivityLogResult {
     Queued,
     Dropped,
     SkippedNonOauth,
     Disabled,
+    /// The writer has already stopped. Distinct from `Disabled` because it is a
+    /// lost update: the API keeps serving in-flight requests through the HTTP
+    /// drain after the writer exits, and those records have nowhere to go.
+    WriterStopped,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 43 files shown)


dcadenas (approved):
✅ The bounds land where they need to. I took over and pushed fixes rather than handing the list back.

The instrumentation-before-tuning call is the right one, and the histogram buckets reach far enough out to see the 30s tail that started this.

### Changes I pushed

- The activity writer can no longer miss graceful shutdown, and stops losing updates silently when a flush fails, when it gives up, or when records arrive after it has stopped.
- A failing flush no longer retries once per request — the interval paces retries instead of the batch-size trigger.
- Timed-out requests now record a latency sample with their method label and return the endpoint's normal error body.
- Pool exhaustion on the cold handler-load path returns a retryable 503 like the warm status check already did, without echoing connect-level detail into the response.
- The KMS retry loop now fits inside the request budget that waits on it, asserted against a shared constant so a latency retune has to face it.

Bounding a request does not stop the CPU work it was waiting on. An abandoned unwrap keeps its crypto permit and blocking-pool slot to completion, which is deliberate in the existing code and is what makes that cap a true bound — so a 504 plus a client retry can offer more load than it sheds. That is the crypto-pool bound in #257, not something to change here.

### Checks

- `cargo fmt --all -- --check` and `cargo clippy --workspace --all-targets --all-features -- -D warnings -A deprecated`
- `bun run test:ci` with Postgres and Redis: `520 tests run: 520 passed, 16 skipped`, plus the `integration-tests`-feature suites
- Each new regression test verified against the unfixed code first

### Follow-ups

- #291 stays open. Every mutating RPC still takes a pool connection for the live status check, which the issue ranked first for removal. Keeping it live is the right call for a suspension and minor-protection gate, so the demand needs a different answer than a TTL cache.
- The acquire timeout is now 5s process-wide. The signer shares that pool and has no HTTP retry above it; worth confirming its DB-error path degrades acceptably under the saturation this targets.

NotThatKindOfDrLiz:
Pushed follow-up fixes in `f000f63`:

- Kept the shared SQLx acquire timeout at 60s so this PR relies on the `/api/nostr` handler bound instead of changing every pool consumer.
- Moved the OAuth activity writer shutdown after HTTP drain by giving it its own `Notify`.
- Added `ApiError::ServiceUnavailable` and mapped `RepositoryError::Unavailable` to 503 for shared API paths.
- Removed the duplicate `ActivityLogger::dropped_total` test-only counter surface.
- Updated the `sign_canonical` success path, added after this branch was rebased on current `origin/main`, to use the bounded activity logger.
- Updated the PR body to call out the process-wide GCP KMS attempt timeout.

Verification:

- `cargo test -p keycast_api activity_log::tests`
- `cargo test -p keycast_api repository_unavailable_maps_to_503`
- `cargo test -p keycast_api cold_path_pool_exhaustion_is_a_retryable_503_without_leaking_detail`
- `cargo test -p keycast_core kms_retry_loop_fits_inside_the_http_rpc_request_bound`

Broader local checks:

- `cargo test --workspace --verbose` was attempted, but the local run failed before exercising this diff because `api/tests/common/mod.rs` could not connect to the test database (`PoolTimedOut`) in `atproto_http_test`.
- `bun run test` was attempted, but Docker dependencies could not start because Docker was unavailable at `unix:///Users/lizsw/.docker/run/docker.sock`.

PR #352, opened 2026-08-09
feat(rpc): add creator-binding canonical signing

### LINKED ISSUE #186: feat(rpc): expose sign_canonical for C2PA creator-binding
## Summary

Divine mobile's C2PA creator-binding flow needs a `sign_canonical` JSON-RPC
method on the Keycast server so OAuth-signed-in users get a real inline
creator-binding assertion on every Divine post. Today the mobile-side
fallback at `KeycastRpc.signCanonicalPayload` returns `null` (no method on
the backend), so `NostrCreatorBindingService.createAssertion` skips the
binding and the C2PA proof manifest ships without it.

The mobile side is already wired and forward-compatible (mobile PR
[divinevideo/divine-mobile#3683](https://github.com/divinevideo/divine-mobile/pull/3683)):
once this method ships, every Divine OAuth post gets the binding for free
with no further mobile change.

## Method contract

| Field | Value |
|---|---|
| Method name | `sign_canonical` |
| Params | `[base64(payload)]` — single base64-encoded byte string |
| Result | hex-encoded schnorr signature (`String`) |
| Error | standard JSON-RPC error response on unsupported / failure |

## Determinism contract (REQUIRED)

The signature **must be deterministic** for a given `(privateKey, payload)`
pair so repeated signing of the same payload produces the same signature —
this is what keeps creator-binding assertions stable across re-publishes,
re-uploads, and proof-manifest re-evaluations.

The mobile-side local signer at
[`mobile/lib/services/local_key_signer.dart:55-74`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/services/local_key_signer.dart#L55-L74)
implements this as:

1. **Hash**: `digest = SHA-256(payload)` → 32-byte digest, hex-encoded as a
   lowercase string.
2. **Sign**: BIP-340 schnorr signature over the digest using the account's
   private key.
3. **Auxiliary data**: **32 zero bytes** (constant
   `_canonicalPayloadAux = '0' * 64` in hex). BIP-340 randomized aux would
   produce different signatures across calls; that is unacceptable for
   creator-binding stability.

Pseudocode (Rust-ish — adjust to Keycast's stack):

```rust
fn sign_canonical(account_private_key: SecretKey, payload_b64: String) -> Result<String> {
    let payload = base64::decode(payload_b64)?;
    let digest: [u8; 32] = sha256(&payload);
    let aux: [u8; 32] = [0; 32]; // MUST be all zeros for determinism
    let sig = schnorr::sign(account_private_key, digest, aux);
    Ok(hex::encode(sig))
}
```

## Test vector

Computed against the Dart `bip340` package (which the mobile side uses).
A backend implementation should self-verify by signing the same payload
with the same private key and asserting bit-equal output.

| Field | Value |
|---|---|
| `payload` (UTF-8) | `divine-creator-binding-test` |
| `payload` (hex) | `646976696e652d63726561746f722d62696e64696e672d74657374` |
| `payload` (base64) | `ZGl2aW5lLWNyZWF0b3ItYmluZGluZy10ZXN0` |
| `privateKey` | `5ee1c8000ab28edd64d74a7d951ac2dd559814887b1b9e1ac7c5f89e96125c12` |
| `digest` (hex) | `39341a12a2f77007d6e72841f667523d39463a825d82c6c98981881283fb7ed0` |
| `aux` (hex) | `0000000000000000000000000000000000000000000000000000000000000000` |
| **expected `signature`** | `9baed2647e5f9d059b68eb03c6e3e6dcdf53cbe94fb143af70fb6e7332ee9997cc7ba5ac9cdb9049a0e47c8c20e2031843e88c59dcba3c3ff8fc34eeae4a565f` |

If your BIP-340 implementation produces a different signature for these
inputs, the most likely cause is non-zero aux bytes (the default for many
schnorr libraries) — make sure the aux is forced to 32 zero bytes.

## Mobile-side wiring (already shipped on the PR branch)

Permalinks pinned to commit
[`7c6c0f78a`](https://github.com/divinevideo/divine-mobile/commit/7c6c0f78acb3d45b23cc2bf43a5cb9b014101487)
(head of `fix/publish-hang-disconnected-relay`):

- [`mobile/packages/keycast_flutter/lib/src/rpc/keycast_rpc.dart`](https://github.com/divinevideo/divine-mobile/blob/7c6c0f78acb3d45b23cc2bf43a5cb9b014101487/mobile/packages/keycast_flutter/lib/src/rpc/keycast_rpc.dart#L125-L156)
  — `signCanonicalPayload(Uint8List payload)` calls
  `_call('sign_canonical', [base64Encode(payload)], (r) => r as String)`.
  Returns `null` on `RpcException` (today's path). No further mobile work
  needed once this method ships.
- [`mobile/lib/services/nostr_identity.dart`](https://github.com/divinevideo/divine-mobile/blob/7c6c0f78acb3d45b23cc2bf43a5cb9b014101487/mobile/lib/services/nostr_identity.dart)
  — `KeycastNostrIdentity.signCanonicalPayload` falls back to `KeycastRpc`
  when no local signer is available.
- [`mobile/lib/services/nostr_creator_binding_service.dart`](https://github.com/divinevideo/divine-mobile/blob/7c6c0f78acb3d45b23cc2bf43a5cb9b014101487/mobile/lib/services/nostr_creator_binding_service.dart)
  — `createAssertion` returns `null` (skip binding) instead of throwing
  when canonical signing is unsupported.

## Acceptance

- [ ] `sign_canonical` accepts the param shape above and returns a hex
      schnorr signature.
- [ ] Signing the same `(privateKey, payload)` twice produces bit-equal
      output.
- [ ] Cross-implementation signature for the test vector above matches
      `9baed2647e5f9d059b68eb03c6e3e6dcdf53cbe94fb143af70fb6e7332ee9997cc7ba5ac9cdb9049a0e47c8c20e2031843e88c59dcba3c3ff8fc34eeae4a565f`.
- [ ] Failure modes (unknown method, bad params, internal error) surface
      as standard JSON-RPC error responses; mobile already treats those
      as "skip binding gracefully."

## Source

- Tracked from divinevideo/divine-mobile#3688 (sub-item 2)
- Mobile fix: divinevideo/divine-mobile#3683
- Determinism reference:
  [`mobile/lib/services/local_key_signer.dart:55-74`](https://github.com/divinevideo/divine-mobile/blob/main/mobile/lib/services/local_key_signer.dart#L55-L74)

NotThatKindOfDrLiz:
## Summary

- Add `sign_canonical` to `POST /api/nostr` for base64-encoded C2PA creator-binding payloads.
- Sign `SHA-256(payload)` with deterministic BIP-340 Schnorr using all-zero aux data and return lowercase hex.
- Gate canonical signing through creator-binding payload validation plus cached custom permissions.
- Map creator-binding permission to profile/identity signing: full access and kind `0` policies allow it; read-only and non-profile kind policies deny it.
- Update docs/admin method lists and add focused unit/integration coverage.

## Motivation

- OAuth-signed-in creators need Keycast to produce the same deterministic creator-binding signatures as local-key mobile users.
- The new method must not become a generic Nostr-key signing oracle, so the implementation validates the expected assertion domain and preserves policy restrictions before signing.

## Contract Notes

- `sign_canonical` is intentionally not a generic canonical signer. The HTTP method accepts only validated creator-binding JSON for the signing pubkey; arbitrary payload bytes, including the original issue test-vector payload, are rejected at the RPC layer.
- Issue #186's cross-implementation vector is still covered one layer down by `SigningSession::sign_canonical_matches_cross_implementation_vector`, which verifies the deterministic BIP-340 primitive directly.
- Creator-binding signing is treated as a profile/identity signing capability for policy purposes: full access and kind `0` policies allow it; read-only and non-profile kind policies deny it.
- Liz ratifies the branch changes pushed by dcadenas on this author-owned PR, including the security-sensitive validator and permission-boundary tightening.

## Related Issue

- Closes #186
- Follow-up: #353 tracks the pre-existing flaky PKCE logging test.

## Testing

- [x] `cargo test -p keycast_core --lib`
- [x] `cargo test -p keycast_api --lib`
- [x] `cargo test -p keycast_api --features integration-tests --test nostr_rpc_integration_test`
- [x] `cargo nextest run --workspace`
- [x] `cd web && bun run check`

Notes:

- CI passed on head `db7e6b1`, including the `test` job and all five `sign_canonical` integration cases.
- Cleanup verification on `db7e6b1`: `cargo fmt --all -- --check` and `cargo test -p keycast_core --lib`.
- dcadenas separately reported `cargo nextest run --workspace` with 549 passed / 16 skipped, and `cargo test -p keycast_api --features integration-tests --test nostr_rpc_integration_test` with 23 passed.
- `cargo test --workspace --verbose` remains unchecked because the PR was validated with the narrower CI-equivalent integration target plus nextest.
- A pre-existing flaky PKCE logging unit test is tracked separately in #353 and is unrelated to this creator-binding change.

## Visuals

- [x] No visual change

No screenshot attached; the web change is a one-line static addition to the admin supported-methods table and has no runtime UI path.

### DIFF
--- a/api/src/technical_docs.rs
+++ b/api/src/technical_docs.rs
@@ -213,7 +213,7 @@ pub async fn technical_docs() -> Html<&'static str> {
         <li><code>connect</code> - Initial connection handshake</li>
         <li><code>get_public_key</code> - Returns user's public key</li>
         <li><code>sign_event</code> - Signs a Nostr event</li>
-        <li><code>sign_canonical</code> - Signs a C2PA creator-binding payload</li>
+        <li><code>sign_canonical</code> - Signs a C2PA creator-binding payload (Divine extension, <code>POST /api/nostr</code> only; not available over the bunker transport)</li>
         <li><code>nip04_encrypt</code> - Legacy DM encryption</li>
         <li><code>nip04_decrypt</code> - Legacy DM decryption</li>
         <li><code>nip44_encrypt</code> - Modern DM encryption (preferred)</li>
--- a/api/src/technical_docs.rs
+++ b/api/src/technical_docs.rs
@@ -213,6 +213,7 @@ pub async fn technical_docs() -> Html<&'static str> {
         <li><code>connect</code> - Initial connection handshake</li>
         <li><code>get_public_key</code> - Returns user's public key</li>
         <li><code>sign_event</code> - Signs a Nostr event</li>
+        <li><code>sign_canonical</code> - Signs a C2PA creator-binding payload</li>
         <li><code>nip04_encrypt</code> - Legacy DM encryption</li>
         <li><code>nip04_decrypt</code> - Legacy DM decryption</li>
         <li><code>nip44_encrypt</code> - Modern DM encryption (preferred)</li>
--- a/core/src/creator_binding.rs
+++ b/core/src/creator_binding.rs
@@ -7,6 +7,23 @@ use thiserror::Error;
 
 pub const MAX_CREATOR_BINDING_PAYLOAD_BYTES: usize = 64 * 1024;
 
+/// The exact top-level fields divine-mobile's `NostrCreatorBindingService` emits.
+///
+/// The allowlist is closed on purpose: every byte of the payload is signed with
+/// the user's key, so anything accepted here becomes attacker-chosen content
+/// inside a signed blob and escapes the `content_filter` policy, which only
+/// inspects creator-supplied `claims`. Adding a field must happen in lockstep
+/// with the mobile client.
+const ALLOWED_TOP_LEVEL_FIELDS: [&str; 7] = [
+    "version",
+    "pubkey",
+    "sig_alg",
+    "created_at",
+    "claims",
+    "referenced_assertions",
+    "hard_binding",
+];
+
 #[derive(Debug, Error, PartialEq, Eq)]
 pub enum CreatorBindingError {
     #[error("creator-binding payload exceeds maximum size")]
@@ -31,11 +48,17 @@ pub enum CreatorBindingError {
     PreSignedPayload,
     #[error("creator-binding payload is invalid JSON")]
     InvalidJson,
+    #[error("creator-binding payload has unexpected field: {0}")]
+    UnexpectedField(String),
 }
 
 /// Validate that a canonical-signing payload is a creator-binding assertion for
 /// the signing key. The caller signs the original bytes after this check, so this
 /// function must not reserialize or normalize the JSON.
+///
+/// The accepted shape is closed, not just checked for required fields: unknown
+/// top-level fields are rejected so the signed bytes cannot carry
+/// caller-controlled content past the policy layer.
 pub fn validate_creator_binding_payload(
     payload: &[u8],
     signer_pubkey: PublicKey,
@@ -50,6 +73,13 @@ pub fn validate_creator_binding_payload(
 
     if object.contains_key("signature") || object.contains_key("sig") {
         return Err(CreatorBindingError::PreSignedPayload);
+    }
+
+    if let Some(unexpected) = object
+        .keys()
+        .find(|key| !ALLOWED_TOP_LEVEL_FIELDS.contains(&key.as_str()))
+    {
+        return Err(CreatorBindingError::UnexpectedField(unexpected.clone()));
     }
 
     match object.get("version").and_then(Value::as_u64) {
@@ -214,6 +244,33 @@ mod tests {
 
         let err = validate_creator_binding_payload(&payload, keys.public_key()).unwrap_err();
         assert_eq!(err, CreatorBindingError::InvalidUtf8);
+    }
+
+    #[test]
+    fn rejects_unknown_top_level_field() {
+        // An extra field would otherwise ride into the signed bytes untouched by
+        // `content_filter`, which only inspects `claims`.
+        let keys = Keys::generate();
+        let mut value: Value = serde_json::from_slice(&valid_payload(keys.public_key())).unwrap();
+        value["note"] = json!("arbitrary caller-controlled text");
+        let payload = serde_json::to_vec(&value).unwrap();
+
+        let err = validate_creator_binding_payload(&payload, keys.public_key()).unwrap_err();
+        assert_eq!(
+            err,
+            CreatorBindingError::UnexpectedField("note".to_string())
+        );
+    }
+
+    #[test]
+    fn accepts_payload_without_optional_created_at() {
+        let keys = Keys::generate();
+        let mut value: Value = serde_json::from_slice(&valid_payload(keys.public_key())).unwrap();
+        value.as_object_mut().unwrap().remove("created_at");
+        let payload = serde_json::to_vec(&value).unwrap();
+
+        validate_creator_binding_payload(&payload, keys.public_key())
+            .expect("created_at is allowed but not required");
     }
 
     #[test]
--- a/core/src/creator_binding.rs
+++ b/core/src/creator_binding.rs
@@ -10,10 +10,12 @@ pub const MAX_CREATOR_BINDING_PAYLOAD_BYTES: usize = 64 * 1024;
 /// The exact top-level fields divine-mobile's `NostrCreatorBindingService` emits.
 ///
 /// The allowlist is closed on purpose: every byte of the payload is signed with
-/// the user's key, so anything accepted here becomes attacker-chosen content
-/// inside a signed blob and escapes the `content_filter` policy, which only
-/// inspects creator-supplied `claims`. Adding a field must happen in lockstep
-/// with the mobile client.
+/// the user's key, so anything accepted here becomes caller-chosen content
+/// inside a signed blob. `content_filter` scans only the free-text fields
+/// (`claims`, `created_at`, `referenced_assertions`), so an unknown field would
+/// bypass policy entirely. Adding a field means updating this list, the
+/// per-field shape checks below, and `ContentFilter::can_sign_creator_binding`
+/// together, in lockstep with the mobile client.
 const ALLOWED_TOP_LEVEL_FIELDS: [&str; 7] = [
     "version",
     "pubkey",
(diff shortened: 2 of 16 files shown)


core/src/creator_binding.rs
```diff
@@ -0,0 +1,480 @@
...
+/// or content-filtered — and a first-wins reader downstream would see a different
+/// assertion than the one Keycast approved.
+///
+/// Adding a field means updating this struct and
+/// `ContentFilter::can_sign_creator_binding` together, in lockstep with mobile.
+// Most fields are never read: they exist so serde enforces the shape. The
+// caller signs the original bytes, so this type is a validator, not a model.
+#[allow(dead_code)]
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
```
dcadenas, line 32: Deserializing into a closed type rather than inspecting a `serde_json::Value` is doing two jobs here, and the second one is easy to miss.

`serde_json` collapses duplicate keys last-wins, but the caller signs the *original* bytes. A payload carrying `"created_at"` twice would therefore be validated and content-filtered on the surviving copy while the shadowed one rode into the signature having been inspected by nothing. A reader downstream that takes first-wins would then see a different assertion than the one approved here — which matters more than usual for a provenance feature.

serde's derived impl rejects duplicate fields and, with `deny_unknown_fields`, unknown ones, at every depth. That covers both without a hand-rolled key allowlist at each level.

Semantic constraints serde can't express stay explicit below: `version == 1`, the `sig_alg` literal, `pubkey` against the signer, and the hard-binding algorithm and digest format.

core/src/custom_permissions/content_filter.rs
```diff
@@ -48,6 +64,43 @@ impl CustomPermission for ContentFilter {
         }
     }
 
+    fn can_sign_creator_binding(&self, payload: &[u8]) -> bool {
+        let Some(words) = &self.config.blocked_words else {
+            return true;
+        };
+
+        // Scan every caller-controlled free-text surface, and only those.
```
dcadenas, line 72: This scans the three fields that can carry creator text, and deliberately not the whole document.

Matching the full payload is the tempting version, but it denies on structure rather than content. `sig_alg` is the literal `nostr.secp256k1`, so a blocklist containing `sig` would reject every request unconditionally. `pubkey` and the hard-binding digest are hex, so `bad`, `dead`, and `cafe` would reject a fraction of requests depending on the content hash — non-deterministic denial of a feature whose contract is determinism. Structural key names have the same problem: `ass` appears in `referenced_assertions`.

The invariant that makes this safe is that every other field is pinned by the validator to a fixed or hex-only value, so it carries no caller text. `blocked_word_in_pinned_pubkey_hex_does_not_deny` guards it.

Object keys are skipped for the same reason, which is why the claim key set is pinned rather than filtered — scanning keys would deny on `handle` or `platform` for a blocklist holding `and` or `for`.

api/src/technical_docs.rs
```diff
@@ -213,6 +213,7 @@ pub async fn technical_docs() -> Html<&'static str> {
         <li><code>connect</code> - Initial connection handshake</li>
         <li><code>get_public_key</code> - Returns user's public key</li>
         <li><code>sign_event</code> - Signs a Nostr event</li>
+        <li><code>sign_canonical</code> - Signs a C2PA creator-binding payload (Divine extension, <code>POST /api/nostr</code> only; not available over the bunker transport)</li>
```
dcadenas, line 216: This list sits under the NIP-46 section, which describes the bunker flow down to publishing the kind 24133 response. The signer daemon has no `sign_canonical` arm, so a bunker client following this list would get an error.

Marked as HTTP-only rather than removed, since the method is worth finding here. The admin page entry already sits under `POST /api/nostr`, so it was accurate and I left it alone.

dcadenas (approved):
The signing path is sound and the cross-repo contract holds.

The deterministic BIP-340 construction matches the issue's test vector exactly, and the
payload validator was written against what the mobile client actually sends rather than
against the issue's prose, which left the payload opaque. That was the right instinct: an
invented schema here would have rejected every real request and silently disabled the
binding.

My changes are all in one area. `sign_canonical` signs caller-supplied bytes with the user's
Nostr key, so anything the validator accepts but the policy layer never inspects ends up
signed. I closed that gap and tightened the shape to match the client exactly.

### Changes I pushed

- Rejected unknown top-level fields, duplicate keys, and unexpected types by deserializing
  into a closed shape instead of walking untyped JSON.
- Pinned `hard_binding`, `referenced_assertions`, `created_at`, and the claim key set to
  what the client emits.
- Scoped the content filter to the fields that can carry creator text, so blocklists no
  longer match structural names or hex digests.
- Corrected the supported-methods list: `sign_canonical` is available over `POST /api/nostr`
  only, not the bunker transport.
- Aligned test fixtures with the client's real `referenced_assertions` shape.

The closed shape is the debatable one. It means a future claim field is rejected until this
repo ships a matching change, rather than being accepted and signed unvalidated. Degradation
is graceful — the client treats the error as "skip the binding" and the publish proceeds —
and the lockstep requirement is documented at the type. Explicit JSON `null` is now accepted
where a wrong type was previously rejected; null carries no text and nothing reads these
fields.

### Checks

- `cargo nextest run --workspace`: 549 passed, 16 skipped
- `cargo test -p keycast_api --features integration-tests --test nostr_rpc_integration_test`:
  23 passed, including all five `sign_canonical` cases
- Required checks green on this head

The integration tests listed as unrun in the description do pass; they needed the Docker
dependency stack.

### Follow-ups

- `pkce_failure_logging_is_actually_exercised_by_these_tests` fails intermittently under
  `cargo test -p keycast_api --lib`, roughly 4 runs in 60. It predates this branch — the
  same rate reproduces on the commit before these changes — and it is unrelated to creator
  binding, so it is not fixed here. `tracing` caches callsite interest globally, so a
  parallel test can leave the `warn!` callsite disabled before the capturing subscriber is
  installed. `cargo nextest` isolates each test and never sees it; the in-process `cargo
  test` steps can. Worth its own issue.

NotThatKindOfDrLiz:
Cleanup pushed on top of the approved PR state.

What changed:
- Documented the `can_sign_creator_binding` trait precondition: callers must validate with `validate_creator_binding_payload` first, and permission implementations may rely on the closed validated shape.
- Filed #353 for the pre-existing flaky PKCE logging test.
- Updated the PR body with the current testing status, the #186 contract note, the kind-0 permission-boundary callout, the follow-up issue, and author ratification of the branch changes.

Verification:
- `cargo test -p keycast_core --lib`
- GitHub checks are green on `db7e6b1`: semantic PR, Web Tests, CLA, and `test`.

PR #361, opened 2026-08-11
fix(auth): restore email verification app links

### LINKED ISSUE #360: fix(auth): preserve app-link handoff for email verification links
## Problem

Verification emails currently link to `/api/auth/verify-email?token=...`.
That URL completes verification server-side without JavaScript, but it does not match the established mobile association path `/verify-email`.

Both iOS Universal Links and Android App Links claim `https://login.divine.video/verify-email`, and the mobile listener handles that path.
As a result, the current email URL opens in a browser instead of handing off to the installed app when the client otherwise supports link handoff.

The minimal server-rendered response also does not match the normal Keycast visual design when browser fallback is used.

## Reproduction

1. Install the Divine mobile app on a device with associated links enabled.
2. Start email registration.
3. Open the verification link from the received email.
4. Observe that the link uses `/api/auth/verify-email` and remains in the browser rather than matching the app's `/verify-email` association.

## Expected behavior

The canonical public verification link should be:

```text
https://login.divine.video/verify-email?token=...
```

The flow must preserve both outcomes:

- If the OS hands the URL to the installed app, the existing mobile verification listener handles it.
- If the URL remains in a browser, Keycast completes verification on the server without requiring page JavaScript and renders a branded success/error page.

## Acceptance criteria

- Verification emails use `/verify-email?token=...`.
- Direct `GET /verify-email` performs the existing server-side verification flow before rendering a response.
- Browser fallback covers success, expired, superseded, conflict, and retryable states with Keycast styling.
- Existing `POST /api/auth/verify-email` behavior remains unchanged.
- Existing `GET /api/auth/verify-email` links continue to work for emails already sent.
- Tests cover the canonical email URL and JavaScript-free browser completion.

Related: #262 and #268 introduced the server-side GET flow and PIN fallback for unreliable in-app browser handoff.

dcadenas:
## Summary

Verification emails now use the public `/verify-email` path claimed by the mobile apps.
Browsers still complete verification server-side, while first-party flows move to a separate interactive page to avoid redirect loops.

## Motivation

Email links used an API path that was outside the existing iOS Universal Link and Android App Link associations.
That prevented installed apps from receiving the verification token.

The legacy API GET remains available for already-sent emails.
Token-bearing responses use `Cache-Control: no-store`.

## Related Issue

- Closes #360

## Testing

- `cargo test -p keycast_api --features integration-tests test_verify_email_get_headless_returns_success_page -- --nocapture`
- `cargo test -p keycast_api --features integration-tests test_verify_email_get_first_party_token_redirects_to_interactive_page -- --nocapture`
- `cd web && bun run check`
- `cd web && bun run test`
- `cd web && bun run build`
- Verified the server-rendered fallback and interactive page on desktop and a 390px mobile viewport.
- Not tested: native link interception on physical iOS or Android devices.

## Visuals

- [x] UI/web change with screenshot attached
- [ ] No visual change

The screenshot shows the branded server-rendered fallback at a mobile viewport.

Branded email verification fallback at a 390px mobile viewport


api/src/api/http/auth.rs
```diff
@@ -2298,8 +2298,10 @@ pub async fn verify_email_get(
                 Ok(Some(_)) => {
                     // Not an outcome yet: the interactive POST records the terminal one.
                     emit("accepted", Some("interactive_handoff"), 303).await;
-                    let interactive_url =
-                        format!("/verify-email?token={}", urlencoding::encode(&token));
+                    let interactive_url = format!(
+                        "/verify-email/interactive?token={}",
```
NotThatKindOfDrLiz: This handoff path is still inside Android's claimed `android:pathPrefix="/verify-email"` prefix. The current mobile app only defines the exact `/verify-email` GoRoute, so Android can intercept this 303 target and strand the user on an unmatched in-app route. Please move the browser handoff to a path outside that prefix, such as `/email-verification/continue`, or coordinate and land the mobile route first.

keycast/src/main.rs
```diff
@@ -989,7 +989,11 @@ async fn cache_control_middleware(request: Request<Body>, next: Next) -> Respons
     let cache_value = if path.starts_with("/_app/") {
         // SvelteKit hash-versioned assets - cache forever (1 year)
         "public, max-age=31536000, immutable"
-    } else if path.starts_with("/api/") || path.starts_with("/health") || path == "/livez" {
+    } else if path.starts_with("/api/")
+        || path.starts_with("/health")
+        || path == "/livez"
+        || path == "/verify-email"
```
NotThatKindOfDrLiz, line 995: This only no-stores exact `/verify-email`. The token-bearing handoff page currently falls through to the default `public, max-age=0, must-revalidate` policy, so a shared cache can store a URL containing a single-use token. Please cover both token-bearing verification paths, either with `starts_with("/verify-email")` if the path stays here or by explicitly adding the renamed handoff path.

keycast/src/main.rs
```diff
@@ -2329,6 +2340,28 @@ mod tests {
         assert!(!request_id.is_empty());
     }
 
+    #[tokio::test]
+    async fn test_public_verify_email_route_is_not_cached() {
+        let app = Router::new()
+            .route("/verify-email", get(|| async { "verification result" }))
```
NotThatKindOfDrLiz: This test proves the cache middleware handles a stub `/verify-email` route, but it does not prove the new `public_verify_email_route` is mounted. If `.merge(public_verify_email_route)` is removed above, this still passes and the regression this PR is fixing can return. Please add a oneshot test against `public_verify_email_route()` itself.

web/src/routes/verify-email/interactive/+page.svelte
```diff
@@ -0,0 +1,5 @@
+<script lang="ts">
+	import VerifyEmail from '../+page.svelte';
```
NotThatKindOfDrLiz: This wrapper leaves the real Svelte page at `/verify-email`, but production now shadows that URL with the Axum route while Vite dev still serves the SPA there. Please move the page implementation into the interactive route and remove the shadowed `/verify-email` Svelte route so dev and production do not diverge at the same URL.

api/src/api/http/routes.rs
```diff
@@ -41,6 +41,18 @@ pub struct AuthState {
...
 
+/// Build the public email-verification route claimed by the mobile apps.
+pub fn public_verify_email_route(
+    state: Arc<KeycastState>,
+    public_cors: tower_http::cors::CorsLayer,
+    auth_tx: Option<AuthorizationSender>,
+) -> Router {
+    Router::new()
+        .route("/verify-email", get(auth::verify_email_get))
+        .layer(public_cors)
```
NotThatKindOfDrLiz: This public route is a top-level browser navigation, not an XHR endpoint. The legacy GET route does not carry CORS, and adding `allow_origin(Any)` here only widens who can read token-derived responses cross-origin. Please drop the CORS layer from this route unless there is a concrete cross-origin fetch use case.

api/src/api/http/auth.rs
```diff
@@ -2504,13 +2506,19 @@ fn verify_html_page_with_refresh(
...
         "<!doctype html><html lang=\"en\"><head><meta charset=\"utf-8\">{refresh_tag}\
          <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\
-         <title>{heading}</title>\
-         <style>body{{font-family:system-ui,-apple-system,sans-serif;background:#0d1117;color:#e6edf3;\
-         display:flex;min-height:100vh;align-items:center;justify-content:center;margin:0;padding:1rem}}\
-         .card{{max-width:420px;text-align:center;background:#161b22;border:1px solid #30363d;\
-         border-radius:1rem;padding:2rem}}h1{{font-size:1.4rem;margin:0 0 .5rem}}\
-         p{{color:#9da7b3;margin:0;font-size:.95rem}}</style></head>\
-         <body><div class=\"card\"><h1>{heading}</h1><p>{message}</p></div></body></html>",
+         <meta name=\"theme-color\" content=\"#072218\"><title>{heading} - Divine</title>\
```
NotThatKindOfDrLiz: Please use `BRAND_NAME` here instead of hardcoding `Divine`. This server-rendered page is linked from email copy that already uses the brand constant, so a brand rename should not require editing this HTML and its test separately. Also either load the declared Inter/Bricolage fonts in this self-contained page or remove those declarations; unlike the SPA, this page does not include `web/src/app.html`'s Google Fonts link.

NotThatKindOfDrLiz (changes requested):
Thanks for fixing the canonical email link target. The main shape is right, but this needs changes before merge: the new browser handoff path is still under Android's claimed /verify-email prefix, the production/dev route ownership is now split, and the new public route/cache behavior needs stronger coverage and cleanup.

api/src/api/http/routes.rs
dcadenas: Removed the public CORS layer from this browser-navigation route. The route-level test sends an external `Origin` and verifies that no `Access-Control-Allow-Origin` header is returned.

keycast/src/main.rs
dcadenas: Both token-bearing paths now receive `Cache-Control: no-store`. The middleware test covers `/verify-email` and `/email-verification/continue`, and I also verified both response headers against the unified server.

api/src/api/http/auth.rs
dcadenas: The title and logo alt text now use `BRAND_NAME`. I removed the unloaded Inter and Bricolage declarations from the standalone page and added assertions covering both changes.

web/src/routes/verify-email/interactive/+page.svelte
dcadenas: Moved the Svelte implementation to `/email-verification/continue` and removed the Svelte `/verify-email` route. Vite now proxies only the exact canonical `/verify-email` URL to Axum during development, so local and production links exercise the same server GET before the relative continuation redirect returns to Vite.

api/src/api/http/auth.rs
dcadenas: Moved the browser continuation to `/email-verification/continue`, outside Android's claimed `/verify-email` prefix. The first-party redirect integration test now asserts the new location.

keycast/src/main.rs
dcadenas: Added a `oneshot` integration test against `public_verify_email_route()` itself. It distinguishes the registered route from a 404 through the extractor's `Missing Host header` response and also verifies that the route emits no CORS header.

NotThatKindOfDrLiz:
@dcadenas I pushed a small takeover commit (`d306ed1`) for the round-2 follow-ups:

- changed the Vite `/verify-email` dev proxy to use `changeOrigin: true` instead of spoofing `Host: login.divine.video`, so local `/verify-email` resolves the same `localhost` tenant as `/api`
- updated the layout hidden path to `/email-verification/continue` so the continuation page stays a full-page auth experience without the header or verification banner
- refreshed the shared verify-email auth-event comment for the three entry paths
- documented the public `GET /verify-email?token=...` app-link contract in the identity guide

Local validation I ran:

- `git diff --check`
- `cd web && bun run check`

Please review the takeover commit and run the targeted flow verification before merge, especially a local dev click-through/request proving `/verify-email?token=...` reaches the same `localhost` tenant as registration via `/api`. Once you have that, update the PR testing notes with the new verification result.

dcadenas:
Ran the pre-merge verification you asked for.

## The dev proxy resolves the same tenant

Registered through the dev server's `/api` proxy, then opened the verification link through the same dev server.
Both resolve the `localhost` tenant, and the account comes back verified.

```
POST http://localhost:5173/api/headless/register    200
GET  http://localhost:5173/verify-email?token=...   200   "Email verified!"

tenant resolved: localhost      email_verified = t
```

Reverting only the proxy config on that same running server reproduces the old failure, so the change is what moves the outcome:

| proxy config | result |
|---|---|
| `headers: { host: 'login.divine.video' }` | Link no longer valid |
| `changeOrigin: true` | Email verified! |

## Re-verified against d306ed1b

26 server checks pass against an image built from that commit.

Against your review items:

- The handoff redirects to `/email-verification/continue` with `cache-control: no-store`.
- On a headless Android emulator, with App Links verified for `login.divine.video` against the published `assetlinks.json`, `/verify-email` resolves to the app and `/email-verification/continue` resolves to Chrome. The handoff is outside the claimed prefix in behaviour, not only in the diff.
- `no-store` is present on both token-bearing paths.
- No `Access-Control-Allow-Origin` on the public route.
- The status page renders from `BRAND_NAME` and no longer declares fonts it does not load.

Also covered: a live token verifies with no client JS, repeat opens stay 200 rather than erroring, superseded links render the terminal page, and both the legacy GET and the POST still work.

Unit tests pass: email service (11), the branded status page, and the cache-control route test.

## Not covered

A real link opened from a real inbox is not exercised here.
App Links only fire for `https://login.divine.video`, and a local stack serves cleartext on a loopback alias, so no local link is app-link eligible.
That check belongs on a deployed environment, ideally in Gmail, since that is the client keycast#262 was filed about.

iOS Universal Links are not tested.

One item outside this PR: an expired pending registration stops reporting honestly once the cleanup task deletes the row, filed as #362. It touches none of the files here.

PR #374, opened 2026-08-15
fix(signing): do not gate signing on account status

### LINKED ISSUE #373: bug(signing): signer refuses suspended and banned accounts, which disables portability for hosted-signer users
**Related:** `divinevideo/divine-funnelcake#891` (owner self-export), `divinevideo/divine-funnelcake#952` (the second layer of this same failure), `divinevideo/divine-web#591` (the export tool), `divinevideo/support-trust-safety#200`.

## Summary

Keycast refuses to sign for any account whose status is not `active`. That disables account portability for the hosted-signer population — the users least able to do anything about it, and the ones the exit tooling exists to serve.

`signer/src/signer_daemon.rs:523`:

```rust
match status {
    Some((s, verified_minor)) if s == "active" => Ok(verified_minor),
    Some(_) => Err(SignerError::permission_denied("Account restricted")),
    None => Err(SignerError::permission_denied("User not found")),
}
```

`UserStatus` is `Active | Suspended | Banned`, so `Some(_)` is exactly suspended and banned. `check_user_sign_gates` runs before all signing — relay-dispatched (`handle_sign_event`) and direct (`sign_event_direct`) — and `api/src/api/http/ap.rs:183` mirrors it.

## Impact

A suspended or banned OAuth user cannot sign anything. Concretely:

- **They cannot export their own account.** The owner-export endpoint authenticates with NIP-98, which requires signing a kind 27235 event. The signer refuses, so `createNip98AuthHeader` fails and the export never issues a request. `divine-web`'s `/exit/start` currently shows them *"This export could not be signed. Sign in again, then restart the export"* — advice that cannot work, and the exact retry-loop pattern `support-trust-safety#200` was opened to eliminate.
- **They cannot republish anywhere else.** No signing means no events to another relay, no updated relay list, no profile pointing elsewhere.
- **They cannot prove the account is theirs** to any service that asks for a signature.

Users with a local nsec or a bunker are unaffected, since they sign client-side. This is specific to hosted-signer accounts.

## Why this should change

Divine has two separable levers and enforcement should only pull one.

**Refusing to host** — the relay, media, feeds — is what a ban means and is entirely Divine's call. It is already enforced independently: `divine-funnelcake` rejects banned authors at write time (`crates/relay/src/pubsub.rs:514`), so nothing about hosting depends on the signer.

**Refusing to sign** is a different act. It does not remove someone from Divine; it disables their identity everywhere, including on infrastructure Divine does not operate and has no dispute with. The user is not ejected — they are erased.

It also contradicts what Divine publishes. `divine.video/exit` states that an account is "a Nostr keypair that Divine does not own and cannot revoke." For hosted-signer accounts, this check means Divine effectively can revoke it, and the exit ramp is decorative for precisely the population that needs it.

## Proposed change

Signing is not gated on account status.

- `check_user_active` stops returning `permission_denied` for `Suspended` and `Banned`. It still returns `permission_denied` for an unknown user, and still returns the `verified_minor` flag.
- The `verified_minor` DM containment gate (`support-trust-safety#183`) is unchanged. It addresses a different risk and must keep applying.
- `ensure_active` in `api/src/api/http/ap.rs` is brought in line so the two paths do not diverge.

Consider renaming `check_user_active` once it no longer gates on active status, so the name does not outlive its meaning.

## The counter-argument, and why it does not hold

The check plausibly exists as a ban-evasion control: a banned user who can still sign can keep operating under the same identity.

It buys nothing. Divine's relay already rejects banned authors on write, so evasion *on Divine* is blocked without the signer's help. What the check actually prevents is the user operating **somewhere else** — which is not evasion, it is the outcome portability is supposed to produce.

If ban evasion needs a stronger control, the right place is account creation: refuse to issue a *new* Keycast account to a banned operator. That is a registration question, not a signing question.

## Acceptance

- **A-1** A suspended OAuth account can produce a NIP-98 header and complete an owner export.
- **A-2** A banned OAuth account can do the same.
- **A-3** The `verified_minor` DM gate still refuses what it refused before, for accounts in every status.
- **A-4** An unknown user is still refused.
- **A-5** Relay-dispatched and direct signing behave identically; a test covers both.

## Note

`divine-funnelcake#952` must land as well. Fixing only this one lets an enforcement-affected user authenticate and then receive an archive with their moderated events silently missing — an incomplete result that looks complete. Neither fix is sufficient alone; the acceptance test for both is a suspended account and a banned account each downloading a *complete* archive.

NotThatKindOfDrLiz:
Closes #373.

## What this changes

Signing no longer refuses suspended or banned OAuth accounts, on all three paths that carried the check:

- `api/src/api/http/nostr_rpc.rs` — the RPC dispatcher (the path `divine-web`'s `DivineJWTSigner` uses)
- `api/src/api/http/auth.rs` — the fast HTTP signing endpoint
- `signer/src/signer_daemon.rs` — the NIP-46 daemon

## Why

A restricted account could not sign anything, so it could not export itself, republish elsewhere, or prove ownership to any service. It was not removed from Divine — it was disabled everywhere, including on infrastructure Divine does not operate.

Enforcement already lives where it belongs: `divine-funnelcake`'s relay rejects banned authors on write (`crates/relay/src/pubsub.rs:514`), so nothing about *hosting* depended on the signer.

It also broke the exit tooling outright. Owner self-export authenticates with NIP-98, which requires signing a kind 27235 event, so `/exit/start` was telling suspended users to "sign in again, then restart the export" — advice that cannot work, and exactly the retry-loop pattern `support-trust-safety#200` was opened to eliminate.

## The trap worth reviewing carefully

**The status query also carried the `verified_minor` flag**, which drives the DM containment gate from `support-trust-safety#183`. Exempting signing naively — by skipping the status check — would have silently disabled that gate for every signing call.

Both paths still read the row; they only skip the *refusal*. `auth.rs` was already fetching `verified_minor` separately via `get_verified_minor`, so it is unaffected there.

This is the part I would most like a second pair of eyes on.

## Deliberately unchanged

- **Encrypt and decrypt still refuse restricted accounts.** Those are communication capabilities; re-enabling DM egress for banned accounts is a separate safety decision and not required for portability. `test_suspended_user_denied_nip17_wrap_batch` and `..._unwrap_batch` are untouched.
- **`ap.rs::ensure_active` keeps its check.** ActivityPub publishing is a Divine-hosted surface, so gating it stays consistent with "enforcement governs what Divine hosts." Flagging it rather than changing it silently.

## Tests

Three tests encoded the old behaviour and are inverted and renamed, since a test named `..._denied_...` that asserts success is worse than no test:

- `test_suspended_user_allowed_nostr_rpc_sign_canonical`
- `test_suspended_user_allowed_sign_event`
- `test_suspended_user_allowed_nostr_rpc_sign`

Added `test_banned_user_allowed_sign_event` plus a `ban_user` helper — the suite covered suspended accounts but never banned ones, which is the case the exit tooling turns on.

## Verification

`cargo check --workspace`, `cargo check -p keycast_api --features integration-tests --tests`, `cargo clippy --workspace --all-targets`, and `cargo fmt --check` all pass locally.

**The integration tests were not run locally** — they need a Postgres instance and none was available in my environment. CI provisions `postgres:16` and runs the workspace suite, so the behavioural assertions are verified there, not by me. Worth a look at the CI result before merging rather than taking this section on trust.

## Does not stand alone

`divine-funnelcake#952` must land too. With only this fix, an enforcement-affected user authenticates successfully and then receives an archive with their moderated events silently missing — an incomplete result that looks complete, which is worse than the current honest failure.

The acceptance test for the pair: **a suspended account and a banned account each download a complete archive.**

### DIFF
--- a/api/src/api/http/auth.rs
+++ b/api/src/api/http/auth.rs
@@ -3821,13 +3821,14 @@ pub async fn sign_event(
     let pool = &auth_state.state.db;
     let key_manager = auth_state.state.key_manager.as_ref();
 
-    // Check account status before either signing path (fast or slow)
     let user_repo = UserRepository::new(pool.clone());
-    if let Some((status, _, _)) = user_repo.get_user_status(&user_pubkey, tenant_id).await? {
-        if !status.is_active() {
-            return Err(AuthError::Forbidden("Account restricted".to_string()));
-        }
-    }
+
+    // Account status deliberately does not gate signing (keycast#373). Signing is
+    // how an account proves it is itself and how its holder exports and republishes
+    // elsewhere; enforcement governs what Divine hosts — the relay rejects banned
+    // authors on write — not whether the identity can act off Divine. The
+    // verified_minor DM containment gate below is fetched separately and still
+    // applies in every account status.
 
     // Parse unsigned event first for validation
     let unsigned_event: UnsignedEvent = serde_json::from_value(req.event.clone())
--- a/api/src/api/http/nostr_rpc.rs
+++ b/api/src/api/http/nostr_rpc.rs
@@ -222,9 +222,18 @@ async fn nostr_rpc_inner(
     // The same per-request row also carries verified_minor, which drives the DM
     // containment gate below (support-trust-safety#183); reading it here (not
     // from the cached handler) means protection flips apply immediately.
+    // `sign_event` still reads the row, because the verified_minor gate below
+    // depends on it, but is not refused on account status (keycast#373).
     let needs_status_check = !matches!(req.method.as_str(), "get_public_key");
     let verified_minor = if needs_status_check {
-        check_user_status_active(pool, &handler.user_pubkey_hex(), tenant_id).await?
+        let deny_when_restricted = req.method.as_str() != "sign_event";
+        check_user_status_active(
+            pool,
+            &handler.user_pubkey_hex(),
+            tenant_id,
+            deny_when_restricted,
+        )
+        .await?
     } else {
         false
     };
@@ -455,14 +464,22 @@ fn http_rpc_outcome(response: &Result<Json<NostrRpcResponse>, RpcError>) -> &'st
     }
 }
 
-/// Check that the user's account is active before allowing mutating operations.
+/// Read the account gate state before a mutating operation.
 /// This runs a DB query per request (not cached) so status changes take effect immediately.
 /// Returns the account's `verified_minor` flag (same row, no extra query) for
 /// the DM containment gate; a missing user row is a refusal, never a default.
+///
+/// `deny_when_restricted` decides whether a non-active status refuses the call.
+/// It is false for `sign_event` (keycast#373): the row is still read, because the
+/// verified_minor gate depends on it, but a suspended or banned account may still
+/// sign. Signing is how an account proves it is itself and how its holder exports
+/// and republishes elsewhere. Enforcement governs what Divine hosts — the relay
+/// rejects banned authors on write — not whether the identity can act off Divine.
 async fn check_user_status_active(
     pool: &sqlx::PgPool,
     user_pubkey_hex: &str,
     tenant_id: i64,
+    deny_when_restricted: bool,
 ) -> Result<bool, RpcError> {
     let status_started = Instant::now();
     METRICS.set_http_rpc_db_pool_state(pool.size(), pool.num_idle() as u32);
@@ -509,6 +526,14 @@ async fn check_user_status_active(
 
     let result = match status {
         Some((s, verified_minor)) if s == "active" => Ok(verified_minor),
+        Some((s, verified_minor)) if !deny_when_restricted => {
+            tracing::info!(
+                event = "rpc.sign_allowed_for_restricted_account",
+                status = %s,
+                "Signing allowed for a restricted account (keycast#373)"
+            );
+            Ok(verified_minor)
+        }
         Some(_) => Err(RpcError::AccountSuspended("Account restricted".to_string())),
         None => Err(RpcError::Auth(AuthError::InvalidToken)),
     };
--- a/api/tests/nostr_rpc_integration_test.rs
+++ b/api/tests/nostr_rpc_integration_test.rs
@@ -1068,7 +1068,7 @@ async fn test_sign_canonical_denied_by_readonly_policy() {
 
 #[tokio::test]
 #[serial]
-async fn test_suspended_user_denied_nostr_rpc_sign_canonical() {
+async fn test_suspended_user_allowed_nostr_rpc_sign_canonical() {
     let pool = setup_db().await;
     let tenant_id = create_test_tenant(&pool).await;
     let (user_keys, pubkey) = create_test_user();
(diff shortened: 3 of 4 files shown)


NotThatKindOfDrLiz:
CI caught an over-correction on my part, now fixed in `21102ef`.

`sign_canonical` signs a **creator-binding payload** — a Divine-hosted capability, not the generic event signing an account needs to export itself or republish elsewhere. The production change already exempted only `sign_event`, so the code was right; inverting `test_suspended_user_denied_nostr_rpc_sign_canonical` alongside the other two was the error. It is restored to asserting denial, with the reason recorded beside it.

This is a useful boundary to have made explicit, so stating it plainly for review: **generic event signing is ungated; Divine-specific signing capabilities stay gated.** `sign_canonical` and `ap.rs::ensure_active` both sit on the gated side, along with encrypt and decrypt.

rabble (approved):
Approving. The policy is the one I want: enforcement governs what Divine hosts, not whether a person's identity can act anywhere else. A suspended or banned account that cannot sign is not moderated, it is disabled everywhere, including on infrastructure we do not run — and it cannot export itself, which is the outcome I care most about avoiding.

## The trap you flagged, checked

You asked for eyes on the `verified_minor` interaction. I checked all three paths against the head of this branch, because silently dropping the minor DM containment gate would have been much worse than the bug being fixed:

- `nostr_rpc.rs` — `check_user_status_active` still reads the row and still returns `verified_minor`; `deny_when_restricted` only suppresses the refusal. A missing user row is still `InvalidToken`, so absence still fails closed.
- `auth.rs` — the removed block was never the source of the flag. `get_verified_minor` is fetched separately for minor-gated kinds and `.ok_or_else(...)` makes a missing row a `Forbidden`. Both fast and slow paths call `enforce_minor_dm_sign` off that value.
- `signer_daemon.rs` — `check_user_sign_gates` passes `false` but still resolves the flag and still calls `validate_minor_sign`. `check_user_active` (`true`) remains on the encrypt/decrypt callers at 629 and 697.

The gate survives in every status on every path. That was the thing worth getting right, and it is right.

## The boundary this draws

Worth stating plainly on the record, because it is the part a future reader will need: generic event signing is ungated, Divine-specific signing capabilities stay gated. `sign_canonical` (creator binding), `ap.rs::ensure_active` (ActivityPub publishing), encrypt and decrypt all stay on the gated side. Your correction in 21102ef is the right call, and a test named `..._denied_...` asserting success would have been worse than no test at all.

The consequence I am accepting knowingly: a banned account can now sign a kind-1 with its divine.video identity and publish it to relays we do not run. That is the intended shape. Our relay still rejects banned authors on write, which is where hosting is decided. This matches the account-status decision I gave on divinevideo/divine-funnelcake#1081 — a ban stops publishing here, it is not a retention hold and not an identity revocation.

## Merge order

Your "does not stand alone" section is the operative constraint, not a caveat. Landing this alone gives an enforcement-affected user a successful authentication followed by an archive that is silently missing their moderated events — a complete-looking incomplete result, which is worse than today's honest failure. divinevideo/divine-funnelcake#952 has to land with it, and the acceptance test is the one you named: a suspended account and a banned account each download a complete archive.

## Verification

I did not run anything locally. CI on this head is green, including `test` (5m57s) and Web Tests, which is where the Postgres-backed assertions actually run — you were right to say so rather than let the local `cargo check` list read as coverage it isn't.

